PR #14, opened 2023-06-07
Add api key and secret configs to the operator

apurvam:
This is a port of https://github.com/responsivedev/responsive-platform/pull/15

### DIFF
--- a/operator/build.gradle.kts
+++ b/operator/build.gradle.kts
@@ -31,10 +31,17 @@ dependencies {
     implementation(libs.jackson)
     implementation(libs.javaoperatorsdk)
 
+    implementation("commons-cli:commons-cli:1.5.0")
+    implementation("org.apache.commons:commons-configuration2:2.8.0")
+    implementation("commons-beanutils:commons-beanutils:1.9.4")
+
     annotationProcessor(libs.javaoperatorsdk)
     annotationProcessor(libs.crd.generator.atp)
 
     testImplementation(testlibs.bundles.base)
+    testImplementation("io.grpc:grpc-testing:1.53.0")
+    testImplementation("io.grpc:grpc-examples:0.15.0")
+
 }
 
 responsive_docker.dockerImage.set("responsive-operator:$version")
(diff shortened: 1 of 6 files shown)


operator/build.gradle.kts
```diff
@@ -31,10 +31,17 @@ dependencies {
     implementation(libs.jackson)
     implementation(libs.javaoperatorsdk)
 
+    implementation("commons-cli:commons-cli:1.5.0")
```
agavra, line 34: nit: can we follow the `libs.` convention in the other dependencies? see the top level `settings.gradle.kts` file

especially for common stuff like `org.apache.commons`

operator/src/main/java/dev/responsive/k8s/operator/OperatorMain.java
```diff
@@ -21,12 +21,94 @@
...
+      return configurations.properties(new File(fileName));
+    } catch (Exception e) {
+      LOGGER.error("Error loading configuration properties: {}", e.getMessage());
+    }
+    return new PropertiesConfiguration();
+  }
+
+  private static Options getOptions() {
+    Options options = new Options();
+    Option serverOption = Option.builder("server-url")
```
agavra, line 93: thoughts on putting this in the config file instead? I think that would make it a bit easier when interacting with the docker image for the operator (maybe?) 

also we may want to name it `controller.url` instead in case we eventually have other types of servers

operator/src/main/java/dev/responsive/k8s/operator/OperatorMain.java
```diff
@@ -21,12 +21,94 @@
...
     final Operator operator = new Operator();
     Serialization.jsonMapper().registerModule(new Jdk8Module());
-    operator.register(new ResponsivePolicyReconciler(new ControllerGrpcClient(args[0])));
+    operator.register(new ResponsivePolicyReconciler(new ControllerGrpcClient(target,
+            apiKey, secret)));
     operator.start();
   }
+
+  private static Configuration getConfigurations(String fileName) {
+    Configurations configurations = new Configurations();
```
agavra, line 82: just curious if there was any reason you used this instead of `java.util.Properties`? Even if there's no reason we don't need to change it, I was just wondering.

agavra (approved):
LGTM

PR #18, opened 2023-06-09
wire in api secrets with helm and volumes

agavra:


operator/docker/scripts/run-operator
```diff
@@ -1 +1,8 @@
-java -Dorg.apache.logging.log4j.level=INFO -cp "/usr/share/java/responsive-operator/*" dev.responsive.k8s.operator.OperatorMain --controllerUrl ${CONTROLLER_EP} --configFile "/dummy"
+if test -f /etc/secrets/ctl-api-key; then
```
rodesai, line 1: We should do this slightly differently. We really shouldn't be copying the secrets onto the local filesystem, as it could expose them outside the container. The "idiomatic" way to do this would be to:
1. Put the key/secret in the same k8s secret, formatted as a properties file
2. Pass the path to this properties file directly to operator. This does mean that we will likely eventually need two separate properties files in operator (one for this secret and one for other properties). But for now you could just rename the option to something like `secretsFile`.

agavra: https://github.com/responsivedev/responsive-pub/pull/21

PR #21, opened 2023-06-12
use b64 encoded secrets file

agavra:
Use `java.util.Properties` because it accepts a reader which can either read directly from a non-b64 encoded file or just load a `Reader` with the decoded contents.


apurvam (commented):
Sorry, I missed it in the discussion earlier. Why do we need to pass a base64 encoded secrets file here? How will this file be populated? In particular, I'm assuming that the user won't have to do the base64 encoding right?

agavra:
talked offline - linking Rohan's comment here too: https://github.com/responsivedev/responsive-pub/pull/18#discussion_r1227146311

apurvam (approved):
Synced offline. This seems like the the best way to pass in the secrets using the k8s secrets systems, and the UX seems OK too.

PR #26, opened 2023-06-15
better handling of get target state errors

rodesai:
This patch improves handling of get target state errors. When we get an error, we still want to run the main reconciler, because the controller relies on periodic updates from operator. So, instead of returning empty from the poller, we return an updated status with an empty target state - which forces the framework to re-run reconciliation.

### DIFF
--- a/operator/src/main/java/dev/responsive/k8s/operator/reconciler/DemoPolicyPlugin.java
+++ b/operator/src/main/java/dev/responsive/k8s/operator/reconciler/DemoPolicyPlugin.java
@@ -74,25 +74,35 @@ public class DemoPolicyPlugin implements PolicyPlugin {
     );
 
     final var maybeTargetState = ctx.getSecondaryResource(TargetStateWithTimestamp.class);
-    if (maybeTargetState.isPresent()) {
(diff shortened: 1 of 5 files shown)


operator/src/main/java/dev/responsive/k8s/operator/reconciler/DemoPolicyPlugin.java
```diff
@@ -74,25 +74,35 @@ public void reconcile(
...
+    LOGGER.info("target state {} {}", appName, targetState);
+
+    if (targetState.getTargetState().isEmpty()) {
+      LOGGER.info(
+          "we were not able to get a target state from controller, so don't try to reconcile one");
+      return;
+    }
+    final var targetReplicas = targetState.getTargetState().get().getDemoState().getReplicas();
+    if (targetReplicas != deployment.getSpec().getReplicas()) {
+      LOGGER.info("Scaling from {} to {}", deployment.getSpec().getReplicas(), targetReplicas);
```
agavra, line 92: nit - for all of these logs lets include the deployment name

agavra (approved):
lgtm!

PR #28, opened 2023-06-16
fix naming in policy spec

agavra:
A policy that looks like:
```
apiVersion: "application.responsive.dev/v1"
kind: "ResponsivePolicy"
metadata:
  name: example-policy
  namespace: policies
spec:
  applicationNamespace: responsive
  applicationName: example
  status: POLICY_STATUS_MANAGED
  policyType: DEMO
  demoPolicy:
    maxReplicas: 2
```
should manage `responsive/example` not `policies/example-policy`


rodesai (approved):
LGTM

PR #34, opened 2023-06-21
Add responsive client supplier

rodesai:
This patch adds a client supplier that supplies responsive consumer/producer to KafkaStreams,
and hooks these up to our metrics publishers.

Depends on: https://github.com/responsivedev/responsive-pub/pull/33

PR #33, opened 2023-06-21
add consumer/producer listeners that publish metrics

rodesai:
EndOffsetsPoller: listens for partitions assigned/revoked, polls the end offsets
    for the assigned partitions, and publishes the offset as a metric.
MetricsPublishingCommitListener: listens for partitions assigned/revoked and for
    committed offsets, and publishes the current commmitted offset for all assigned
    partitions.

Depends on: https://github.com/responsivedev/responsive-pub/pull/32

PR #36, opened 2023-06-22
add responsive client supplier for publishing metrics

ableegoldman:
Just opening a mock PR to facilitate review of https://github.com/responsivedev/responsive-pub/pull/34

This shows [PR #34](https://github.com/responsivedev/responsive-pub/pull/34) relative to [PR #33](https://github.com/responsivedev/responsive-pub/pull/33)

### DIFF
--- a/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
+++ b/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
@@ -122,13 +122,20 @@ public class EndOffsetsPoller {
 
   private void initPoller() {
     LOGGER.info("init end offsets poller");
-    assert poller == null;
+    if (poller != null) {
+      throw new IllegalStateException("poller already initialized");
+    }
     poller = new Poller(factories.createAdminClient(configs), executor, this::getThreadMetrics);
   }
 
   private void stopPoller() {
     LOGGER.info("stopping end offsets poller");
-    poller.stop();
+    try {
+      poller.stop();
+    } catch (final RuntimeException e) {
+      LOGGER.warn("poller stop returned an unexpected error. It will be ignored, and the poller "
+          + "task + admin client might be leaked.", e);
+    }
     poller = null;
   }
 
@@ -213,14 +220,14 @@ public class EndOffsetsPoller {
     }
 
     private void update(final Map<TopicPartition, ListOffsetsResultInfo> endOffsets) {
-      for (final var tp : endOffsets.keySet()) {
-        this.endOffsets.computeIfPresent(tp, (k, v) -> {
-          logger.info("update end offset metric of {}/{} to {}",
-              tp.topic(),
-              tp.partition(),
-              endOffsets.get(tp)
+      for (final var entry : endOffsets.entrySet()) {
+        this.endOffsets.computeIfPresent(entry.getKey(), (k, v) -> {
+          logger.info("update end offset of {}/{} to {}",
+              k.topic(),
+              k.partition(),
+              entry.getValue().offset()
           );
-          return endOffsets.get(tp).offset();
+          return entry.getValue().offset();
         });
       }
     }
--- a/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
+++ b/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
@@ -45,26 +45,38 @@ public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
   private final List<Listener> listeners;
   private final Logger logger;
 
-  private class RebalanceListener implements ConsumerRebalanceListener {
-    final Optional<ConsumerRebalanceListener> wrappedRebalanceListener;
+  private static class RebalanceListener implements ConsumerRebalanceListener {
+    private final Optional<ConsumerRebalanceListener> wrappedRebalanceListener;
+    private final List<Listener> listeners;
+    private final Logger logger;
 
-    public RebalanceListener() {
-      this(Optional.empty());
+    public RebalanceListener(final List<Listener> listeners, final Logger logger) {
+      this(Optional.empty(), listeners, logger);
     }
 
-    public RebalanceListener(final ConsumerRebalanceListener wrappedRebalanceListener) {
-      this(Optional.of(wrappedRebalanceListener));
+    public RebalanceListener(
+        final ConsumerRebalanceListener wrappedRebalanceListener,
+        final List<Listener> listeners,
+        final Logger logger
+    ) {
+      this(Optional.of(wrappedRebalanceListener), listeners, logger);
     }
 
-    private RebalanceListener(final Optional<ConsumerRebalanceListener> wrappedRebalanceListener) {
+    private RebalanceListener(
+        final Optional<ConsumerRebalanceListener> wrappedRebalanceListener,
+        final List<Listener> listeners,
+        final Logger logger
+    ) {
       this.wrappedRebalanceListener = wrappedRebalanceListener;
+      this.listeners = listeners;
+      this.logger = logger;
     }
 
     @Override
     public void onPartitionsLost(final Collection<TopicPartition> partitions) {
       wrappedRebalanceListener.ifPresent(l -> l.onPartitionsLost(partitions));
       for (final var l : listeners) {
-        ignoreException(() -> l.onPartitionsLost(partitions));
+        ignoreException(logger, () -> l.onPartitionsLost(partitions));
       }
     }
 
@@ -72,7 +84,7 @@ public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
     public void onPartitionsRevoked(final Collection<TopicPartition> partitions) {
       wrappedRebalanceListener.ifPresent(l -> l.onPartitionsRevoked(partitions));
       for (final var l : listeners) {
-        ignoreException(() -> l.onPartitionsRevoked(partitions));
+        ignoreException(logger, () -> l.onPartitionsRevoked(partitions));
       }
     }
 
@@ -80,7 +92,7 @@ public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
     public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
       wrappedRebalanceListener.ifPresent(l -> l.onPartitionsAssigned(partitions));
       for (final var l : listeners) {
-        ignoreException(() -> l.onPartitionsAssigned(partitions));
+        ignoreException(logger, () -> l.onPartitionsAssigned(partitions));
       }
     }
   }
@@ -113,17 +125,17 @@ public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
 
   @Override
   public void subscribe(final Collection<String> topics) {
-    wrapped.subscribe(topics, new RebalanceListener());
+    wrapped.subscribe(topics, new RebalanceListener(listeners, logger));
   }
 
   @Override
   public void subscribe(final Collection<String> topics, final ConsumerRebalanceListener callback) {
-    wrapped.subscribe(topics, new RebalanceListener(callback));
+    wrapped.subscribe(topics, new RebalanceListener(callback, listeners, logger));
   }
 
   @Override
   public void subscribe(final Pattern pattern, final ConsumerRebalanceListener callback) {
-    wrapped.subscribe(pattern, new RebalanceListener(callback));
+    wrapped.subscribe(pattern, new RebalanceListener(callback, listeners, logger));
   }
 
   @Override
@@ -354,6 +366,10 @@ public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
   }
 
   private void ignoreException(final Runnable r) {
+    ignoreException(logger, r);
+  }
+
+  private static void ignoreException(final Logger logger, final Runnable r) {
     try {
       r.run();
     } catch (final Throwable t) {
--- a/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
+++ b/kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
@@ -1,0 +1,333 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.clients;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import org.apache.kafka.clients.CommonClientConfigs;
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.common.Configurable;
+import org.apache.kafka.common.metrics.JmxReporter;
+import org.apache.kafka.common.metrics.KafkaMetricsContext;
+import org.apache.kafka.common.metrics.MetricConfig;
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.MetricsContext;
+import org.apache.kafka.common.metrics.MetricsReporter;
+import org.apache.kafka.common.utils.Time;
+import org.apache.kafka.streams.KafkaClientSupplier;
+import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.processor.internals.DefaultKafkaClientSupplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An implementation of KafkaClientSupplier for Responsive applications. The supplier returns
+ * wrapped kafka producers/consumers that are instrumented to work with the responsive platform
+ * by, for example, emitting metrics useful for debug and scaling.
+ *
+ * Synchronization: This class is read/written during client initialization and close (via the
+ * close callbacks on the returned clients). As these calls are not on the performance path we
+ * rely on coarse-grained locks around sections reading/writing shared state.
+ */
+public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier, Configurable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ResponsiveKafkaClientSupplier.class);
+  private final SharedListeners sharedListeners = new SharedListeners();
+  private final KafkaClientSupplier wrapped;
+  private final Factories factories;
+  private boolean eosV2 = false;
+  private boolean configured = false;
+  private Metrics metrics;
+  private EndOffsetsPoller endOffsetsPoller;
+
+  public ResponsiveKafkaClientSupplier() {
+    this(new Factories() {}, new DefaultKafkaClientSupplier());
+  }
+
+  ResponsiveKafkaClientSupplier(final Factories factories, final KafkaClientSupplier wrapped) {
+    this.factories = factories;
+    this.wrapped = wrapped;
+  }
+
+  @Override
+  public void configure(final Map<String, ?> configs) {
+    LOGGER.trace(
+        "Configuring the client supplier. Got configs: {}",
+        configs.entrySet().stream()
+            .map(e -> e.getKey() + ":" + e.getValue())
+            .collect(Collectors.joining("\n"))
+    );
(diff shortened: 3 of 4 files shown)


kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+
+  @Override
+  public void configure(final Map<String, ?> configs) {
+    LOGGER.info(
+        "Configuring the client supplier. Got configs: {}",
+        configs.entrySet().stream()
+            .map(e -> e.getKey() + ":" + e.getValue())
+            .collect(Collectors.joining("\n"))
+    );
+    eosV2 = (configs.get(StreamsConfig.PROCESSING_GUARANTEE_CONFIG))
```
ableegoldman, line 84: Should we check for eos v1 just in case?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+  @Override
+  public Consumer<byte[], byte[]> getConsumer(final Map<String, Object> config) {
+    assert configured;
+    if (!eosV2) {
+      return wrapped.getConsumer(config);
+    }
+    LOGGER.info("Creating responsive consumer");
+    final String tid = threadIdFromConsumerConfig(config);
+    synchronized (this) {
+      // TODO: the end offsets poller call is kind of heavy for a synchronized block
```
ableegoldman, line 135: Yeah why not just do that up front/outside this lock?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+
+  @Override
+  public Consumer<byte[], byte[]> getGlobalConsumer(final Map<String, Object> config) {
+    return wrapped.getGlobalConsumer(config);
+  }
+
+  private String threadIdFromProducerConfig(final Map<String, Object> config) {
+    final String clientId = (String) config.get(CommonClientConfigs.CLIENT_ID_CONFIG);
+    LOGGER.info("Found producer client ID {}", clientId);
+    final var regex = Pattern.compile(".*-(StreamThread-\\d+)-producer");
```
ableegoldman, line 172: This doesn't hold up for eos-v1 (just a note for the future in case we run into any sad clients still using it)

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+              )
+          )
+      );
+    }
+  }
+
+  @Override
+  public Consumer<byte[], byte[]> getConsumer(final Map<String, Object> config) {
+    assert configured;
+    if (!eosV2) {
```
ableegoldman, line 129: Why this?

ableegoldman (commented):
Haven't finished looking through this but I'll leave my first batch of comments for the night

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
rodesai: I guess there's nothing here that wouldn't work for v1. I'll add that.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  public ResponsiveConsumer.Listener addForThread(final String threadId) {
+    LOGGER.info("add end offset metrics for thread {}", threadId);
+    final var tm = new ThreadMetrics(threadId);
+    synchronized (this) {
+      assert !threadIdToMetrics.containsKey(threadId);
+      if (threadIdToMetrics.isEmpty()) {
+        initPoller();
+        threadIdToMetrics.put(threadId, tm);
```
agavra, line 92: should this be outside of the `if`?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+  private Map<String, Object> consumerConfigs(final Map<String, ?> providedConfigs) {
+    final HashMap<String, Object> configs = new HashMap<>(providedConfigs);
+    configs.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, BytesDeserializer.class);
+    configs.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, BytesDeserializer.class);
+    return Map.copyOf(configs);
+  }
+
+  public ResponsiveConsumer.Listener addForThread(final String threadId) {
+    LOGGER.info("add end offset metrics for thread {}", threadId);
+    final var tm = new ThreadMetrics(threadId);
```
agavra, line 87: nit: creating this is practically free, can we just synchronize the whole method?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+    @Override
+    public void onPartitionsRevoked(final Collection<TopicPartition> revoked) {
+      for (final var p : revoked) {
+        metrics.removeMetric(metricName(p));
+        endOffsets.remove(p);
+      }
+    }
+
+    @Override
+    public void onPartitionsAssigned(final Collection<TopicPartition> assigned) {
```
agavra, line 168: I believe @ableegoldman confirmed with me that this is only called after the first `poll` (you don't get assigned any partitions on an assignment until you poll) - I believe we'd never get anything assigned to us?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+      threadMetrics = List.copyOf(this.threadIdToMetrics.values());
+    }
+    for (final var tm : threadMetrics) {
+      partitions.addAll(tm.endOffsets.keySet());
+    }
+    final var endOffsets = consumer.endOffsets(partitions);
+    threadMetrics.forEach(tm -> tm.update(endOffsets));
+  }
+
+  private void initPoller() {
```
agavra, line 124: can we also make this and `stopPoller` both synchronized methods? all access to them seems to be while holding `this` so it should be free and gives us an extra guarantee that they won't be called at the same time

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  private void stopPoller() {
+    LOGGER.info("stopping end offsets poller");
+    try {
+      consumer.close(Duration.ofNanos(0));
+    } catch (final Exception e) {
+      LOGGER.warn("error closing consumer", e);
+    }
+    consumer = null;
```
agavra, line 140: I believe there's a race here with `pollEndOffsets` - we could set this to null/close it and then after that a previously scheduled `pollEndOffsets` call gets triggered and line 120 would throw an NPE (or already closed exception)

is it really so bad to just synchronize everything in this class so we don't have to think about races? how often are we starting/stopping consumers?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+    pollFuture.cancel(true);
+    executor.shutdown();
+    executor = null;
+  }
+
+  private class ThreadMetrics implements ResponsiveConsumer.Listener {
```
agavra, line 146: I think I commented elsewhere, but it would be great if we made this one static and passed in what we need - especially since this is getting passed elsewhere we could leak a reference to the poller

EDIT: ahh I just saw the way this is used with the synchronization, this makes me a bit nervous that we're also leaking the lock on `this` and passing that around. I'm going to write a top-level comment about suggesting a different architecture.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+      }
+    }
+
+    private void update(final Map<TopicPartition, Long> endOffsets) {
+      for (final var tp : endOffsets.keySet()) {
+        this.endOffsets.computeIfPresent(tp, (k, v) -> {
+          LOGGER.info("update end offset of {}/{} to {}",
+              tp.topic(),
+              tp.partition(),
+              endOffsets.get(tp)
```
agavra, line 184: suggestion: avoid this extra lookup by just iterating `endOffsets.entrySet()`

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -0,0 +1,220 @@
...
+        metrics.addMetric(
+            metricName(p),
+            (Gauge<Long>) (config, now) -> endOffsets.getOrDefault(p, -1L)
+        );
+      }
+    }
+
+    private void update(final Map<TopicPartition, Long> endOffsets) {
+      for (final var tp : endOffsets.keySet()) {
+        this.endOffsets.computeIfPresent(tp, (k, v) -> {
```
agavra, line 180: why are we only computing if present here? wouldn't we want to add all endOffsets to the map?

agavra (commented):
The architecture here makes me a bit nervous - the synchronization and reference counting is really difficult to reason about and I'm worried it will prove to be brittle. Here's a suggested alternative architecture:

1. we use a thread safe, in-memory queue that functions as a buss where all the callbacks publish metrics to (and the End Offsets Poller). The event that's enqueued would be `thread_id, partition, metric, value`.
2. we have a single thread (this can be kicked of via the `ResponsiveDriver#connect` method) which consumes from this in memory queue and publishes metrics to where they need to go.

If I'm understanding things correctly, this should really simplify things as we won't need to reference count anything and we won't need to reason much about synchronization. Instead the main thread just reads from this long-lived message bus and when events stop coming in from a certain thread/partition metrics for it stop being produced.

Might be worth syncing offline.

kafka-instrumentation/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
+/*
```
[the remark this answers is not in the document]
rodesai: I wasn't sure what the intent for that package was. That's what I started with, but there were no actual kafka clients there so I added this new thing. If that was the idea all along I'll just move it.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: done already in #33.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: Ah good catch. I'd rather not block this on the client call - but I'll fix this race.

kafka-instrumentation/build.gradle.kts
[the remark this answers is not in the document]
agavra: 😆 it was meant to indicate it's responsive's kafka-related artifacts but I guess "kafka client" means something specific

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+    this.factories = factories;
+    this.wrapped = wrapped;
+  }
+
+  @Override
+  public void configure(final Map<String, ?> configs) {
+    LOGGER.info(
+        "Configuring the client supplier. Got configs: {}",
+        configs.entrySet().stream()
+            .map(e -> e.getKey() + ":" + e.getValue())
```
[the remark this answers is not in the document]
rodesai: it will - I'll change to a trace mode logger

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+    return wrapped.getAdmin(config);
+  }
+
+  @Override
+  public Producer<byte[], byte[]> getProducer(final Map<String, Object> config) {
+    assert configured;
+    if (!eosV2) {
+      return wrapped.getProducer(config);
+    }
+    LOGGER.info("Creating responsive producer");
```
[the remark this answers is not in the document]
rodesai: I wanted to know that I actually wrapped the producer (the check above didn't fire)

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ResponsiveConsumer<K, V> implements Consumer<K, V> {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ResponsiveConsumer.class);
+
+  private final Consumer<K, V> wrapped;
+  private final List<Listener> listeners;
+
+  private class RebalanceListener implements ConsumerRebalanceListener {
```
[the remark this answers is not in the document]
rodesai: I can make it static - though I don't really understand the reference leaking concern here. I really don't think it should be package-private as it's strictly an implementation detail of this class.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+
+  private void ignoreException(final Runnable r) {
+    try {
+      r.run();
+    } catch (final Throwable t) {
+      LOGGER.error("error calling rebalance listener", t);
+    }
+  }
+
+  interface Listener {
```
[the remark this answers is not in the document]
rodesai: > nit - this seems to make more sense as abstract class Listener extends ConsumerRebalanceListener

I generally don't like to use abstract classes unless there really needs to be some common state

> I also think we should have a different interface CloseListener that both this and ResponsiveProducer.Listener extend.

why?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: > I believe we'd never get anything assigned to us?

I'm not following - when the stream thread calls `poll` and an assignment happens - this will get called right?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: LMK if you still want this after the patch that cleans this up.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: done

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
rodesai: End offsets has all the end offsets that were polled (for all threads).

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
rodesai: I think this is fine after my last patch

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
rodesai: Ack - will leave my eosv2 only check in for now then.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  @Override
+  public Admin getAdmin(final Map<String, Object> config) {
+    return wrapped.getAdmin(config);
+  }
+
+  @Override
+  public Producer<byte[], byte[]> getProducer(final Map<String, Object> config) {
+    assert configured;
```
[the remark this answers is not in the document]
rodesai: good call. Need to clean these up. I like having lots of them when I'm doing the initial testing.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+  }
+
+  @Override
+  public void subscribe(final Pattern pattern, final ConsumerRebalanceListener callback) {
+    wrapped.subscribe(pattern, new RebalanceListener(callback));
+  }
+
+  @Override
+  public void subscribe(final Pattern pattern) {
+    wrapped.subscribe(pattern);
```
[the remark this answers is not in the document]
rodesai: yep, fixing.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -0,0 +1,375 @@
...
+
+  @Override
+  public void close() {
+    wrapped.close();
+    listeners.forEach(l -> ignoreException(l::onClose));
+  }
+
+  @Override
+  public void close(final Duration timeout) {
+    wrapped.close(timeout);
```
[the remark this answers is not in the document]
rodesai: close doesn't throw on a timeout

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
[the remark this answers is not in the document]
agavra: the reference leaking of non-static inner classes means that if another class holds onto this it will also have a reference to the parent class, which can make garbage collection leaks difficult to debug - I believe this is passed to the other classes as part of the rebalance listener API, right?

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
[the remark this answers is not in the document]
agavra: ah nice find

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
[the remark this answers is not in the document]
agavra: makes sense to me - for the second one it's just a bit sneaky when you have a listener that implements both and only has one `onClose` method (name overloading). I didn't even realize that was possible for a second!

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
agavra: I think I may have misread the code, but I thought we were using a consumer that we created just for the `EndOffsetsPoller` - I can't find it now though so we can ignore that.

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -0,0 +1,328 @@
...
+    final T val;
+    int count;
+
+    private ReferenceCounted(final T val) {
+      this.val = val;
+      this.count = 1;
+    }
+
+    private void ref() {
+      LOGGER.info("bumping ref count to {}", count + 1);
```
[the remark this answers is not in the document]
rodesai: yep totally agree - changed it to pass in a name for logging

kafka-instrumentation/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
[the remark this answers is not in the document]
rodesai: Yeah, it's passed and held internally within the consumer in the objects that maintain metadata about the subscription. There's definitely a cycle in the reference chain. Something like:

stream thread -> responsive consumer -> consumer -> subscription metadata -> responsive rebalance listener -> responsive consumer.

But my understanding was that's fine - they will all get cleaned up once the stream thread drops the reference to the consumer (or the stream thread goes away).

So I'm still not understanding the concern. Not trying to be argumentative (I've already changed it to static) - I'm just genuinely curious how a non-static class could cause an issue with leak - it's not something I've run into or considered before.

PR #37, opened 2023-06-22
support for Global KTables

agavra:
Review Guide:

1. Overall strategy is to use a consumer group for the global consumer instead of manual assignment so that all the nodes can share the work of updating the global store.
2. To do this, we use a custom global consumer `RepsonsiveGlobalConsumer` - the comments at the top are pretty comprehensive
3. We also use a custom store that simply writes everything to the same Cassandra Partition `ResponsiveGlobalStore`. We can optimize this by preparing statements instead of using the `CassandraClient` directly. Currently this part doesn't have unit test coverage.

I'll work on integrating it with Rohan's PRs in a bit.


kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -159,6 +183,48 @@ public <K, V> Materialized<K, V, WindowStore<Bytes, byte[]>> materialized(
...
+  @Override
+  public <K, V> Materialized<K, V, KeyValueStore<Bytes, byte[]>> globalMaterialized(
+      final String name
+  ) {
+    return Materialized.<K, V>as(globalKv(name))
+        .withCachingDisabled();
+  }
+
+  @Override
+  public Admin getAdmin(final Map<String, Object> config) {
```
ableegoldman, line 195: This and all the below needs to be provided to the user as a `KafkaClientSupplier` rather than as individual callbacks. They'll do something like this:

```
final ResponsiveKafkaClientSupplier clientSupplier = responsiveDriver.getKafkaClientSupplier();
final KafkaStreams streams = new KafkaStreams(topology, config, clientSupplier);
```

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -48,9 +54,14 @@
...
- * The driver can be reused to create new state stores, even across
- * different Kafka Streams applications.
+ *
+ * <p>It is also required to use this as a {@link KafkaClientSupplier} when
+ * creating your Kafka Streams application to ensure that all advanced
+ * Responsive functionality functions properly.</p>
  */
+// TODO(agavra): we should put more thought into this API and consider splitting
+// it up into a "reusable" session class and a "per-streams" driver so that we
+// can properly track resources created by the driver
```
ableegoldman, line 64: +1 for sure

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -159,6 +183,48 @@ public <K, V> Materialized<K, V, WindowStore<Bytes, byte[]>> materialized(
...
+  public Consumer<byte[], byte[]> getRestoreConsumer(final Map<String, Object> config) {
+    return delegateClientSupplier.getRestoreConsumer(config);
+  }
+
+  @Override
+  public Consumer<byte[], byte[]> getGlobalConsumer(final Map<String, Object> config) {
+    config.put(
+        ConsumerConfig.GROUP_ID_CONFIG,
+        this.config.getString(StreamsConfig.APPLICATION_ID_CONFIG) + "-global");
+    config.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
ableegoldman, line 219: Turn off auto-commit?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+ *   them to {@link #subscribe(Collection)}</li>
+ *
+ *   <li>It ignores calls to {@link #seek(TopicPartition, long)} so that it
+ *   can properly use the committed offsets.</li>
+ *
+ *   <li>It allows looking up {@link #position(TopicPartition)} for partitions
+ *   that this consumer does not own by proxying the Admin for those partitions.</li>
+ *
+ *   <li>It returns {@link ConsumerRecords} from {@link #poll(Duration)} that will
+ *   return all records when calling {@link ConsumerRecords#records(TopicPartition)}
```
ableegoldman, line 64: So smooth 😄

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+ *   instead of only the requested partition. This is necessary because of the way that
+ *   the {@link org.apache.kafka.streams.processor.internals.GlobalStateManagerImpl}
+ *   handles restore. Specifically, it will loop partition-by-partition and update
+ *   the store for those partitions. The issue with this is that if we use {@code subscribe}
+ *   in place of {@code assign} we may get events from partitions that were not
+ *   specified in the {@code assign} call. These records will be dropped.</li>
```
ableegoldman, line 70:
```suggestion
 *   specified in the {@code assign} call. These records would otherwise be dropped.</li>
```

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+ *
+ *   <li>It commits before every call to {@link #poll(Duration)} because
```
ableegoldman, line 72: Oof...this could put a LOT of load on the brokers 😬

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+ *   otherwise the Streams implementation for global stores will never commit. We
+ *   should consider changing this to every {@code N} polls, but this would be
+ *   more difficult to test (we could make it configurable).</li>
+ * </ol>
+ *
+ * <p>This class breaks a lot of abstraction barriers, but allows us to
+ * support remote {@link org.apache.kafka.streams.kstream.GlobalKTable}s without
+ * forking Kafka Streams.</p>
```
ableegoldman, line 80: So honest! 😛

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+ */
+public class ResponsiveGlobalConsumer implements Consumer<byte[], byte[]> {
+
+  private final Consumer<byte[], byte[]> delegate;
+  private final Integer defaultApiTimeoutMs;
```
ableegoldman, line 85: why `Integer` (rather than int)?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
agavra: @ableegoldman I made `ResponsiveDriver` implement `KafkaClientSupplier` so they can just pas that right in :)

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  @Override
+  public void unsubscribe() {
+    delegate.unsubscribe();
+  }
+
+  @Override
+  public void assign(final Collection<TopicPartition> partitions) {
+    subscribe(
```
ableegoldman, line 137: I'm a little worried about unintended side effects of repeatedly calling `subscribe` on an ever-changing set of topics...For example, a consumer group's committed offsets will eventually be deleted for topics that no one has subscribed to. Not sure off the top of my head exactly what determines the retention semantics there, I can look into it I guess.

It would be great if we could just store the set of all topics we've seen so far and stay subscribed to all of them, but of course this is thwarted by the GlobalStateManager's ridiculous restoration loop which restores one store at a time. Unfortunately this also makes it all the more likely for us to lose one topic's offsets while restoring state from another, since it seems to literally go through each entire global store, one by one.

Hopefully they only have one global table? I didn't even think to ask them... :/

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+
+  @Override
+  @Deprecated
+  public ConsumerRecords<byte[], byte[]> poll(final long timeoutMs) {
+    return poll(Duration.ofMillis(timeoutMs));
+  }
+
+  @Override
+  public ConsumerRecords<byte[], byte[]> poll(final Duration timeout) {
+    commitSync();
```
ableegoldman, line 153: Man, yikes...at the very least, if we're going to do this in the POC, let's make sure to maximize the number of records we can process each time. 100 is the default `max.poll.records` which is way too low, I'd say 1,000 or 10,000 even 

10,000 could be significant memory pressure though, and unfortunately last I checked (admittedly several years ago) the consumer's size-related configs don't provide a way to limit the fetch buffer memory completely. WDYT about 1,000?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  @Override
+  public void commitAsync(final OffsetCommitCallback callback) {
+    delegate.commitAsync(callback);
+  }
+
+  @Override
+  public void commitAsync(final Map<TopicPartition, OffsetAndMetadata> offsets,
+      final OffsetCommitCallback callback) {
```
ableegoldman, line 191: super nit, but this is terrible formatting. Check your IDE settings! lol

ableegoldman (commented):
Still going through it, but here's the first batch

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
agavra: yeah :/ I was actually specifically going to ping you on this one and see if you had other suggestions.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
agavra: auto code-gen 😆 fixed

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
agavra: this should happen pretty infrequently right? just on restore and then once when the global thread starts up normally per stream app. if we really wanted to avoid some overhead we could check whether or not the topics are already subscribed to and then not make the call to `subscribe`?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalConsumer.java
agavra: 🤷 haha I think this actually is the "right" thing based on Google checkstyle conventions because otherwise it breaks 100 chars. if I were to manually do this I'd put them all on their own lines.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
ableegoldman: Oh...why? Just seems rather sneaky and unintuitive, was there a reason to do it this way?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveGlobalKeyValueBytesStoreSupplier.java
```diff
@@ -0,0 +1,62 @@
...
+      final ScheduledExecutorService executorService
+  ) {
+    this.client = client;
+    this.name = name;
+
+    // we maintain the name without quotes because quotes are
+    // not valid in Kafka topics, but on the other hand they
+    // are necessary to ensure that Cassandra can accept whatever
+    // the client tosses at it
+    awaitTable = client.awaitTable('"' + name + '"', executorService);
```
ableegoldman, line 45: > they are necessary to ensure that Cassandra can accept whatever the client tosses at it

Oh yeah I've been meaning to ask, can you elaborate on why we have to add quotes here? So the table name would be something like `String tableName = "\"store-name\"";`, is that our requirement or Scylla/C*'s? Who does "the client" refer to here?

agavra: it's a SQL thing - if you don't use quotes then (a) it's not case sensitive and (b) it can't use some characters that are valid in Kafka topics but not in SQL tables. Here's an example. I remember this being necessary for some of my tests, though I'll confirm when I get back from my trip.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
agavra: based on our discussions, I assume there will be a time where the ResponsiveDriver needs to know which clients it handed out - this makes that dependency clear. Maybe I'm thinking too far ahead though and we can figure out that problem when we get there and for now we can just make it provide a supplier.

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
```diff
@@ -0,0 +1,180 @@
...
+      client.initializeOffset(tableName, partition);
+
+      open = true;
+
+      context.register(root, (key, value) -> {
+        if (value == null) {
+          delete(new Bytes(key));
+        } else {
+          put(new Bytes(key), value);
+        }
```
ableegoldman, line 103: Should we be worried about the overhead of all these single non-batched writes to C*? I know global stores are small, updated infrequently, etc...but loading data from an existing global table will make the migration/restoration period rough 😬

Can we implement a batching version of `putAll()`? If so, we can pass in a `RecordBatchingStateRestoreCallback` here that restores batches with all the records returned by `poll`. 

You probably noticed the GlobalStateManagerImpl invokes `restoreBatch` but that gets split into individual records depending on which version of the `StateRestoreCallback` is passed in right here (it's kind of a ridiculous API)

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
```diff
@@ -0,0 +1,180 @@
...
+    final byte[] old = get(key);
+    if (old == null) {
+      put(key, value);
+    }
+    return old;
+  }
+
+  @Override
+  public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
+    entries.forEach(kv -> put(kv.key, kv.value));
```
ableegoldman, line 137: See comment about batching above...technically `putAll` itself might never be invoked unless we do it ourselves but we may as well put the batching logic here and then just call it above

ableegoldman (commented):
Just noticed I had some pending comments here...it's possible some are out of date by now. Anyways here's the next batch, some things we might want to discuss when you get back

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
ableegoldman: I mean I'm totally on board with the ResponsiveDriver maintaining a connection to the client supplier, I just think it's a confusing API to present our poor users with. 

(Hope "sneaky and unintuitive" didn't come off as too harsh haha)

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
agavra: Cassandra/Scylla actually do better without batched writes (they are for atomicity, not performance): https://docs.datastax.com/en/cql-oss/3.x/cql/cql_using/useBatch.html so this should be better

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
agavra: > (Hope "sneaky and unintuitive" didn't come off as too harsh haha)

haha, you're saying sneaky isn't a compliment? 😉 just kidding, I see where you're coming from

kafka-client/src/main/java/dev/responsive/kafka/consumer/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Override
+  @Deprecated
+  public ConsumerRecords<byte[], byte[]> poll(final long timeoutMs) {
+    return poll(Duration.ofMillis(timeoutMs));
+  }
+
+  @Override
+  public ConsumerRecords<byte[], byte[]> poll(final Duration timeout) {
+    commitSync();
```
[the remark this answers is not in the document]
agavra: I thought the concern was that atuo-offset committing happens independently of processing, which means that it's possible that we'll commit after reading a batch but before writing it to remote storage. It's OK if the same data gets written twice (we process a batch and write to remote, but don't commit the offset) but it's not OK if we commit the offsets but don't actually process the batch.

I don't see how what you're saying gets around this problem?

kafka-client/src/main/java/dev/responsive/kafka/consumer/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+  private final int defaultApiTimeoutMs;
+  private final Admin admin;
+
+  public ResponsiveGlobalConsumer(
+      final Map<String, Object> config,
+      final Consumer<byte[], byte[]> delegate,
+      final Admin admin
+  ) {
+    super(delegate);
+    final ConsumerConfig consumerConfig = new ConsumerConfig(config);
```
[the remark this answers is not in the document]
agavra: ugh... that constructor is `protected`. no way to disable logging here :/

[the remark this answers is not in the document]
agavra: ugh.. that construct is protected for some unknown reason so there's no way to disable that logging

kafka-client/src/main/java/dev/responsive/kafka/consumer/ResponsiveGlobalConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+      return result == null ? 0 : result.offset();
+    } catch (final InterruptedException | ExecutionException e) {
+      throw new RuntimeException(e);
+    } catch (final TimeoutException e) {
+      throw new org.apache.kafka.common.errors.TimeoutException(e);
+    }
+  }
+
+  /**
+   * A hack that will return all records that were polled when calling
```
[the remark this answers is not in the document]
agavra: I like this approach at all, I wasn't aware of that API!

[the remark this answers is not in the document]
agavra: oh hmm, that'll work for the restore procedure but not for the normal processing (all topic partitions are passed into `assign`)

kafka-client/src/main/java/dev/responsive/kafka/consumer/ResponsiveGlobalConsumer.java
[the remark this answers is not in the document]
agavra: OK I have an idea, let's just do what I'm doing now (commit sync before reading a batch) but use the auto commit ms config to configure how often we do that and disable it from the underlying delegate. that should get us what we want, right?

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
+import org.apache.kafka.common.TopicPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ResponsiveConsumer<K, V> extends DelegatingConsumer<K, V> {
```
agavra, line 32: the only changes to this file were removing the methods that directly call `delegate`'s implementation and make it implement `DelegatingConsumer`

cc @rodesai

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
ableegoldman: FWIW I found this blog series to provide a good (and short!) high-level summary: [post #1](https://foundev.medium.com/cassandra-batch-loading-without-the-batch-keyword-40f00e35e23e) and [post #2](https://foundev.medium.com/cassandra-batch-loading-without-the-batch-the-nuanced-edition-dd78d61e9885)

The 2nd one has some solid memes

agavra: Heh, I had spent quite a bit of time reading up on batching etc.. in scylla/cassandra when researching it for the CAS use case we had so I'm aware of logged/unlogged batching and the limitations around single-partitions/cross-partition writes.

My opinion here is that we shouldn't use batches in hope of improved performance because that's not what they're meant for. There might be situations (as described in post 2) where they will improve the performance, but I would argue that ideally, this kind of performance batching is a client optimization that is provided by the vendor. I see no reason why the Scylla driver can't eventually come up with a "write buffer size" configuration that batches writes to the network, and if we explicitly used Cassandra Batches we would not be able to take advantage of such a feature.

I do agree with your assessment that we should keep an eye out for this, and if global state restoration proves to be an issue we should work with our contacts at Scylla to figure out optimization - but for now I think changing semantics would be premature optimization.

agavra: Also I think we should eventually move away from global partitions being a single partition in the remote store - that just made this PR a bit simpler because I could reuse all the existing client code for things like range scans, but it's not a requirement (we could create an iterator across each partition when doing global lookups, or use `ALLOW_FILTERING`).

agavra: https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/tips/benchmark-tips.html#query-recommendations hmm, maybe I didn't understand this properly, as they do actually recommend batching for performance in this document

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -48,9 +50,14 @@
 /**
  * The {@code ResponsiveDriver} should be instantiated once per JVM
  * and maintains a session and connection to the remote storage server.
- * The driver can be reused to create new state stores, even across
- * different Kafka Streams applications.
+ *
+ * <p>It is also required to use this as a {@link KafkaClientSupplier} when
```
ableegoldman, line 54: @agavra need to update/remove this

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -60,6 +67,7 @@ public class ResponsiveDriver implements StreamsStoreDriver, Closeable {
   private final CqlSession session;
   private final CassandraClient client;
   private final Admin admin;
+  private final StreamsConfig config;
```
ableegoldman, line 70: This is no longer used in this PR

ableegoldman (approved):
Alright I think this is in a good place -- I'm happy to have it merged as-is

PR #38, opened 2023-06-24
Add Kafka Streams and client metrics

ableegoldman:
Baseline set of metrics I would like to include

### DIFF
--- a/etc/otel-jmx.config.yaml
+++ b/etc/otel-jmx.config.yaml
@@ -82,7 +82,7 @@ rules:
      process-ratio:
        metric: kafka.streams.thread.process.ratio
        type: gauge
-       desc: ratio of time spent processing
+       desc: fraction of time spent in process
      commit-rate:
        metric: kafka.streams.thread.commit.rate
        type: gauge
@@ -91,7 +91,7 @@ rules:
      commit-ratio:
        metric: kafka.streams.thread.commit.ratio
        type: gauge
-       desc: ratio of time spent committing
+       desc: fraction of time spent in commit
      commit-latency-avg:
        metric: kafka.streams.thread.commit.latency.avg
        type: gauge
@@ -110,7 +110,7 @@ rules:
      poll-ratio:
        metric: kafka.streams.thread.poll.ratio
        type: gauge
-       desc: ratio of time spent polling
+       desc: fraction of time spent in poll
      poll-records-avg:
        metric: kafka.streams.thread.poll.records.avg
        type: gauge
@@ -119,7 +119,7 @@ rules:
      punctuate-ratio:
        metric: kafka.streams.thread.punctuate.ratio
        type: gauge
-       desc: ratio of time spent punctuating
+       desc: fraction of time spent in punctuate
      task-closed-rate:
        metric: kafka.streams.thread.task.closed.rate
        type: gauge
@@ -137,13 +137,19 @@ rules:
        unit: '{milliseconds}'
 
  # Kafka Streams task metrics
- ## THESE ARE GENERALLY **DEBUG** METRICS AND MAY NOT BE ENABLED
+ ## THESE ARE GENERALLY **DEBUG** METRICS AND MIGHT NOT BE ENABLED
  ## Most Streams metrics below the client and thread level are debug or even trace
  - bean: kafka.streams:type=stream-task-metrics,thread-id=*,task-id=*
    metricAttribute:
      thread: param(thread-id)
      task: param(task-id)
    mapping:
+     ## INFO
+     active-process-ratio:
+       metric: kafka.streams.task.process.ratio
+       type: gauge
+       desc: fraction of time spent processing this task out of all the active tasks
+     ## DEBUG
      process-total:
        metric: kafka.streams.task.process.total
        type: gauge
--- a/etc/otel-jmx.config.yaml
+++ b/etc/otel-jmx.config.yaml
@@ -1,7 +1,6 @@
 ---
 rules:
-  # kafka client metrics
-
+ # Kafka Consumer metrics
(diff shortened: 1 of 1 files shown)


etc/otel-jmx.config.yaml
```diff
@@ -12,25 +11,46 @@ rules:
...
    mapping:
      records-lag-max:
        metric: kafka.streams.records.lag.max
        type: gauge
-       desc: the current lag of the partition
-       unit: '{messages}'
- - bean: kafka.streams:type=stream-thread-metrics,thread-id=*
+       desc: the max lag of all partitions
+       unit: '{records}'
+ - bean: kafka.consumer:type=consumer-coordinator-metrics,client-id=*
```
ableegoldman, line 24: sorry, git diff is a bit messed up -- I didn't remove or change any metrics (other than the minor fixes right above this), just organized them according to the client/level

etc/otel-jmx.config.yaml
```diff
@@ -12,25 +11,46 @@ rules:
        metric: kafka.streams.records.lag
        type: gauge
        desc: the current lag of the partition
-       unit: '{messages}'
+       unit: '{records}'
  - bean: kafka.consumer:type=consumer-fetch-manager-metrics,client-id=*
    metricAttribute:
      clientid: param(client-id)
```
ableegoldman, line 17: @rodesai should (or can) this be `client-id` instead of `clientid`?

etc/otel-jmx.config.yaml
```diff
@@ -43,31 +63,89 @@ rules:
...
- - bean: kafka.streams:type=stream-task-metrics,thread-id=*,task-id=*
+
+ # Kafka Streams thread metrics
+ - bean: kafka.streams:type=stream-thread-metrics,thread-id=*
    metricAttribute:
      thread: param(thread-id)
-     task: param(task-id)
    mapping:
      process-total:
-       metric: kafka.streams.task.process.total
```
ableegoldman, line 52: again, didn't change this, just moved the task-level metric to the end

etc/otel-jmx.config.yaml
```diff
@@ -43,31 +63,89 @@ rules:
...
+       unit: '{tasks-per-second}'
      blocked-time-ns-total:
        metric: kafka.streams.thread.blocked.time.total.ns
        type: gauge
        desc: total time the stream thread was blocked
        unit: '{nanoseconds}'
- - bean: kafka.streams:type=stream-thread-metrics,thread-id=*
+     thread-start-time:
+       metric: kafka.streams.thread.start.time
+       type: value
```
ableegoldman, line 135: Is this a valid type? Not sure what the options are, but this isn't a "gauge" unless that has a different meaning here -- it's not like this will be changing

etc/otel-jmx.config.yaml
```diff
@@ -43,31 +63,89 @@ rules:
...
+       desc: the time the kafka streams thread was started
+       unit: '{milliseconds}'
+
+ # Kafka Streams task metrics
+ ## THESE ARE GENERALLY **DEBUG** METRICS AND MAY NOT BE ENABLED
```
ableegoldman, line 140: I've spoken with Rohan about this already, but imo we may as well consider anything below INFO a non-starter

etc/otel-jmx.config.yaml
```diff
@@ -12,25 +11,46 @@ rules:
...
+ ## Note the translation from 'client-id' to 'application-id' in our tags
+ - bean: kafka.streams:type=stream-metrics,client-id=*
+   metricAttribute:
+     application-id: param(client-id)
+   mapping:
+     state:
+       metric: kafka.streams.state
+       type: gauge
+       desc: the current state of this Streams application
+       unit: '{KafkaStreams.State}'
```
ableegoldman, line 51: Any opinion on whether to give this a unit at all?

etc/otel-jmx.config.yaml
```diff
@@ -43,31 +63,95 @@ rules:
...
-   metricAttribute:
-     thread: param(thread-id)
-   mapping:
+     process-rate:
+       metric: kafka.streams.thread.process.rate
+       type: gauge
+       desc: rate of records processed
+       unit: '{records-per-second}'
+     process-ratio:
+       metric: kafka.streams.thread.process.ratio
```
ableegoldman, line 83: Left all the ratios unit-less but open to suggestions

agavra (approved):
LGTM, a screenshot of a JMX console or jterm session to confirm we don't have any bugs here might be worthwhile testing

etc/otel-jmx.config.yaml
```diff
@@ -12,25 +11,46 @@ rules:
        metric: kafka.streams.records.lag
        type: gauge
        desc: the current lag of the partition
-       unit: '{messages}'
+       unit: '{records}'
```
rodesai, line 14: I'm going to change to "offsets" - make sense @ableegoldman

etc/otel-jmx.config.yaml
rodesai: I suppose it could, but we'd need to change the policy as well. Let's do that in a follow up iteration.

etc/otel-jmx.config.yaml
rodesai: I don't think it matters.

etc/otel-jmx.config.yaml
rodesai: that's fine - I think the units are just informational anyway

etc/otel-jmx.config.yaml
rodesai: no - I think we can just use the gauge type.

etc/otel-jmx.config.yaml
ableegoldman: yeah that's fine, was just thinking we should standardize units but I don't think it really matters too much

PR #39, opened 2023-06-26
Refactor ManagedApplication into ManagedDeployment and ManagedStatefulSet.

apurvam:
Follow up to #31


agavra (approved):
refactor makes sense to me, didn't look at the logic 🚢 🇮🇹

PR #40, opened 2023-06-27
log operator logs to our logging pipeline

rodesai:
This patch sets operator deployments up to log to our logging pipeline:
- packages operator with log4j2, which ships with a maintained kafka appender
- adds a customized login module for authenticating with ccloud brokers. It reads username/password from a separate file (rather than the log4j properties file). This way we can put the api key/secret in a k8s secret.
- docker run script sets up a log4j properties file that logs to kafka, if configured to do so.
- helm chart mounts kafka secret and sets up environment variables so that log4j with kafka can be used.

Tested using local operator in kind running against a real observability stack"

### DIFF
--- a/operator/build.gradle.kts
+++ b/operator/build.gradle.kts
@@ -34,6 +34,8 @@ dependencies {
     implementation(libs.bundles.commons)
     implementation(libs.bundles.logging)
 
+    implementation(libs.kafka.clients)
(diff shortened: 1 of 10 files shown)


rodesai:
merging so I can test the official chart. Would still love 👁️ 👁️

PR #41, opened 2023-06-27
Make sure the Admin client supplied to each global consumer is closed

ableegoldman:
Minor thing I missed before -- if we create a new admin with the client supplier just to be passed in to the global consumer, we need to make sure that admin client also gets closed.


agavra (approved):
good catch! thanks :)

PR #42, opened 2023-06-27
Proposal for new layered API and strict StreamsConfigs

ableegoldman:
I've been mulling over something like this for a while, and saw that Almog left a TODO with pretty much exactly what I had in mind in his last PR so I decided to go for it. The high-level idea is to split up the monolith driver we have today into separate concerns scoped according to their lifetime and resource management, while also hopefully making it easier for new users to see at a glance everything they need to change in their code with some better API compatibility guarantees.

Essentially they connect to a reusable/stateless Responsive client session for the process lifetime, which provides a `StreamsApplicationDriver` for managing everything scoped to one Streams application. This lets us control what ultimately gets passed in to the `KafkaStreams` such as our custom kafka client supplier and StreamsConfig that's been validated with necessary overrides applied.

The `StreamsApplicationDriver` (new interface/class) in turn returns a `StreamsStoreDriver` (old interface, implementation moved to new class). The store driver isolates everything related to state, and is where various StoreSuppliers and Materialized objects can be obtained.

Strictly speaking only the config validation/overrides are absolutely necessary for the first trial runs, but it just seems like a nice thing to do for our potential customers to land on a relatively stable API up front. I think this gives us more room to make changes under the hood without asking them to rewrite their code after the initial migration, but open to feedback or suggestions.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
@@ -33,48 +33,31 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.util.Map;
 import java.util.Properties;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
-import org.apache.kafka.clients.admin.Admin;
-import org.apache.kafka.common.config.TopicConfig;
 import org.apache.kafka.common.config.types.Password;
-import org.apache.kafka.common.utils.Bytes;
-import org.apache.kafka.streams.kstream.Materialized;
-import org.apache.kafka.streams.state.KeyValueBytesStoreSupplier;
-import org.apache.kafka.streams.state.KeyValueStore;
-import org.apache.kafka.streams.state.WindowBytesStoreSupplier;
-import org.apache.kafka.streams.state.WindowStore;
+import org.apache.kafka.streams.StreamsConfig;
+
 
 /**
  * The {@code ResponsiveDriver} should be instantiated once per JVM
  * and maintains a session and connection to the remote storage server.
  */
-// TODO(agavra): we should put more thought into this API and consider splitting
-// it up into a "reusable" session class and a "per-streams" driver so that we
-// can properly track resources created by the driver
-public class ResponsiveDriver implements StreamsStoreDriver, Closeable {
+public class ResponsiveDriver implements Closeable {
 
-  private static final Map<String, String> CHANGELOG_CONFIG = Map.of(
-      TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_DELETE);
-
-  private final ScheduledExecutorService executor = new ScheduledThreadPoolExecutor(2);
   private final CqlSession session;
-  private final CassandraClient client;
-  private final Admin admin;
 
   /**
-   * @param props the properties to pass in
-   * @return a new {@code ResponsiveDriver} and opens connections to remote Responsive servers
+   * @param props the Responsive properties for this session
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -33,48 +33,31 @@
...
  * and maintains a session and connection to the remote storage server.
  */
-// TODO(agavra): we should put more thought into this API and consider splitting
-// it up into a "reusable" session class and a "per-streams" driver so that we
-// can properly track resources created by the driver
-public class ResponsiveDriver implements StreamsStoreDriver, Closeable {
-
-  private static final Map<String, String> CHANGELOG_CONFIG = Map.of(
-      TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_DELETE);
+public class ResponsiveDriver implements Closeable {
```
ableegoldman, line 44: Maybe should be renamed to `ResponsiveSession` or something like that, but didn't want to add unnecessary review burden. Can save any name changes for just before merging

agavra (commented):
I have a love/hate relationship with API design because I feel like it's a prime candidate for bikeshedding... but I also feel like it's important to align on goals since it's the first thing our customers will see. 

As a suggestion, I think we should first write the quickstart/migration guide to look the way we want it to look and only then put up a PR for actually making the API changes. That will help us focus this conversation from the user's perspective instead of from ours.

That being said, bikeshedding below!

---------
```
.
.
.
.
.
.
.
.
```
-------
Starting from first principles, here would be my goals for our API design:

1. I want the perception of Responsive for Kafka Streams to be a minor surgery that plugs in to your Streams app and "just works" with minimal code changes.
2. We should not make assumptions on how users have written their streams app - we should be as flexible as possible to avoid requiring structural changes to their code.
3. Prefer config to code so users can easily blue/green test us

I think that what you've proposed here is great if the goal is to make new apps as streamlined as possible, but (and this is subjective so there's certainly room for debate) I don't think its ideal for the API goals I've listed above. The way I think about it is: if we're writing a quickstart, the impression should be that they didn't have to learn many new concepts and the structure of their code remains identical.

Specifically:
- The `ResponsiveApplicationDriver` method "feels" like a big deal. It's intercepting the main entry point (`kafkaStreams(topology)`) into their application and gives us freedom to do really whatever we want to their app as opposed to plugging in to their code. It flips the API that was there originally, where the entry point remains the same. An example of where this can be annoying is that there are many different Kafka Streams constructors (Topology, Properties, KafkaClientSupplier, Time) in various permutations: they might have weird ways of creating each of these that they pass in, but this PR has an opinion about how it should be created.
- I know this was something I suggested, but I think we should give more thought about how we can split the concept of a global session and an application without creating overhead for the simple case. This pattern was something that felt like an annoyance when dealing with the C* driver API (in their old API you have to first create a session, then create a client from the session). Instead of keeping the simple things simple and the complicated things possible, it sacrifices ease of use for the common case (single kafka streams app) to make the complicated case (multiple streams apps) streamlined.
- For configuration, I'm not a big fan of changing their passed-in properties (again, hypocrisy! I know I did this in `GlobalConsumer` code...). If they set something (such as `num.standby.replicas`) in their properties file it should be respected (namely we should throw an exception indicating that isn't supported and they should set it to 0). I think overriding what they've set can cause confusion. Instead, I would consider having a `ResponsiveStreamsConfig` that `extends StreamsConfig` and additionally validates the config (changing defaults is OK).

ableegoldman:
Closing this as we've settled on a much better approach that aligns more closely to what users would be doing in vanilla Kafka Streams.

Kicked off a good discussion though. And thanks for the bikeshedding Almog -- even if we don't agree on every single point you brought up, that was valuable feedback and it led to an outcome I think we're all much happier with (even if parts are still being debated 😄 )

PR #43, opened 2023-06-28
add more logging

agavra:
minor commit to add more logging to the state stores - we should still integrate with the the datastax driver query logger so that we can have more details there. There's also server-side logging we can enable.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
+++ b/kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
@@ -49,6 +49,8 @@ import java.util.concurrent.ScheduledExecutorService;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.state.KeyValueIterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * {@code CassandraClient} wraps a {@link CqlSession} with utility methods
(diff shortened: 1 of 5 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveWindowStore.java
```diff
@@ -134,21 +134,14 @@ public void init(final ProcessorContext context, final StateStore root) {
   @Override
   public void init(final StateStoreContext context, final StateStore root) {
     try {
+      LOG.info("Initializing state store {} with remote table name {}", name, tableName);
       this.context = asInternalProcessorContext(context);
       this.recordCollector = asRecordCollector(context);
 
       partition = context.taskId().partition();
-      if (partition == 0) {
```
rodesai, line 141: was this unintentional? Or are we removing this because Scylla supports LWT?

agavra: yeah that was intentional - probably should've split it up into two PRs but this was already done for the non-windowed tables. It was only AWS Keyspaces that didn't support it.

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
```diff
@@ -58,6 +60,8 @@
  */
 public class CassandraClient {
 
+  private static final Logger LOG = LoggerFactory.getLogger(CassandraClient.class);
```
rodesai, line 63: We seem to be using LOG some places and LOGGER elsewhere. It seems consistent amongst the storage stuff though.

agavra: I prefer `LOG` because it's three fewer chars for long log lines 😆 I don't really care though if it's inconsistent

rodesai (approved):
LGTM

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
rodesai: we need emoji support in the next java. 🪵 .info("hello")

PR #44, opened 2023-06-28
support more store names

agavra:
this PR will replace all characters that are passed in as the store name that are invalid in cassandra with `_`. it also removes the `"` (cc @ableegoldman 😆) and just deals with case insensitivity (which I think was the reason for adding them in first place).


rodesai (approved):
LGTM. Up to you if you want to escape '_' first

PR #47, opened 2023-06-29
ensure source topic optimization off

rodesai:


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
```diff
@@ -83,6 +84,7 @@ public void init(final ProcessorContext context, final StateStore root) {
   public void init(final StateStoreContext context, final StateStore root) {
     try {
       LOG.info("Initializing global state store {}", name);
+      StoreUtil.validateTopologyOptimizationConfig(context.appConfigs());
```
agavra, line 87: I don't think we actually need it in this one as the global store doesn't have the concept of a topology as far as I'm aware

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
agavra: not blocking, we can address this in a follow-up when we support the optimization

PR #48, opened 2023-06-30
Add timestampedKeyValueStoreBuilder to API

ableegoldman:
Just the timestamped key-value StoreBuilder which sets the cleanup policy to `[delete]` and makes sure this doesn't get overridden accidentally afterwards

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
@@ -31,6 +31,7 @@ import dev.responsive.utils.SessionUtil;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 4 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -152,6 +160,44 @@ public KeyValueBytesStoreSupplier globalKv(final String name) {
...
+      ) {
+        final String cleanupPolicy = config.get(TopicConfig.CLEANUP_POLICY_CONFIG);
+
+        if (cleanupPolicy == null || cleanupPolicy.equals(TopicConfig.CLEANUP_POLICY_COMPACT)) {
+          log.debug("Overriding the changelog topic cleanup.policy from compact to delete");
+          final Map<String, String> configCopy = new HashMap<>(config);
+          configCopy.put(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_DELETE);
+          return super.withLoggingEnabled(configCopy);
+
+          // We allow both [delete] and [compact, delete]
```
agavra, line 188: what's the benefit to supporting `[compact, delete]` if we're manually truncating? also does that work (can you truncate a `[comapct, delete]` topic)?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
```diff
@@ -152,6 +160,44 @@ public KeyValueBytesStoreSupplier globalKv(final String name) {
...
+    return new TimestampedKeyValueStoreBuilder<>(
+        timestampedKv(name),
+        keySerde,
+        valueSerde,
+        Time.SYSTEM) {
+      private final Logger log =
+          new LogContext(String.format("store-builder [%s]", name)).logger(ResponsiveDriver.class);
+
+      @Override
+      public StoreBuilder<TimestampedKeyValueStore<K, V>> withLoggingEnabled(
```
agavra, line 177: I still think we should throw an exception here instead of silently overriding - otherwise we risk accidentally truncating/deleting a topic that they don't want deleted, which would be something that (even if it's their "fault" for not understanding our code) would be something really hard to recover trust from.

agavra (commented):
thanks for the quick turnaround! comments inline - I think the most important part is the API itself that we can just ask people to use (`timestampedKeyValueStoreBuilder(...)`) instead of overriding the `withLoggingEnabled` method; we should get that merged first and then discuss the safety part of it.

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveDriver.java
ableegoldman: > can you truncate a [compact, delete] topic

That's actually a good question -- I misremembered the thing about not being able to truncate compacted topics (actually I thought I'd tested that, it definitely worked locally -- don't suppose you have the exception/error msg on hand or saved anywhere? Would be good to verify the case where both are set)

agavra: I just tested it with CCloud - it works with `compact,delete` but fails with the following error when it's just `compact`:
```
org.apache.kafka.common.errors.PolicyViolationException: Request parameters do not satisfy the configured policy.
...
Caused by: org.apache.kafka.common.errors.PolicyViolationException: Request parameters do not satisfy the configured policy.
```

PR #49, opened 2023-06-30
dont use OnCommitCallback

agavra:
the on commit callback is only called when we flush anyway, so this ultimately caused us to call the `CommitBuffer#flush` method twice back to back leading to confusing statements cluttering the log (empty buffers)


apurvam (approved):
I think you should probably wait for another review if you want validation of the reasoning.

rodesai (approved):
LGTM

PR #53, opened 2023-07-01
API: Pt.3, connect static stores with Responsive clients

ableegoldman:
Next part to the main API refactor: I stubbed in a static `ResponsiveStores` factory with minimal changes to the Streams `Stores` since there's some ongoing discussion here. Hopefully this is basic enough to be non-controversial, and we can move forward on the overall refactor to unblock the other work depending on this.

This PR follows Pt. 2 but includes a modified version of the changes in Pt. 1 to remove everything controversial (again, I hope)


ableegoldman:
FYI don't review this just yet, I'm still responding to everyone's comments on the first 2 PRs but just figured I'd open this now to showcase what it will look like all hooked up

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * A factory for creating Kafka Streams state stores on top of a Responsive storage backend.
+ * Use these when building your {@link org.apache.kafka.streams.Topology} to easily swap in
+ * Responsive stores wherever state is used.
+ *
+ * See {@link org.apache.kafka.streams.state.Stores} for instructions on how to plug in custom
+ * state stores and configure them.
+ */
+public final class ResponsiveStores {
```
ableegoldman, line 25: @agavra I tried to back out the API changes you took offense to so this PR isn't blocked on that discussion, but let me know if there's anything still in here you're uncomfortable with that I missed. The only change from the way it is today, other than being static, is to provide StoreBuilders (only) rather than StoreSuppliers. Either way they're both offered in `Stores` so I figured this would be ok?

responsive-test-utils/src/test/java/dev/responsive/kafka/api/TopologyTestDriverExampleTest.java
```diff
@@ -45,7 +45,7 @@ public void shouldRunWithoutResponsiveConnection() {
     props.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class);
     props.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class);
 
-    final Topology topo = topology(new TestStoreDriver());
+    final Topology topo = topology();
```
ableegoldman, line 48: I realize this test doesn't really do anything anymore, we'll need to rethink the TestStoreDriver approach now that we have static store factories. We can put this on the list of things to discuss

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * Create a {@link Materialized} that can be used to build a Responsive {@link KeyValueStore}
+   * and materialized in the DSL. If using the low-level Processor API, use
+   * {@link #keyValueStoreBuilder} instead.
+   *
+   * @param name the store name
+   * @return a Materialized configuration that can be used to build a key value store with the
+   *         given options that uses Responsive's storage for its backend
+   */
+  public static <K, V> Materialized<K, V, KeyValueStore<Bytes, byte[]>> materialized(
```
agavra, line 160: what did you think about moving these into `ResponsiveMaterialized` class and naming them the same as existing methods (e.g. `as`)? That way (again) they can just `s/Materialized/ResponsiveMaterialized/g` during the migration

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -54,7 +56,7 @@
  * close callbacks on the returned clients). As these calls are not on the performance path we
  * rely on coarse-grained locks around sections reading/writing shared state.
  */
-public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier, Configurable {
+public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier {
```
agavra, line 59: isn't it necessary to make this `Configurable` if they want to pass it in using the `default.client.supplier` config?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
```diff
@@ -93,8 +89,12 @@ public void init(final StateStoreContext context, final StateStore root) {
       // to be more evently distributed and take the hit only on range
       // queries
       partition = 0;
+      final SharedClients sharedClients = new SharedClients(context.appConfigs());
+      client = sharedClients.cassandraClient;
+
+      final RemoteMonitor monitor = client.awaitTable(name.cassandraName(), sharedClients.executor);
```
agavra, line 95: the nice thing about sharing this across threads is that if there are many partitions only one of them needs to satisfy the condition and then we'll stop querying scylla and we're not hammering the server with metadata requests

thoughts on adding this to the shared clients?

agavra (approved):
Thanks! This is a huge improvement over the original API

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -131,24 +134,41 @@ private static StreamsConfig verifiedStreamsConfigs(
...
           TASK_ASSIGNOR_CLASS_OVERRIDE
       );
       LOG.error(errorMsg);
       throw new ConfigException(errorMsg);
-    } else {
-      propsWithOverrides.put(
-          InternalConfig.INTERNAL_TASK_ASSIGNOR_CLASS, TASK_ASSIGNOR_CLASS_OVERRIDE
+    }
+
+    if (!TASK_ASSIGNOR_CLASS_OVERRIDE.equals(o.toString())) {
```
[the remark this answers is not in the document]
ableegoldman: They're separate because I wanted to give different error messages. But in light of this, doesn't matter:

> I'm OK overriding this if they passed in null

Ah must have misremembered the outcome of that discussion 😅 Glad to hear though

agavra (commented):
👍

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -0,0 +1,235 @@
...
+public final class ResponsiveStores {
+
+  /**
+   * See for example {@link Stores#inMemoryKeyValueStore(String)}
+   *
+   * @param name the store name
+   * @return a supplier for a key-value store with the given options
+   *         that uses Responsive's storage for its backend
+   */
+  public static KeyValueBytesStoreSupplier keyValueStore(final String name) {
```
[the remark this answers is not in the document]
agavra: also yes... I am aware that it's technically a "persistent" store since we return `false` for that, but for all intents and purposes (and as far as the user cares) it is persistent.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
ableegoldman: >  keeping the API with the supplier since we can use the supplier so we can eventually provide different implementations optimized for different use cases.

Personally I see this as an argument for moving towards a universal `KeyValueStoreOptions` rather than aligning to the existing Stores API...we can easily add new optional configurations that way, and it's easy to tell what's a variation on a fundamental store type vs different actual store types. This is part of what's so confusing about the existing `Stores` API and wild proliferation of different StoreSupplier static methods. 

Opens up so many questions...when can I use which store supplier? I mean just focusing on one aspect there's all this (not even getting into the new `versioned` store variants starting to roll out):
1. Are `#persistentKeyValueStore` and `#persistentTimestampedKeyValueStore` interchangeable? 
2. Are `#persistentKeyValueStore` and `#inMemoryKeyValueStore` interchangeable? 
3. Are `#persistentTimestampedKeyValueStore` and `#inMemoryKeyValueStore` interchangeable?

Want to guess? The answer may surprise you 😜

ableegoldman: Answer:
1. NEVER! Whether using the DSL or PAPI, you absolutely have to use the correct one -- this means somehow knowing to use the timestamped version for correct semantics in the DSL (ie via Materialized), and making sure to use the correct supplier API _only_ with its the corresponding builder API 
2. ALWAYS! Whether using the DSL or PAPI, you can actually swap out a `#persistentTimestampedKeyValueStore` for a `#inMemoryKeyValueStore` at any time or place in the code, or vice versa, it's all compatible
3. SOMETIMES! You can pass an `#inMemoryKeyValueStore` into either the timestamped or regular builder in the PAPI, but can't swap in a regular `#persistentKeyValueStore` to the timestamped builder. In the DSL, changing between these two will give you different semantics 😱

agavra: ah... yikes 😨

ableegoldman: (This is why I'm leaning so heavily into providing only builder and materialized APIs for each type of store -- not only is there no actual need for the direct store suppliers*, it's a nightmare trying to figure out which one can/should be passed into which builder/materialized. Because the runtime is opinionated about this, but the `Stores` API is not)

* the one notable exception here being stream-stream  joins, but those are a nightmare to configure as well for a different reason, though the underlying cause is the same: the runtime is opinionated about there being correct values to the parameters, but the API makes this look the user's choice.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
ableegoldman: 😞

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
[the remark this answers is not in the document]
ableegoldman: But this is a lie... semantically we're closer to the in-memory variants of the store suppliers (related to https://github.com/responsivedev/responsive-pub/pull/53#discussion_r1251334656)

(Also why should they get a compile error for IM stores?)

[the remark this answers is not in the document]
ableegoldman: If the goal is to make migration stupid simple, simply doing a find-and-replace to change the method name seems well within that

Not to mention, and this is going to be part of the larger discussion so I won't get into it too much, but the (admittedly clever) `s/Stores/ResponsiveStores/g` swap really will only help what is likely to be a very small minority of users, who are already plugging in custom stores everywhere they possibly have state...

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
ableegoldman: I actually did think about it! Glad you called this out, I meant to leave a comment asking you but forgot...I was wondering if this would have any nontrivial side effect or is more of a nice-to-have optimization?

The trouble is that with the new pattern, the first time we get access to both the `cassandraName` (ie store name) and the configs (with shared clients) is here in `#init`. So we can't do this any earlier, right?

Beyond the scope of this particular PR, but I was wondering if we could instead move the table creation outside of the state stores entirely -- for example, after creating the other shared clients in  `ResponsiveKafkaStreams` we would also parse the `Topology` for all the store names and create a RemoteMonitor for each one to stash alongside the other shared clients. Does that make sense? Maybe a followup PR?

agavra: I see that makes sense, let's punt the optimization to a future PR. Not urgent for now.

ableegoldman: I filed a ticket to revisit this https://www.notion.so/responsivedev/Create-C-tables-during-shared-client-creation-b28c4c54f3e745e4a7b4a1a4ef2d3294?pvs=4

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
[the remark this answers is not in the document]
agavra: heh, I like the elegance of a true `s//g` but I'll concede here - it's probably fine to keep it as is. I won't block the PR on it.

>  will only help what is likely to be a very small minority of users

I guess my sample size is 1 ;) but for the others I think the eventual path is to support `dsl.default.store` to be a custom store implementation. Also wouldn't this apply to anyone using the Processor API?

responsive-test-utils/src/test/java/dev/responsive/kafka/api/TopologyTestDriverExampleTest.java
ableegoldman: For reference, here's a ticket to address this: https://www.notion.so/responsivedev/Write-a-test-driver-client-with-in-memory-backend-d2e674cf264147d0b8fa83934260fda7?pvs=4

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
ableegoldman: I don't think there is a `default.client.supplier` config? At least not in Streams, so the way it is right now it's never getting configured at all unless explicitly invoked.

Anyways part of this PR and the new API improvements are that we pass in our client supplier ourselves rather than making the user pass it in, however that would be. Whether or not there's a config for this, you can definitely pass it in via the `KafkaStreams` constructor, so that's what we're now doing in `ResponsiveKafkaStreams`

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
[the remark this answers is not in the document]
ableegoldman: > the eventual path is to support `dsl.default.store` to be a custom store implementation.

Yeah I 100% agree this should be our API #goalz -- or something like it that also applies to the PAPI.

> Also wouldn't this apply to anyone using the Processor API?

Strict PAPI users yes, but in my experience most users who touch the PAPI still do it via transformers -- ie they use a mix of DSL and PAPI. This is what Metronome does IIRC. And we don't want them to miss plugging in our stores to some random stateful DSL operator elsewhere in their topology. 

I have no idea what the overall userbase breaks down into w.r.t DSL vs PAPI vs mix, but if I had to guess I'd put money on DSL or mix heavily outweighing PAPI-only. As to how many of the "mix" category actually has state in the DSL or the PAPI or both, well, I wouldn't put money on that bet because I really have no idea 😄

PR #55, opened 2023-07-03
use context to implicitly create global stores

agavra:
`ResponsiveStore` is now just a wrapper that delegates to either `ResponsivePartitionedStore` or `ResponsiveGlobalStore` depending on the context passed in during initialization.

This allows us to remove the `globalX` methods in our API and match the existing streams apis.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
@@ -1,0 +1,292 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.store;
+
+import static org.apache.kafka.streams.processor.internals.ProcessorContextUtils.asInternalProcessorContext;
+import static org.apache.kafka.streams.processor.internals.ProcessorContextUtils.changelogFor;
+
+import com.datastax.oss.driver.api.core.cql.BoundStatement;
+import dev.responsive.db.CassandraClient;
+import dev.responsive.kafka.api.InternalConfigs;
+import dev.responsive.kafka.clients.SharedClients;
+import dev.responsive.model.Result;
+import dev.responsive.utils.RemoteMonitor;
+import dev.responsive.utils.StoreUtil;
+import dev.responsive.utils.TableName;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.TimeoutException;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.errors.ProcessorStateException;
+import org.apache.kafka.streams.processor.ProcessorContext;
+import org.apache.kafka.streams.processor.StateStore;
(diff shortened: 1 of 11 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
```diff
@@ -100,176 +73,110 @@ public void init(final ProcessorContext context, final StateStore root) {
...
-          admin,
-          PLUGIN
-      );
-
-      open = true;
-
-      context.register(root, buffer);
-    } catch (InterruptedException | TimeoutException e) {
-      throw new ProcessorStateException("Failed to initialize store.", e);
+    if (context instanceof GlobalProcessorContextImpl) {
```
ableegoldman, line 76: So slick! Also I like the new names -- I imagine a future where users don't think in terms of Kafka Streams "global" vs "local" stores but rather "full topic" vs "partitioned" views. Ultimately there's just one physical table but your processors can subscribe to a read-only "full view" or a read-write "partition view"

Going even further we can have IQ be more natively and intuitively integrated, like it's just a regular view that's read-only and either "full", "partition", or "single-key" (point lookup)

Maybe that's the real insight here...what is IQ if not an external-facing global table? 😈

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
```diff
@@ -16,57 +16,36 @@
...
-  private final Position position;
-
-  private boolean open;
-  private CommitBuffer<Bytes> buffer;
 
-  @SuppressWarnings("rawtypes")
-  private InternalProcessorContext context;
-  private int partition;
+  // Visible for testing
+  KeyValueStore<Bytes, byte[]> delegate;
```
ableegoldman, line 48: Should there be a package-private getter? Makes me uncomfortable to expose the real thing like this 😅

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
agavra: hah! I didn't even think about it that deeply when I named it. "local" just didn't feel right so I called it `Partitioned` but that's a good point

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
agavra: 🙈 I have no idea what possessed me to do it this way

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
```diff
@@ -100,176 +73,110 @@ public void init(final ProcessorContext context, final StateStore root) {
...
 
   @Override
-  public KeyValueIterator<Bytes, byte[]> all() {
-    return new LocalRemoteKvIterator<>(
-        buffer.all(),
-        client.all(name.cassandraName(), partition),
-        PLUGIN
-    );
+  @Evolving
+  public <R> QueryResult<R> query(final Query<R> query, final PositionBound positionBound,
```
ableegoldman, line 133: nit: formatting

(maybe this is another thing I just assumed, but we're standardizing on separate lines per parameter for long lines, right? If not, that would be my vote)

agavra: 👍 yes we are, these are just auto-generated and sometimes IntelliJ doesn't respect my style guide... generally though I feel that if checkstyle doesn't have a check for it we shouldn't enforce it on PRs - so either we find a way to enforce it with a custom checkstyle or we're just happy go lucky the way we are

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
```diff
@@ -0,0 +1,293 @@
...
+import org.apache.kafka.streams.processor.internals.RecordCollector;
+import org.apache.kafka.streams.processor.internals.RecordCollector.Supplier;
+import org.apache.kafka.streams.query.Position;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.internals.StoreQueryUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ResponsivePartitionedStore implements KeyValueStore<Bytes, byte[]> {
```
agavra, line 52: no changes here from the old `RepsonsiveStore` just renamed

PR #57, opened 2023-07-04
map kafka partitions into multiple remote partitions

agavra:
commit buffer now flushes to N partitions instead of just 1 partition - the general strategy is to map the original record to `hash(key) % original_partition * explode_factor + hash(key) % explode_factor`. The first part of this formula "maps" the original partition to a new space (explodes it) and the second part finds which sub-partition within the exploded partition the key belongs to. 

the actual flushing mechanism will still start a LWT with the remote server and store the same offset for every sub-partition, even if no writes came in to that sub-partition.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
@@ -64,6 +64,14 @@ public class ResponsiveDriverConfig extends AbstractConfig {
       + "responsive server. This applies both to metadata requests and query execution.";
   private static final long REQUEST_TIMEOUT_MS_DEFAULT = 5000L;
 
+  // ------------------- other configurations ---------------------------------
+  public static final String PARTITION_EXPLODE_FACTOR_CONFIG = "responsive.partition"
+      + ".explode.factor";
+  public static final String PARTITION_EXPLODE_FACTOR_DOC = "Explodes kafka partitions "
+      + "into remote partition counts";
+  public static final int PARTITION_EXPLODE_FACTOR_DEFAULT = 31;
+
+  public static final String INTERNAL_PARTITIONER = "__internal.responsive.partitioner__";
 
   // ------------------ required StreamsConfig overrides ----------------------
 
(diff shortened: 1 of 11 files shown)


kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
```diff
@@ -64,6 +64,12 @@ public class ResponsiveDriverConfig extends AbstractConfig {
       + "responsive server. This applies both to metadata requests and query execution.";
   private static final long REQUEST_TIMEOUT_MS_DEFAULT = 5000L;
 
+  // ------------------- other configurations ---------------------------------
+  public static final String PARTITION_EXPLODE_FACTOR_CONFIG = "responsive.partition"
+      + ".explode.factor";
+  public static final String PARTITION_EXPLODE_FACTOR_DOC = "Explodes kafka partitions "
+      + "into remote partition counts";
+  public static final int PARTITION_EXPLODE_FACTOR_DEFAULT = 31;
```
ableegoldman, line 72: why 31?

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
```diff
@@ -64,6 +64,12 @@ public class ResponsiveDriverConfig extends AbstractConfig {
       + "responsive server. This applies both to metadata requests and query execution.";
   private static final long REQUEST_TIMEOUT_MS_DEFAULT = 5000L;
 
+  // ------------------- other configurations ---------------------------------
+  public static final String PARTITION_EXPLODE_FACTOR_CONFIG = "responsive.partition"
```
ableegoldman, line 68: Can we keep configs on the same line and just put in a LineLength suppression for this class? This is one of the few cases it's important to keep everything together, imo

agavra: +1 that's a great idea. this drove me nuts but for some reason I didn't think of the sensible thing...

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
agavra: ah, originally I was using an algorithm that wouldn't work if you exploded by the same number as partitions so I wanted to choose an unlikely number. that's no longer a limitation. Maybe I'll change this config actually so that it'll be a target number of partitions instead of a factor.

PR #58, opened 2023-07-05
API: Pt.4, add ResponsiveTopologyTestDriver outline

ableegoldman:
Needs a followup to finish implementing the `CassandraClientStub`


responsive-test-utils/src/test/java/dev/responsive/kafka/api/TopologyTestDriverExampleTest.java
```diff
@@ -45,7 +44,7 @@ public void shouldRunWithoutResponsiveConnection() {
     props.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class);
 
     final Topology topo = topology();
-    final TopologyTestDriver driver = new TopologyTestDriver(topo, props);
+    final TopologyTestDriver driver = new ResponsiveTopologyTestDriver(topo, props);
```
agavra, line 47: we should see if we can fail creation of a responsive store if you're running via the normal topology test driver

responsive-test-utils/build.gradle.kts
```diff
@@ -20,6 +20,12 @@ plugins {
 
 dependencies {
     implementation(project(":kafka-client"))
+    implementation(libs.kafka.streams.test.utils)
+    implementation(libs.kafka.clients) {
+        artifact {
+            classifier = "test"
+        }
```
agavra, line 27: what does this do? are we depending on the test module only of the kafka.clients?

agavra (approved):
nice!

responsive-test-utils/build.gradle.kts
ableegoldman: yeah, it's for the MockAdminClient -- seems to be published in a separate jar in https://repo.maven.apache.org/maven2/org/apache/kafka/kafka-clients/3.4.0/

responsive-test-utils/src/test/java/dev/responsive/kafka/api/TopologyTestDriverExampleTest.java
ableegoldman: That would be nice, but I'm not sure there's a good way to do that :/

The only thing I can think of, and wouldn't even work for all cases, is to check the configs during the store's `init` and if we see that it contains the known dummy values the TTD stubs in for `bootstrap.servers` and `application.id` but doesn't have our stubbed test-driver version of the cassandra client, then we can give a more informative error message

But that's a lot of brittle code just to give a slightly better error message and not even help in all cases (most people probably don't create separate, dummy configs just for the test I would guess)

Any ideas?

ableegoldman: Maybe we could settle for something in between, for example: if we go to retrieve the CassandraClient in the TTD and don't find one, say something in the error message like "did you forget to use the Responsive version of KafkaStreams or TopologyTestDriver?"

Actually we probably should already be saying something like this...not sure exactly how to word that in a professional way though 😂 Or whether we should mention just KafkaStreams or also the TTD in the error message?

PR #59, opened 2023-07-06
optimize restore and support compacted changelogs

rodesai:
optimize restore and support compacted changelogs

- When restoring, fetch the last committed offset from the remote
   store, and restore starting from that offset.
- Only truncate the changelog is the changelog topic is not a
      topology source, and is not compacted.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
@@ -1,0 +1,86 @@
+package dev.responsive.kafka.clients;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.OptionalLong;
+import java.util.function.Function;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+
+/**
+ * Consumer implementation to use for restoring Responsive stores. This Restore Consumer emulates
+ * consumping a topic that has been truncated up to some offset on a subset of topic partitions.
+ * The specific offsets are provided by a function mapping the topic partition to the start
+ * offset. This consumer is used in conjunction with Responsive Stores, which provide a safe
+ * starting offset that has been committed to a remote storage system. This is useful for
+ * minimizing the length of restores when using a Responsive Store over a compacted changelog
+ * topic (which cannot be truncated).
+ *
+ * @param <K> Type of the record keys
+ * @param <V> Type of the record values
+ */
+public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
+  final Function<TopicPartition, OptionalLong> startOffsets;
+
+  public ResponsiveRestoreConsumer(
+      final Consumer<K, V> delegate,
+      final Function<TopicPartition, OptionalLong> startOffsets
+  ) {
+    super(delegate);
+    this.startOffsets = Objects.requireNonNull(startOffsets);
+  }
+
+  private List<TopicPartition> seekInitial(final Collection<TopicPartition> partitions) {
+    final List<TopicPartition> notFound = new LinkedList<>();
+    for (final TopicPartition p : partitions) {
+      final OptionalLong start = startOffsets.apply(p);
+      if (start.isPresent()) {
+        delegate.seek(p, start.getAsLong());
+      } else {
+        notFound.add(p);
+      }
+    }
+    return Collections.unmodifiableList(notFound);
+  }
+
+  @Override
+  public void assign(Collection<TopicPartition> partitions) {
+    super.assign(partitions);
+    seekInitial(partitions);
+  }
+
+  @Override
+  public void seek(final TopicPartition partition, final long offset) {
+    super.seek(partition, Math.max(offset, startOffsets.apply(partition).orElse(offset)));
+  }
+
+  @Override
+  public void seek(final TopicPartition partition, final OffsetAndMetadata offsetAndMetadata) {
+    super.seek(
+        partition,
+        new OffsetAndMetadata(
+            Math.max(
+                offsetAndMetadata.offset(),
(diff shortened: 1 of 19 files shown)


kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
```diff
@@ -65,4 +81,73 @@ public static long durationToMillis(final Duration duration, final String parame
...
+        changelogTopic
+    );
+    LOG.info("Changelog topic {}: compacted({}), source({})", changelogTopic, compacted, source);
+    if (compacted.isEmpty()) {
+      LOG.warn("Could not find topic {}. This should not happen. Will not truncate.",
+          changelogTopic
+      );
+      return false;
+    }
+    if (compacted.get()) {
```
agavra, line 143: if it's compact & delete do we want to truncate?

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.OptionalLong;
+import java.util.function.Function;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+
+public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
```
agavra, line 14: nit: javadoc explaining the nuances of the interaction between this and the streams restore protocol would be helpful

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -63,14 +63,16 @@ class CommitBuffer<K> implements RecordBatchingStateRestoreCallback {
...
 
   CommitBuffer(
       final CassandraClient client,
       final String tableName,
       final TopicPartition changelog,
       final RecordCollector.Supplier recordCollector,
       final Admin admin,
-      final BufferPlugin<K> plugin
+      final BufferPlugin<K> plugin,
+      final boolean truncateChangelog
```
agavra, line 75: the UX of moving from compact -> delete & truncate would be to first change configs, then restart the node. I think that makes sense to me - I do wonder if we want a per-topology config to turn off truncation entirely, but that can be a follow-up.

kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
```diff
@@ -65,4 +81,73 @@ public static long durationToMillis(final Duration duration, final String parame
...
+    boolean compaction = cleanupPolicyEntry != null
+        && cleanupPolicyEntry.value().contains(TopicConfig.CLEANUP_POLICY_COMPACT);
+    return Optional.of(compaction);
+  }
+
+  private static boolean changelogIsSource(
+      final TopologyDescription description,
+      final String topicName
+  ) {
+    for (final var subtopology : description.subtopologies()) {
```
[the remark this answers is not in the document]
rodesai: I'm not sure I follow the concern - we still don't want to truncate in this case right?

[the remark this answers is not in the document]
agavra: ah yes... not sure what I was thinking!

kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
rodesai: I'm not sure. I conservatively assumed Kafka would fail the delete whenever there was compaction enabled.

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
rodesai: I'll let @ableegoldman chime in here. As per our discussion yesterday she was planning on experimenting with some options here.

kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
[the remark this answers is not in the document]
ableegoldman: >  there is a topology that treats the source as a stream and joins it against a table version of itself 

"Fun" fact: unless this changed recently, which I doubt has or will any time soon, it's not possible to read the same topic as multiple distinct source nodes anyway (within a single subtopology). You can of course reuse a source KStream or KTable with multiple downstream child nodes, and you can even re-read a topic (or `Pattern`) to create a new KStream or KTable but only if it has exactly the same parameters as the first KStream/KTable you created (eg offset reset policy, SerDe) and is the same type -- so you can't have a both a KStream and KTable source node for the same topic.

(To self-join the way you described, you would have to read it as a KStream and then use `#toTable` to get it again as a KTable for the stream-table join. But that's not a true source KTable anymore)

kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
ableegoldman: I suppose we should test this. I did try asking in the Confluent Community slack just now, so we'll see. I'm not sure how active the other channels are

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
ableegoldman: Yeah it's not worth worrying about this until we settle on the broader API, but fwiw I agree -- we should expose this with both an application/topology-wide default* and a store-level granular optimization. I'd imagine most users either want it on for everything or nothing. Then a handful who will possibly want it on as the topology default, but with carveouts for select changelogs that are eg read downstream or by another app

kafka-client/src/main/java/dev/responsive/utils/StoreUtil.java
rodesai: Based on the discussion (thanks for following up on it @ableegoldman !) from community slack (https://confluentcommunity.slack.com/archives/C499EFQS0/p1688760013869749) - it should be safe to truncate `compact, delete` topics. I'll update.

PR #60, opened 2023-07-07
flush state stores on every commit

rodesai:
This patch flushes state stores on every streams commit. The stores discover commits and the associated offsets via a notification registered with ResponsiveStoreRegistry. The notification is called by a ResponsiveProducer callback on each commit

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/MetricPublishingCommitListener.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/MetricPublishingCommitListener.java
@@ -72,7 +72,7 @@ public class MetricPublishingCommitListener implements ResponsiveConsumer.Listen
   ) {
     this.metrics = Objects.requireNonNull(metrics);
     this.threadId = Objects.requireNonNull(threadId);
-    offsetRecorder.addCommitCallback(this::onCommit);
+    offsetRecorder.addCommitCallback(this::commitCallback);
   }
 
   private MetricName metricName(final RecordingKey k) {
@@ -88,7 +88,7 @@ public class MetricPublishingCommitListener implements ResponsiveConsumer.Listen
     return new MetricName("committed-offset", "responsive.streams", "", tags);
   }
 
-  public void onCommit(
+  private void commitCallback(
       final Map<RecordingKey, Long> committedOffsets,
       final Map<TopicPartition, Long> unused
   ) {
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/MetricPublishingCommitListener.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/MetricPublishingCommitListener.java
@@ -16,7 +16,7 @@
 
 package dev.responsive.kafka.clients;
 
-import dev.responsive.kafka.clients.ResponsiveProducer.RecordingKey;
+import dev.responsive.kafka.clients.OffsetRecorder.RecordingKey;
 import java.io.Closeable;
 import java.util.Collection;
 import java.util.HashMap;
@@ -41,8 +41,7 @@ import org.slf4j.LoggerFactory;
  *
  * Synchronization: All shared access goes through the "offsets" map.
  */
-public class MetricPublishingCommitListener
-    implements ResponsiveProducer.Listener, ResponsiveConsumer.Listener, Closeable {
+public class MetricPublishingCommitListener implements ResponsiveConsumer.Listener, Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(MetricPublishingCommitListener.class);
   private final Metrics metrics;
   private final String threadId;
@@ -66,9 +65,14 @@ public class MetricPublishingCommitListener
     }
   }
 
-  public MetricPublishingCommitListener(final Metrics metrics, final String threadId) {
+  public MetricPublishingCommitListener(
+      final Metrics metrics,
+      final String threadId,
+      final OffsetRecorder offsetRecorder
+  ) {
     this.metrics = Objects.requireNonNull(metrics);
     this.threadId = Objects.requireNonNull(threadId);
+    offsetRecorder.addCommitCallback(this::onCommit);
   }
 
   private MetricName metricName(final RecordingKey k) {
@@ -84,8 +88,10 @@ public class MetricPublishingCommitListener
     return new MetricName("committed-offset", "responsive.streams", "", tags);
   }
 
-  @Override
-  public void onCommit(final Map<RecordingKey, Long> committedOffsets) {
+  public void onCommit(
+      final Map<RecordingKey, Long> committedOffsets,
+      final Map<TopicPartition, Long> unused
+  ) {
     for (final var e : committedOffsets.entrySet()) {
       offsets.computeIfPresent(
           e.getKey().getPartition(),
@@ -104,10 +110,6 @@ public class MetricPublishingCommitListener
           }
       );
     }
-  }
-
-  @Override
-  public void onClose() {
   }
 
   @Override
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
@@ -32,7 +32,15 @@ class OffsetRecorder {
     commitCallback.add(callback);
   }
 
-  public synchronized void recordConsumedOffsets(
+  public ResponsiveProducer.Listener getProducerListener() {
+    return producerListener;
+  }
+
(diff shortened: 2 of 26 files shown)


kafka-client/src/test/java/dev/responsive/kafka/store/ResponsiveStoreRestoreIntegrationTest.java
```diff
@@ -42,6 +42,7 @@
 import static org.apache.kafka.streams.StreamsConfig.producerPrefix;
```
rodesai, line 42: I need to rename this class to ResponsiveStoreIntegrationTest

kafka-client/src/main/java/dev/responsive/kafka/clients/MetricPublishingCommitListener.java
```diff
@@ -84,8 +88,10 @@ private MetricName metricName(final TopicPartition partition, final String consu
     return new MetricName("committed-offset", "responsive.streams", "", tags);
   }
 
-  @Override
-  public void onCommit(final Map<RecordingKey, Long> committedOffsets) {
+  public void onCommit(
```
agavra, line 91:
```suggestion
  @Override
  public void onCommit(
```

kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public synchronized void recordConsumedOffsets(
+      final String consumerGroup,
+      final TopicPartition partition,
+      final long offset
+  ) {
+    uncommitted.put(new RecordingKey(partition, consumerGroup), offset);
+  }
+
+  public synchronized void recordProduce(final RecordMetadata recordMetadata) {
```
agavra, line 43:
```suggestion
  public synchronized void onProduced(final RecordMetadata recordMetadata) {
```
(same for `recordX` below) at first I was confused because I thought this produces records!

kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
```diff
@@ -0,0 +1,163 @@
...
+    written.compute(
+        partition,
+        (k, v) -> v == null ? recordMetadata.offset() : Math.max(recordMetadata.offset(), v)
+    );
+  }
+
+  public synchronized void recordCommit() {
+    final Map<RecordingKey, Long> committedOffsets = Map.copyOf(uncommitted);
+    final Map<TopicPartition, Long> writtenOffsets = Map.copyOf(written);
+    commitCallback.forEach(c -> c.onCommit(committedOffsets, writtenOffsets));
```
agavra, line 55: these callbacks will all run within the synchronized block, I suspect that was intentional? If I'm following the code correctly that means we're flushing to the remote within the synchronized block, is there any reason that's necessary? We're making copies of things anyway, and if (again) I'm reading the code correctly that means each thread will block on one another if commits happen concurrently.

on that thought, why does this need to be synchronized at all? It looks like it's only created once per stream thread?

kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
```diff
@@ -0,0 +1,163 @@
...
+                                           String consumerGroupId) {
+      if (!eos) {
+        throw new IllegalStateException("producer commit is not expected with alos");
+      }
+      offsets.forEach((p, o) -> recordConsumedOffsets(consumerGroupId, p, o.offset()));
+    }
+
+    @Override
+    public void onCommit() {
+      // TODO: throw if not eos
```
agavra, line 104: seems easy enough to just do?

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -177,31 +174,13 @@ int size() {
...
 
-  public void flush() {
+  public void flush(long offset) {
     if (buffer.isEmpty()) {
       // no need to do anything if the buffer is empty
       LOG.info("Ignoring flush() to empty commit buffer for {}[{}]", tableName, partition);
       return;
     }
 
-    // TODO: what happens if flush is called before there's any committed offsets?
```
agavra, line 187: yay!

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStore.java
```diff
@@ -138,7 +137,8 @@ public void init(final StateStoreContext context, final StateStore root) {
       registration = new ResponsiveStoreRegistration(
           name.kafkaName(),
           topicPartition,
-          offset == -1 ? 0 : offset
+          offset == -1 ? 0 : offset,
+          buffer::flush
```
agavra, line 141: we probably don't want to actually flush every commit and instead replicate the behavior in `StateManagerUtil#checkpointNeeded` (and make it configurable). Based on my testing, if we flush every 100ms we're going to be in a bit of a pickle! especially when we explode partitions since we'll be sending an offset write to a partition even if there are no data writes to that partition

agavra (commented):
I already have a test as part of the virtual partition thing testing EOS flushing intervals, I can separate that into its own PR and merge it so we can use that test here as well.

2 main comments: re synchronization and actual flush interval

kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
rodesai: whoops - I wrote this comment before wiring in the eos flag. Fixing

kafka-client/src/main/java/dev/responsive/kafka/clients/OffsetRecorder.java
rodesai: yeah I suppose we don't need to do the actual flushes when synchronized (though it's not the end of the world as this should be the only thread trying to acquire this lock at that point). The reason its synchronized is because the producer send callbacks happen on the producer I/O threads.

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveProducer.java
```diff
@@ -144,46 +90,109 @@ public void sendOffsetsToTransaction(
...
   }
 
   @Override
   public void abortTransaction() throws ProducerFencedException {
     wrapped.abortTransaction();
-    offsetRecorder.onAbort();
+    listeners.forEach(Listener::onAbort);
+  }
+
+  private static class RecordingFuture implements Future<RecordMetadata> {
```
agavra, line 122: suggestion: can we move inner classes to the bottom of the file? it makes it a bit easier to read through the flow

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
```diff
@@ -57,6 +57,18 @@ public class ResponsiveDriverConfig extends AbstractConfig {
   public static final String CLIENT_SECRET_CONFIG = "responsive.client.secret";
   private static final String CLIENT_SECRET_DOC = "The client secret for authenticated access";
 
+  // ------------------ store tuning parameters -------------------------------
+
+  // TODO: we should have another config that's applied globally, that sets a size bound on
+  //       the total amount of buffered data. That config can be used to keep a bound on
+  //       memory usage and restore times. This config sets a store-level threshold for optimal
+  //       transaction sizes.
+  public static final String STORE_FLUSH_RECORDS_THRESHOLD
```
agavra, line 66: (for a follow-up) an additional time based one can be really useful for us when we're demo'ing/plugging in to customer staging environments that may not produce 10K records reliably - defaulting it to something like 1min is probably reasonable.

agavra (approved):
We're going to have lots of merge conflicts between this and #62  😆

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveDriverConfig.java
```diff
@@ -57,6 +57,18 @@ public class ResponsiveDriverConfig extends AbstractConfig {
...
+      = "responsive.store.flush.records.interval";
+  public static final String STORE_FLUSH_RECORDS_THRESHOLD_DOC = "The number of records to"
+      + " accumulate in each store before flushing to remote";
+  public static final int STORE_FLUSH_RECORDS_THRESHOLD_DEFAULT = 10000;
```
ableegoldman, line 70: I honestly think this is kind of large

ableegoldman (approved):
LGTM

PR #61, opened 2023-07-07
refactor CommitBuffer to use async calls

agavra:
move code around and use `CompletionStage` to set up for the virtual partition PR

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/AtomicWriter.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/AtomicWriter.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/AtomicWriter.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * {@code AtomicWriter} writes data atomically using
+ * LWTs to the remote store.
+ */
+class AtomicWriter<K> {
```
rodesai, line 37: it would be good to unit test this class

rodesai (approved):
code LGTM - would be good to test the writer.

PR #63, opened 2023-07-10
break up restore batches in to chunks of the maximum batch size

rodesai:
- Break up received restore batches into batches that use the Cassandra max batch size. Otherwise, the atomic writer will break them up, but will use the offset for the entire received batch as the fencing offset for all the batches - which will cause the subsequent batches to fail (as the protocol expects that once the fencing offset is set, all data up to that offset has been written).
- Make the maximum batch size parameterizable. My local test was timing out with the default of 1000. -

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
@@ -20,8 +20,8 @@ import dev.responsive.db.CassandraClient;
 import dev.responsive.db.CassandraClient.OffsetRow;
 import dev.responsive.model.Result;
 import dev.responsive.utils.Iterators;
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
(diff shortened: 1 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -270,6 +296,20 @@ private AtomicWriteResult flush(final long offset, final UUID txnid) {
 
   @Override
   public void restoreBatch(final Collection<ConsumerRecord<byte[], byte[]>> records) {
+    final List<ConsumerRecord<byte[], byte[]>> batch = new LinkedList<>();
```
agavra, line 299: we should definitely use an `ArrayList` here and just size it to the `maxBatchSize`

I've had some really surprisingly bad experiences with LinkedLists where something innocuous like this causes really heavy GC load

agavra (commented):
I'm somewhat concerned that your local tests were timing out with that small of a batch size! I turned it down from 10K, which was originally working for me... we might want to look into that.

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
ableegoldman: +1

ableegoldman (approved):
LGTM, modulo change in data structure

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
ableegoldman: Since the max batch size is a constant, we should just create the ArrayList of that size up front in the constructor and then we can reuse it here. Especially since we're repeatedly clearing/refilling it anyway

rodesai: Yep, good call on the ArrayList. Using a long-lived buffer gives me the heebie-jeebies a little bit. I've been bitten in the past by that kind of thing where something winds up sticking around in the buffer that doesn't get cleared between calls for whatever reason (though of course we could/should clear it as the first thing the method does, it's somewhat hard to guarantee that always happens as the class evolves). I don't think there is a significant cost to just allocating it on each restore batch.

PR #64, opened 2023-07-11
add local flush triggers

rodesai:
Adds some more local (per-state-store) flush triggers:
- size-based: triggers flushes when the accumulated state is over some size in bytes
- time-based: triggers flushes when some number of milliseconds has elapsed since the last flush

Renames the existing config so the config names are consistent

Changes the default to flush every 30 seconds

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
@@ -91,20 +91,17 @@ public class ResponsiveConfig extends AbstractConfig {
   //       the total amount of buffered data. That config can be used to keep a bound on
   //       memory usage and restore times.
 
-  public static final String STORE_FLUSH_RECORDS_TRIGGER
-      = "responsive.store.flush.trigger.local.records";
+  public static final String STORE_FLUSH_RECORDS_TRIGGER_CONFIG = "responsive.store.flush.trigger.local.records";
   public static final String STORE_FLUSH_RECORDS_TRIGGER_DOC = "The number of records to"
       + " accumulate in each store before flushing to remote";
   public static final int STORE_FLUSH_RECORDS_TRIGGER_DEFAULT = Integer.MAX_VALUE;
 
-  public static final String STORE_FLUSH_BYTES_TRIGGER
-      = "responsive.store.flush.trigger.local.bytes";
+  public static final String STORE_FLUSH_BYTES_TRIGGER_CONFIG = "responsive.store.flush.trigger.local.bytes";
   public static final String STORE_FLUSH_BYTES_TRIGGER_DOC = "The size in bytes of buffered"
       + "records to accumulate in each store before flushing to remote";
   public static final Long STORE_FLUSH_BYTES_TRIGGER_DEFAULT = Long.MAX_VALUE;
 
-  public static final String STORE_FLUSH_INTERVAL_TRIGGER
-      = "responsive.store.flush.trigger.local.interval.ms";
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER_MS_CONFIG = "responsive.store.flush.trigger.local.interval.ms";
   public static final String STORE_FLUSH_INTERVAL_TRIGGER_DOC = "The maximum time to wait "
       + "between consecutive flushes of a given store. Note that this is only evaluated at "
       + "commit time.";
@@ -182,7 +179,7 @@ public class ResponsiveConfig extends AbstractConfig {
           Importance.MEDIUM,
           REQUEST_TIMEOUT_MS_DOC
       ).define(
-          STORE_FLUSH_RECORDS_TRIGGER,
+          STORE_FLUSH_RECORDS_TRIGGER_CONFIG,
           Type.INT,
           STORE_FLUSH_RECORDS_TRIGGER_DEFAULT,
           Importance.MEDIUM,
@@ -200,13 +197,13 @@ public class ResponsiveConfig extends AbstractConfig {
           Importance.MEDIUM,
           MAX_CONCURRENT_REMOTE_WRITES_DOC
       ).define(
-          STORE_FLUSH_BYTES_TRIGGER,
+          STORE_FLUSH_BYTES_TRIGGER_CONFIG,
           Type.LONG,
           STORE_FLUSH_BYTES_TRIGGER_DEFAULT,
           Importance.MEDIUM,
           STORE_FLUSH_BYTES_TRIGGER_DOC
       ).define(
-          STORE_FLUSH_INTERVAL_TRIGGER,
+          STORE_FLUSH_INTERVAL_TRIGGER_MS_CONFIG,
           Type.LONG,
           STORE_FLUSH_INTERVAL_TRIGGER_DEFAULT,
           Importance.MEDIUM,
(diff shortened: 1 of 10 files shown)


kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
```diff
@@ -88,6 +77,30 @@ public class ResponsiveConfig extends AbstractConfig {
...
+      + " accumulate in each store before flushing to remote";
+  public static final int STORE_FLUSH_RECORDS_TRIGGER_DEFAULT = Integer.MAX_VALUE;
+
+  public static final String STORE_FLUSH_BYTES_TRIGGER
+      = "responsive.store.flush.trigger.local.bytes";
+  public static final String STORE_FLUSH_BYTES_TRIGGER_DOC = "The size in bytes of buffered"
+      + "records to accumulate in each store before flushing to remote";
+  public static final Long STORE_FLUSH_BYTES_TRIGGER_DEFAULT = Long.MAX_VALUE;
+
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER
```
agavra, line 96:
```suggestion
  public static final String STORE_FLUSH_INTERVAL_TRIGGER_MS
```

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -225,12 +235,61 @@ int size() {
...
+      );
+      recordsTrigger = true;
+    } else {
+      LOG.debug("records buffered {} not over trigger {}",
+          buffer.size(),
+          flushTriggers.getRecords()
+      );
+    }
+    // This is a little inefficient (as opposed to maintaining the size as we go along).
+    // On the other hand, it's much less error-prone.
```
agavra, line 261: let's not call this if the configuration is set to `Long.MAX_VALUE`

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -225,12 +235,61 @@ int size() {
...
+        .mapToLong(e -> plugin.bytes(e.key).get().length + (e.isTombstone ? 0 : e.value.length))
+        .sum();
+  }
+
+  private boolean shouldFlush() {
+    boolean recordsTrigger = false;
+    boolean bytesTrigger = false;
+    boolean timeTrigger = false;
+    if (buffer.size() >= flushTriggers.getRecords()) {
+      LOG.info("will flush due to records buffered {} over trigger {}",
```
agavra, line 249: let's include store name + partition in all of these logs (my log format is `tableName[kafkaPartition:subpartition]` where `subpartition` is optional)

either that or migrate to AK's prefixed log class

agavra (approved):
some comments inline

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
rodesai: I got a little nervous about this so I just refactored the buffer part into a size-tracking buffer

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -222,19 +230,66 @@ long offset() {
...
+          flushTriggers.getRecords()
+      );
+      recordsTrigger = true;
+    } else {
+      LOG.debug("records buffered {} not over trigger {}",
+          buffer.reader().size(),
+          flushTriggers.getRecords()
+      );
+    }
+    // This is a little inefficient (as opposed to maintaining the size as we go along).
```
agavra, line 256: comment no longer applicable

kafka-client/src/main/java/dev/responsive/kafka/store/SizeTrackingBuffer.java
```diff
@@ -0,0 +1,38 @@
...
+    buffer = new TreeMap<>(plugin);
+    reader = Collections.unmodifiableNavigableMap(buffer);
+  }
+
+  public long getBytes() {
+    return bytes;
+  }
+
+  public void put(final K key, final Result<K> value) {
+    bytes += plugin.bytes(key).get().length + (value.isTombstone ? 0 : value.value.length);
```
agavra, line 26: if this is a replace we should subtract the previous element, right?

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
```diff
@@ -88,6 +77,30 @@ public class ResponsiveConfig extends AbstractConfig {
       + "configuration does not apply to global stores.";
   private static final int STORAGE_DESIRED_NUM_PARTITIONS_DEFAULT = 4096;
 
+  // TODO: we should have another config that's applied globally, that sets a size bound on
+  //       the total amount of buffered data. That config can be used to keep a bound on
+  //       memory usage and restore times.
+
+  public static final String STORE_FLUSH_RECORDS_TRIGGER
```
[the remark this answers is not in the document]
ableegoldman: Also the other config variable names end in `_CONFIG`, can we maintain that here?

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
ableegoldman: 💯  the log prefix is great we should use that all the time where there's going to consistent relevant context (like name of store, partition, thread, etc)

kafka-client/src/main/java/dev/responsive/kafka/store/SizeTrackingBuffer.java
ableegoldman: Also if it's a tombstone we don't just add `0`, we should subtract the old size since this is a delete

ableegoldman: I guess that's basically what Almog was saying

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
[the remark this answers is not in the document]
rodesai: > FYI there's no line limit on this class so you can make this one line

As a person who doesn't use a monitor:

<grumble grumble> I suppose <grumble grumble>

kafka-client/src/main/java/dev/responsive/kafka/store/SizeTrackingBuffer.java
rodesai: 🤦 thanks guys. Fixed now.

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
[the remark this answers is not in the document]
agavra: :shipit: my current setup doesn't use a monitor :shipit:

kafka-client/src/main/java/dev/responsive/kafka/store/SizeTrackingBuffer.java
```diff
@@ -0,0 +1,45 @@
...
+  public long getBytes() {
+    return bytes;
+  }
+
+  public void put(final K key, final Result<K> value) {
+    if (buffer.containsKey(key)) {
+      bytes -= sizeOf(key, buffer.get(key));
+    }
+    bytes += sizeOf(key, value);
+    buffer.put(key, value);
```
agavra, line 30:
```suggestion
    bytes += sizeOf(key, value);
    final var old = buffer.put(key, value);
    bytes -= old == null ? 0 : sizeOf(old)
```

avoid an extra lookup ;)

PR #66, opened 2023-07-12
add E2E tests for subpartitioning

agavra:


kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
```diff
@@ -241,7 +253,8 @@ public SubPartitioner getSubPartitioner(
       final TableName name,
       final String changelogTopicName
   ) {
-    final OptionalInt actualRemoteCount = cassandraClient.numPartitions(name.cassandraName());
+    // TODO(agavra): write the actual remote partition count into cassandra
```
agavra, line 256: I had to undo this because each store initializes only its offsets, which means that the second store that tries to check this will get 1/Nth partitions when it queries cassandraClient.numPartitions for that table.

The fix is to persist it in the metadata row.

PR #70, opened 2023-07-13
switch the fencing protocol to use epochs instead of offsets/permits

agavra:


kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
```diff
@@ -348,37 +293,43 @@ public void prepareStatements(final String tableName) {
...
             .build()
     ));
 
-    finalizeTxn.computeIfAbsent(tableName, k -> session.prepare(
+    ensureEpoch.computeIfAbsent(tableName, k -> session.prepare(
         QueryBuilder
             .update(tableName)
-            .setColumn(PERMIT.column(), QueryBuilder.literal(UNSET_PERMIT))
-            .setColumn(OFFSET.column(), bindMarker(OFFSET.bind()))
+            .setColumn(EPOCH.column(), bindMarker(EPOCH.bind()))
```
rodesai, line 329: this makes me a little bit nervous, because the database could technically optimize this whole statement away and still be "correct". I wonder if it would make sense to set another column with some random uuid that always changes or something to ensure this always is applied.

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -444,48 +424,42 @@ static <I> AtomicWriteResult drain(
...
 
-  private static AtomicWriteResult maybeSetFailure(
-      final AtomicReference<AtomicWriteResult> failed,
+  private static AtomicWriteResult setIfNotApplied(
+      final AtomicReference<AtomicWriteResult> other,
       final AtomicWriteResult awr
   ) {
     if (!awr.wasApplied()) {
-      failed.set(awr);
+      other.set(awr);
```
rodesai, line 462: nit: maybe rename to `aggregateResult`

rodesai (commented):
LGTM overall. One concern inline

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -155,22 +155,40 @@ static <K> CommitBuffer<K> from(
...
+    subPartitioner.all(partition).forEach(p -> client.initializeMetadata(tableName, p));
+    lastFlush = clock.get();
+
+    // attempt to reserve an epoch - all epoch reservations will be done
+    // under the first sub-partition and then "broadcast" to the other
+    // partitions
+    final int basePartition = subPartitioner.first(partition);
+    this.epoch = client.metadata(tableName, basePartition).epoch + 1;
+
+    for (final int subPartition : subPartitioner.all(partition).toArray()) {
```
rodesai, line 174: do we have a test that ensures these are iterated in order?

agavra: will add 👍

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -405,17 +385,17 @@ private AtomicWriteResult flush(final long offset, final UUID txnid, final int b
       return writeResult;
     }
 
-    writeResult = drain(writers.values(), AtomicWriter::finalizeTxn);
+    // this offset is only used for recovery, so it can (and should) be done only
+    // when all the flushes above have completed
+    writeResult = drain(writers.values(), writer -> writer.setOffset(offset));
```
rodesai, line 390: We can just track the offset in the base partition now that it's just a recovery optimization. Might help reduce some write load. Not something we need to do in this PR though.

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
agavra: I'm not too concerned as the guarantee is "Use the IF keyword to introduce a condition that must return TRUE for the update to succeed." - `IF` and `WHERE` are not equivalent, I can imagine it optimizing it away with a `WHERE` clause.

OTOH I don't think there's much overhead for adding an additional column, but I'll research that tomorrow as a follow up.

rodesai: ah - I did not catch that distinction. Very cool

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
agavra: will do in a follow-up - I actually think this could be really helpful to reduce the write load (it'll cut it in half)

rodesai (approved):
:shipit:

PR #71, opened 2023-07-13
only write offset to base subpartition

agavra:
follow up from #70 

image

Doesn't seem to affect the max process rate (expected) but the max commit time ms definitely has a stat-sig drop, which I think will reduce load on C*:

image


rodesai (approved):
LGTM

PR #73, opened 2023-07-13
make all session commands idempotent

agavra:


apurvam (approved):
LGTM, assuming that it was just an oversight to not have them be idempotent from the outset.

PR #76, opened 2023-07-17
improve flushing by not flushing unnecessary partitions

agavra:


kafka-client/src/test/java/dev/responsive/kafka/store/CommitBufferTest.java
```diff
@@ -123,7 +123,7 @@ public void shouldFlushWithCorrectEpochForAllSubpartitionsAndOffsetForBaseSubpar
     // When:
     // key 0 -> subpartition 2 * 3 + 0 % 3 = 6
     // key 1 -> subpartition 2 * 3 + 1 % 3 =  7
-    // nothing in subpartition 8, but offset should still be flushed
+    // nothing in subpartition 8, should not be flushed
```
apurvam, line 126: Is there supposed to be a check that changes here?

apurvam (approved):
LGTM, assuming that there is a test case for this.

kafka-client/src/test/java/dev/responsive/kafka/store/CommitBufferTest.java
agavra: I realized that the test wasn't testing what it claimed to be testing earlier - it was actually testing the init behavior. I'm not sure of a good way to test that something "doesn't" happen here given it's just a performance optimization and not correctness

agavra: I guess I could mock the client and verify calls, but that requires a lot of mocking...

agavra: I'll just do that, would be good to have a mocked framework to test this class anyway

agavra: 😆 at the conversation with myself, but I figured out an easier way to test this. done

PR #78, opened 2023-07-17
support unencrypted channel b/w operator/controller

rodesai:
Adds a -disableTls option to operator. When set at the commandline, the operator connects to controller using a plain tcp connection.


rodesai:
I'm going to go ahead and merge this, as it's fairly innocuous I need it for testing.

PR #79, opened 2023-07-18
extend policy spec to support rate-based diagnosers

rodesai:
- Extend the policy CRD to support a configurable list of diagnosers, and add specs for rate-based diagnosers that can scale up or down based on processing rate per node.
- Clean up the validation logic in the CRDs. No validation happens in the consturctor, as exceptions there cause operator to crash. Instead the reconciler validates and marks the policy as failed if the speci s bad

### DIFF
--- a/controller-api/src/main/proto/responsive/controller/internal/v1/controller.proto
+++ b/controller-api/src/main/proto/responsive/controller/internal/v1/controller.proto
@@ -31,7 +31,9 @@ message PolicyState {
 }
 
 message Action {
-  ChangeApplicationState change_application_state = 1;
+  oneof action {
+    ChangeApplicationState change_application_state = 1;
+  }
 }
 
 message ChangeApplicationState {
--- a/operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
+++ b/operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
@@ -16,11 +16,19 @@
 
 package dev.responsive.k8s.controller;
 
+import dev.responsive.k8s.crd.DemoPolicy;
 import dev.responsive.k8s.crd.ResponsivePolicy;
 import dev.responsive.k8s.crd.ResponsivePolicySpec;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 import responsive.controller.v1.controller.proto.ControllerOuterClass;
 import responsive.controller.v1.controller.proto.ControllerOuterClass.ApplicationPolicy;
 import responsive.controller.v1.controller.proto.ControllerOuterClass.ApplicationState;
+import responsive.controller.v1.controller.proto.ControllerOuterClass.DemoPolicy.Diagnoser;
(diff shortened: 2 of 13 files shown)


controller-api/src/main/proto/responsive/controller/internal/v1/controller.proto
```diff
@@ -31,7 +31,9 @@ message PolicyState {
 }
 
 message Action {
-  ChangeApplicationState change_application_state = 1;
+  oneof action {
+    ChangeApplicationState change_application_state = 1;
```
agavra, line 35: do we need a particular upgrade path here? I'm not sure if we've enabled the autoscaler yet for the staging deployment, but if so we should make sure to turn it off and then roll this and the controller

generally we should consider how to make these kind of schema changes once we have customers in prod

operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
```diff
@@ -48,6 +56,55 @@ public static ControllerOuterClass.EmptyRequest emptyRequest(final ResponsivePol
         .build();
   }
 
+  private static ProcessingRateDiagnoser processingRateDiagnoserFromK8sResource(
```
agavra, line 59: these are all specific to `DemoPolicy` right? it might make sense to split this class into something that builds the proto just for the demo polcy and the general purpose stuff

operator/src/main/java/dev/responsive/k8s/crd/DemoPolicy.java
```diff
@@ -0,0 +1,133 @@
...
+      PROCESSING_RATE_SCALE_DOWN,
+      LAG_SCALE_UP
+    }
+
+    private final Type type;
+    private final Optional<ProcessingRateDiagnoser> processingRateScaleUp;
+    private final Optional<ProcessingRateDiagnoser> processingRateScaleDown;
+
+    @JsonCreator
+    public Diagnoser(
```
agavra, line 44: I think this should probably use `JsonSubTypes` and allowing us to multiplex instead of having each diagnoser be passed in as optional. https://github.com/confluentinc/ksql/blob/master/ksqldb-rest-model/src/main/java/io/confluent/ksql/rest/entity/KsqlEntity.java is a good example of this

agavra (approved):
change LGTM, I think we can clean up the API

controller-api/src/main/proto/responsive/controller/internal/v1/controller.proto
rodesai: This is actually a backwards-compatible change (moving a singular field into a `oneof`). But yeah in general I agree.

operator/src/main/java/dev/responsive/k8s/crd/DemoPolicy.java
rodesai: I tried but I couldn't get the yaml serialization/deserialization to work =/

rodesai: I think the problem is that the schema generator, which generates the CRD yaml schema doesn't know how to handle polymorphism. Kind of a bummer.

PR #81, opened 2023-07-20
support creating ResponsiveKafkaStreams via Properties instead of Map

agavra:
confirmed that you can pass in `Properties` where `Map<?, ?>` is expected


rodesai (approved):
:shipit:

PR #82, opened 2023-07-24
allow setting an environment prefix in operator

rodesai:
Adds a command line option --environment to set an environment
prefix in calls to the controller. This can be used to distinguish applications
in different k8s clusters (or other grouping of the user's choice). Also
adds support to the helm chart for specifying options.

### DIFF
--- a/operator/src/main/helm/values.yaml
+++ b/operator/src/main/helm/values.yaml
@@ -44,3 +44,5 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+operatorOpts: ""
--- a/operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
+++ b/operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
@@ -24,27 +24,32 @@ import responsive.controller.v1.controller.proto.ControllerOuterClass.Applicatio
 
 public final class ControllerProtoFactories {
   public static ControllerOuterClass.UpsertPolicyRequest upsertPolicyRequest(
+      final String environment,
       final ResponsivePolicy policy) {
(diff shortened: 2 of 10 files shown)


operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
```diff
@@ -63,7 +68,13 @@ private static ApplicationPolicy policyFromK8sResource(final ResponsivePolicySpe
     return builder.build();
   }
 
-  private static String getFullApplicationName(final ResponsivePolicy policy) {
-    return policy.getSpec().getApplicationNamespace() + "/" + policy.getSpec().getApplicationName();
+  private static String getFullApplicationName(
+      final String environment,
+      final ResponsivePolicy policy
+  ) {
+    final String prefix = environment.isEmpty() ? "" : environment + "/";
```
agavra, line 75: thoughts on making this required?

operator/src/main/helm/values.yaml
```diff
@@ -44,3 +44,5 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+operatorOpts: ""
```
agavra, line 48: can we get this value from k8s?

agavra (approved):
LGTM - thoughts inline

operator/src/main/helm/values.yaml
agavra: I guess not :( https://github.com/helm/helm/issues/2055

operator/src/main/java/dev/responsive/k8s/controller/ControllerProtoFactories.java
rodesai: I thought about it. I don't feel super strongly either way. I erred on the side of not breaking stuff that's already running.

PR #83, opened 2023-07-28
setup the build and workflows to use snapshot versioning.

rodesai:
With this patch, suppose the last released version of kafka-clients is 0.4.0. Then, all master builds will build kafka-clients with 0.5.0-SNAPSHOT. The next publish job will release 0.5.0, and then master will build 0.6.0-SNAPSHOT.

After this patch, I will manually run the following to get everything on the "right" snapshot version.:
```
./gradlew markNextVersion -Prelease.dryRun | grep Creating | commit --allow-empty -F -
 git push
./gradlew markNextVersion
```

### DIFF
--- a/.github/workflows/github-main.yaml
+++ b/.github/workflows/github-main.yaml
@@ -38,11 +38,6 @@ jobs:
         with:
           arguments: build
 
-      - name: Release
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: release
-
       - name: Publish
         uses: gradle/gradle-build-action@v2
         with:
--- a/.github/workflows/github-publish.yaml
+++ b/.github/workflows/github-publish.yaml
@@ -46,20 +46,6 @@ jobs:
         with:
           arguments: :${{ inputs.module }}:build
 
-      - name: Create Next Version Commit
-        run: |
-          git config --global user.name 'Responsive Tools'
-          git config --global user.email 'tools@responsive.dev'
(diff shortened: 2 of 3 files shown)


.github/workflows/github-main.yaml
```diff
@@ -38,11 +38,6 @@ jobs:
         with:
           arguments: build
 
-      - name: Release
```
rodesai, line 41: we no longer run the release plugin on main, because we want snapshot versions from the main build.

.github/workflows/github-publish.yaml
```diff
@@ -46,20 +46,6 @@ jobs:
...
-            -Prelease.incrementer=${{ inputs.increment }} \
-            -Prelease.dryRun | grep "Creating" | git commit --allow-empty -F -
-          git push
-
-      - name: Increment Version
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: :${{ inputs.module }}:markNextVersion -Prelease.incrementer=${{ inputs.increment }}
-
       - name: Release
```
rodesai, line 49: we now _first_ run the release plugin to tag HEAD with the current version (what we've been building SNAPSHOTs of). Then we publish that and then create the next version. Note that this should be idempotent if publish fails. The next iteration of the job will just skip the release step with ("nothing to release")

agavra: makes sense, previously we were working on 0.X.0 and releasing 0.X+1.0, this swaps that 👍

agavra (approved):

PR #88, opened 2023-08-02
Update publish GHA workflow to run smoke tests and promote images

rodesai:
Update publish GHA workflow to run smoke tests and promote images. I tested this
in my own fork of this repo to avoid messing with the release tags here.

### DIFF
--- a/.github/workflows/github-publish.yaml
+++ b/.github/workflows/github-publish.yaml
@@ -61,13 +61,76 @@ jobs:
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
           SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
 
-      - name: Publish Images
+      - name: Publish Operator Images
         uses: gradle/gradle-build-action@v2
         if: ${{ inputs.module == 'operator' }}
         with:
-          arguments: pushCRD pushDocker pushHelm -PdockerRegistry=${{ steps.login-ecr-public.outputs.registry }}/j8q9y0n6 -PhelmRegistry=${{ steps.login-ecr-public.outputs.registry }}/j8q9y0n6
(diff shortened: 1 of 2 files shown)


.github/workflows/github-publish.yaml
```diff
@@ -61,14 +61,77 @@ jobs:
...
+
+      - name: Set Operator Version
+        run: |
+          echo "OPERATOR_VERSION=$(./gradlew operator:cV | grep Project.version | sed 's/Project version: //')" >> $GITHUB_ENV
+
+      - name: Set Kafka Client Version
+        run: |
+          echo "KAFKA_CLIENTS_VERSION=$(./gradlew kafka-client:cV  | grep Project.version | sed 's/Project version: //')" >> $GITHUB_ENV
+
+      - name: Run Smoke Test (Operator)
```
agavra, line 111: if either of these fail, we probably want to "unpublish" the images, right? or is there a way for us not to publish them to the public location at all until after we've tested them?

agavra (approved):
lgtm, one comment inline

agavra:
hmm looks like `dev.responsive.kafka.store.ResponsiveGlobalStoreIntegrationTest` is flaky. i'll take a look at that

.github/workflows/github-publish.yaml
rodesai: yeah I was thinking about this and landed on either:
- we can just cut a new release
- we can setup a release patching process.
I think either approach is fine for where we're at now.

I swear I did consider running these before publishing but now I can't remember why I didn't implement it that way. Something didn't work when I tried. Let me think about it. I agree that would be preferable.

rodesai: I'll try to make it work. I'm just going to merge this first. And then do a follow-up with that change.

PR #90, opened 2023-08-03
separate out a fencing token and some minor refactoring

agavra:


kafka-client/src/main/java/dev/responsive/db/CassandraKeyValueSchema.java
```diff
@@ -95,10 +90,75 @@ public void create(final String tableName) {
...
+   * <p>Note that this method is idempotent as it uses Cassandra's
+   * {@code IF NOT EXISTS} functionality.
+   *
+   * @param table          the table that is initialized
+   * @param kafkaPartition the partition to initialize
+   */
   @Override
-  public void prepare(final String tableName) {
-    metadataStatements.prepare(tableName);
+  public FencingToken init(
```
agavra, line 105: I moved initialization code all into this method - it both creates the metadata rows if they don't exist and reserves the offset, this sets it up for a future Schema implementation that doesn't use epochs or fencing tokens (returns NoOpFencingToken) at all, but does use offsets.

kafka-client/src/main/java/dev/responsive/db/CassandraKeyValueSchema.java
```diff
@@ -302,22 +365,4 @@ protected static KeyValue<Bytes, byte[]> rows(final Row row) {
     );
   }
 
-  @Override
-  public int compare(final Bytes o1, final Bytes o2) {
-    return Objects.compare(o1, o2, Bytes::compareTo);
-  }
-
-  // Visible for Testing
-  public static class KeyValueMetadataKeys implements MetadataKeys {
```
agavra, line 311: this was just a confusing pattern, I removed this so each schema is self contained instead of having `MetadataStatements`

kafka-client/src/main/java/dev/responsive/kafka/store/AtomicWriter.java
```diff
@@ -108,9 +113,11 @@ public int partition() {
   }
 
   public AtomicWriteResult setOffset(final long offset) {
-    final var result = client.execute(
-        plugin.metadata().setOffset(tableName, partition, offset, epoch));
+    final BatchStatementBuilder builder = new BatchStatementBuilder(BatchType.UNLOGGED);
```
agavra, line 116: we now use a batch instead of a single statement so that a set offset statement isn't assumed to be a LWT

PR #91, opened 2023-08-03
support fact table schemas

agavra:
Introduces an idempotent KV schema that performs significantly better in the case where writes are known to be idempotent (such as in a deduplicator, time-series or fact-table style of data). 

To use this new store, users will use the following API:
```
ResponsiveStores.idempotentKeyValueStore(name);
```

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
+++ b/kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
@@ -1,0 +1,289 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.db;
+
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.bindMarker;
+import static dev.responsive.db.ColumnNames.DATA_KEY;
+import static dev.responsive.db.ColumnNames.DATA_VALUE;
+import static dev.responsive.db.ColumnNames.METADATA_KEY;
+import static dev.responsive.db.ColumnNames.OFFSET;
+
+import com.datastax.oss.driver.api.core.cql.BoundStatement;
+import com.datastax.oss.driver.api.core.cql.PreparedStatement;
+import com.datastax.oss.driver.api.core.cql.Row;
+import com.datastax.oss.driver.api.core.type.DataTypes;
+import com.datastax.oss.driver.api.querybuilder.QueryBuilder;
+import com.datastax.oss.driver.api.querybuilder.SchemaBuilder;
+import dev.responsive.db.partitioning.SubPartitioner;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import javax.annotation.CheckReturnValue;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CassandraIdempotentKeyValueSchema implements RemoteKeyValueSchema {
+
+  private static final Logger LOG = LoggerFactory.getLogger(
+      CassandraIdempotentKeyValueSchema.class);
+
+  private final CassandraClient client;
+
+  // use ConcurrentHashMap instead of ConcurrentMap in the declaration here
+  // because ConcurrentHashMap guarantees that the supplied function for
+  // computeIfAbsent is invoked exactly once per invocation of the method
+  // if the key is absent, else not at all. this guarantee is not present
+  // in all implementations of ConcurrentMap
+  private final ConcurrentHashMap<String, PreparedStatement> get;
+  private final ConcurrentHashMap<String, PreparedStatement> insert;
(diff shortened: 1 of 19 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/IdempotentKeyValueBytesStoreSupplier.java
```diff
@@ -0,0 +1,55 @@
...
+    if (isTimestamped) {
+      return new ResponsiveTimestampedStore(name, SchemaType.IDEMPOTENT);
+    } else {
+      return new ResponsiveStore(name, SchemaType.IDEMPOTENT);
+    }
+  }
+
+  @Override
+  public String metricsScope() {
+    return "responsive";
```
ableegoldman, line 53:
```suggestion
    return "responsive-idempotent";
```

or whatever name we land on

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -48,6 +48,34 @@ public static KeyValueBytesStoreSupplier timestampedKeyValueStore(final String n
     return new ResponsiveKeyValueBytesStoreSupplier(name, true);
   }
 
+  /**
+   * See for example {@link Stores#persistentKeyValueStore(String)}. This differs
+   * from {@link #keyValueStore(String)} in that it requires idempotent writes for
```
ableegoldman, line 53: Even if we don't go with another name, we should at the very least clarify much more what exactly we mean by "idempotent" here

kafka-client/src/main/java/dev/responsive/kafka/store/SchemaType.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.store;
+
+public enum SchemaType {
+  KEY_VALUE,
+  IDEMPOTENT
```
ableegoldman, line 21: What about something like, `KEY_VALUE` vs `VALUE_AGNOSTIC` ...or even `VALUE_LESS`?

This is a little awkward but maybe you'll have better ideas along this line of thought

agavra: I think @rodesai suggested `KeySet`, I think that might be a good thing to do here and just literally implement it as a key set. It still isn't 100% accurate because deletes can cause a zombie to corrupt data... 

`ResponsiveStores.keySetStore()` seems fine and `SchemaType.KEY_SET` is nice.

kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
```diff
@@ -0,0 +1,289 @@
...
+   */
+  @Override
+  public FencingToken init(
+      final String table,
+      final SubPartitioner partitioner,
+      final int kafkaPartition
+  ) {
+    client.execute(
+        QueryBuilder.insertInto(table)
+            .value(DATA_KEY.column(), QueryBuilder.literal(metadataKey(kafkaPartition)))
```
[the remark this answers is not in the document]
agavra: that's on my TODO list, I think we need that anyway and it wouldn't be that hard. I'll do that in a follow-up

kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
```diff
@@ -0,0 +1,289 @@
...
+  @Override
+  public KeyValueIterator<Bytes, byte[]> all(
+      final String tableName,
+      final int partition
+  ) {
+    throw new UnsupportedOperationException("all is not supported on Idempotent schemas");
+  }
+
+  private static ByteBuffer metadataKey(final int partition) {
+    final ByteBuffer key = ByteBuffer.wrap(new byte[METADATA_KEY.get().length + Integer.BYTES]);
```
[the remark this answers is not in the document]
agavra: I think it needs to be the same type as the key, which is blob - I found this relatively easy to debug with a small tool to generate the hex and then select from C* using `SELECT * FROM table WHERE dataKey = 0x<theHex>`.

I could play around with the format so that we could use `TextAsBlob` in C* for selecting, but I don't think we can make it a string.

kafka-client/src/main/java/dev/responsive/kafka/store/SchemaType.java
agavra: I went with `WRITE_ONCE`

agavra: OK I went with `FACT`. This use case is almost exactly summed up by the warehouse concept of a Fact store - it's not 100% in line with the definition but it's the best thing I could come up with, essentially it's expected to store "facts" where each key is expected to map to only a single value. Keys in fact tables are usually composite primary keys that reference something else (e.g. sensor, date -> measurement)

kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
```diff
@@ -0,0 +1,289 @@
...
+   * {@link ColumnNames#OFFSET} and no {@link ColumnNames#EPOCH}.
+   *
+   * <p>Note that this method is idempotent as it uses Cassandra's
+   * {@code IF NOT EXISTS} functionality.
+   *
+   * @param table          the table that is initialized
+   * @param kafkaPartition the partition to initialize
+   */
+  @Override
+  public FencingToken init(
```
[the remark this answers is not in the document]
agavra: The problem makes sense to me, I will change this in a follow-up PR because I think it would be every better to have the fencing token (I'll rename it) be a factory for the atomic writer. That way we don't have to batch at all if we don't want to for the fact table schema.

rodesai (approved):
LGTM!

kafka-client/src/main/java/dev/responsive/db/CassandraIdempotentKeyValueSchema.java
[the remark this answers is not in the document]
agavra: https://github.com/responsivedev/responsive-pub/pull/94

PR #93, opened 2023-08-05
builds docker images for amd64 and aarch64/arm64

rodesai:
- Add a pushDockerMultiArch task that builds and pushes a multi-architecture docker image.
- Update the github workflows to support using the new task. Each workflow that pushes images now needs extra steps to fetch the qemu emulator and install the docker build extensions.

### DIFF
--- a/operator/docker/Dockerfile
+++ b/operator/docker/Dockerfile
@@ -1,11 +1,15 @@
-FROM amd64/debian:11.6 
+FROM debian:11.6
+
+ARG TARGETPLATFORM
 
 RUN apt update
 RUN apt install -y gettext-base
 RUN apt install -y wget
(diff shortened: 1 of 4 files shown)


operator/docker/Dockerfile
```diff
@@ -1,11 +1,15 @@
-FROM amd64/debian:11.6 
+FROM debian:11.6
```
agavra, line 1: out of interest, how does it know which one to pull as the base image?

agavra (approved):

operator/docker/Dockerfile
rodesai: It's actually smart enough to figure it out based on the architecture that the image needs to run on. The multi-platform "images" are really an index file that point to a set of images - one for each supported platform. When you do `docker pull`, the docker client looks at the index and pulls the right platform-specific image.

PR #94, opened 2023-08-07
support data keys that are equal to the metadata key literal

agavra:
depends on #91 - adds a row type byte to each data key so that we can differentiate between a metadata key and a data key with the same byte contents

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/RowType.java
+++ b/kafka-client/src/main/java/dev/responsive/db/RowType.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/db/RowType.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package dev.responsive.db;
+
+import com.datastax.oss.driver.api.querybuilder.QueryBuilder;
+import com.datastax.oss.driver.api.querybuilder.term.Term;
+
+public enum RowType {
+
+  METADATA_ROW((byte) 0x0),
```
rodesai, line 24: it's probably better to use some non-zero value here to help distinguish an (erroneously) uninitialized value from one that was explicitly set.

rodesai (approved):
Looks good, modulo one in-line comment. We should also version our schemas before starting a production use case.

PR #97, opened 2023-08-09
fix: delay initialization of state stores and disable restore consumer for standby tasks

ableegoldman:
This patch is for mounting a defense against the possibility of standby tasks which may be assigned in a rebalance in order to turn them into active tasks in an immediate followup rebalance, once revoked by their current owner. In general, these standbys should always be temporary and either recycled to active or removed entirely by the next rebalance (though we don't assume this anywhere in case of change in behavior). We can, however, assume that active tasks are never going to be recycled into a standby.

The problem for us is that a standby task will end up fencing the active (and throwing us into a rebalance loop) if it goes through with initialization of the commit buffer during the StateStore's #init. So ultimately, we want to 

1. Skip the `CommitBuffer#init` call if the task is a standby
2. Enforce that the task is active for all put/get calls
3. Pause any changelog partitions belonging to standby tasks assigned to the restoreConsumer
4. After a standby is recycled but before it begins active restoration:
    i. complete the state store initialization
    ii. fetch the stored offset in the remote table
    iii. unpause the changelog

1-2 are relatively easy but unfortunately because a recycle is handled transparently with no calls to the underlying store, 3-4 are a bit tricky. 

Luckily we can leverage the existing seek/offset initialization logic in the restoreConsumer to hook into the state store when restoration starts, and solve 4.i - ii by asking the corresponding ResponsiveStoreRegistration for the stored offset to begin restoring from. The store registration calls the underlying store, which checks whether it's transitioned to ACTIVE and initializes the buffer if so.

For 3 & 4.iv we use a StateRestoreListener to detect when a task begins active restoration, since this listener is not invoked for standbys. Effectively we pause all changelogs when they are first registered, and then mark them as active to unpause them once we see a StateRestoreListener#onRestoreStart

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
@@ -1,5 +1,6 @@
 package dev.responsive.kafka.clients;
 
+import dev.responsive.kafka.store.ResponsiveRestoreListener;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.Collections;
@@ -48,17 +49,24 @@ public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
 
   private final Function<TopicPartition, OptionalLong> startOffsets;
   private final Set<TopicPartition> uninitializedOffsets = new HashSet<>();
+  private final ResponsiveRestoreListener restoreListener;
+  private final Set<TopicPartition> pausedChangelogPartitions = new HashSet<>();
 
   public ResponsiveRestoreConsumer(
       final String clientId,
       final Consumer<K, V> delegate,
-      final Function<TopicPartition, OptionalLong> startOffsets
+      final Function<TopicPartition, OptionalLong> startOffsets,
+      final ResponsiveRestoreListener restoreListener
   ) {
     super(delegate);
+
     this.startOffsets = Objects.requireNonNull(startOffsets);
+    this.restoreListener = Objects.requireNonNull(restoreListener);
     this.log = new LogContext(
         String.format("responsive-restore-consumer [%s]", Objects.requireNonNull(clientId))
     ).logger(ResponsiveConsumer.class);
+
+    restoreListener.addRestoreConsumer(this);
   }
 
   private Set<TopicPartition> initializeOffsets(final Collection<TopicPartition> partitions) {
@@ -86,12 +94,15 @@ public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
     // clear any now-unassigned changelogs
     uninitializedOffsets.retainAll(partitions);
 
+    pausedChangelogPartitions.retainAll(partitions);
+
     super.assign(partitions);
   }
 
   @Override
   public void unsubscribe() {
     uninitializedOffsets.clear();
+    pausedChangelogPartitions.clear();
     super.unsubscribe();
   }
 
@@ -137,4 +148,56 @@ public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
       uninitializedOffsets.removeAll(withUncachedPosition);
     }
   }
+
+  @Override
+  public void pause(final Collection<TopicPartition> partitions) {
+    final Set<TopicPartition> toPause = new HashSet<>(pausedChangelogPartitions);
+    toPause.addAll(partitions);
+    super.pause(toPause);
+  }
+
+  @Override
+  public void resume(final Collection<TopicPartition> partitions) {
+    final Set<TopicPartition> toResume = new HashSet<>(partitions);
+    toResume.removeAll(pausedChangelogPartitions);
+    super.resume(toResume);
+  }
+
+  @Override
+  public void close() {
+    restoreListener.removeRestoreConsumer(this);
+    super.close();
+  }
+
+  @Override
+  public void close(final Duration timeout) {
+    restoreListener.removeRestoreConsumer(this);
+    super.close(timeout);
+  }
+
+  /**
+   * Mark this changelog partition as a (potential) standby to block restoration.
+   * <p>
+   * This method is invoked on all changelogs during changelog registration, which begins with a
+   * {@link #seek}. If the changelog is actually from an active task, it will be removed from this
+   * set and resumed immediately afterward when the changelog registration completes.
+   * See {@link #markChangelogAsActive(TopicPartition)} for more details
+   */
+  public void markChangelogAsStandby(final TopicPartition partition) {
+    pausedChangelogPartitions.add(partition);
+  }
+
+  /**
+   * Mark this changelog partition as active to allow restoration to proceed.
+   * <p>
+   * This method is invoked on active task changelogs only, at the end of the changelog registration
+   * process. It is called from {@link ResponsiveRestoreListener#onRestoreStart} which is itself
+   * invoked only for active tasks, allowing us to differentiate between active and standby
+   * restoration.
+   * See {@link #markChangelogAsStandby(TopicPartition)} for more details
+   */
+  public void markChangelogAsActive(final TopicPartition partition) {
+    pausedChangelogPartitions.remove(partition);
+  }
+
 }
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/ActiveFilteringRestoreCallback.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/ActiveFilteringRestoreCallback.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.responsive.kafka.store;
+
+import java.util.Collection;
+import java.util.function.BooleanSupplier;
+import java.util.function.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.streams.processor.internals.RecordBatchingStateRestoreCallback;
+
+public class ActiveFilteringRestoreCallback implements RecordBatchingStateRestoreCallback {
+
+  private final TopicPartition changelogPartition;
+
+  private final Consumer<Collection<ConsumerRecord<byte[], byte[]>>> batchRestorer;
+  private final BooleanSupplier isActive;
+
+  public ActiveFilteringRestoreCallback(
+      final TopicPartition changelogPartition,
+      final Consumer<Collection<ConsumerRecord<byte[], byte[]>>> batchRestorer,
+      final BooleanSupplier isActive
+  ) {
+    this.changelogPartition = changelogPartition;
+    this.batchRestorer = batchRestorer;
+    this.isActive = isActive;
+  }
+
+  @Override
+  public void restoreBatch(final Collection<ConsumerRecord<byte[], byte[]>> records) {
+    if (isActive.getAsBoolean()) {
+      batchRestorer.accept(records);
+    } else {
+      throw new IllegalStateException("Unexpected attempt to restore batch to standby changelog: "
+                                          + changelogPartition);
+    }
+  }
+}
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
@@ -36,22 +36,23 @@ import java.util.List;
 import java.util.concurrent.TimeoutException;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.errors.ProcessorStateException;
 import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.processor.StateStore;
 import org.apache.kafka.streams.processor.StateStoreContext;
 import org.apache.kafka.streams.processor.internals.InternalProcessorContext;
+import org.apache.kafka.streams.processor.internals.Task.TaskType;
 import org.apache.kafka.streams.query.Position;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.internals.StoreQueryUtils;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ResponsivePartitionedStore implements KeyValueStore<Bytes, byte[]> {
 
-  private static final Logger LOG = LoggerFactory.getLogger(ResponsivePartitionedStore.class);
+  private Logger log;
 
   private final TableName name;
   private final Position position;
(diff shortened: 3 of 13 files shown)


kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -75,12 +106,64 @@ public void seek(final TopicPartition partition, final OffsetAndMetadata offsetA
...
       // is not empty. If it's empty, KafkaConsumer will seek to the actual beginning of _all_
       // partitions.
       super.seekToBeginning(withUncachedPosition);
     }
   }
+
+  @Override
+  public void pause(final Collection<TopicPartition> partitions) {
+    final Set<TopicPartition> toPause = new HashSet<>(pausedChangelogPartitions);
+    toPause.addAll(partitions);
```
agavra, line 121: will `pause` resume partitions that were previously paused but are not included in `partitions` parameter or is this just extra safety?

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -75,12 +106,64 @@ public void seek(final TopicPartition partition, final OffsetAndMetadata offsetA
...
+
+  /**
+   * Mark this changelog partition as a (potential) standby to block restoration.
+   * <p>
+   * This method is invoked on all changelogs during changelog registration, which begins with a
+   * {@link #seek}. If the changelog is actually from an active task, it will be removed from this
+   * set and resumed immediately afterward when the changelog registration completes.
+   * See {@link #markChangelogAsActive(TopicPartition)} for more details
+   */
+  public void markChangelogAsStandby(final TopicPartition partition) {
```
agavra, line 152: I can't find any usage of this method in this PR? also shouldn't this call `pause` ? (and shouldn't the below call `resume`?)

kafka-client/src/main/java/dev/responsive/kafka/store/ActiveFilteringRestoreCallback.java
```diff
@@ -0,0 +1,45 @@
...
+    this.batchRestorer = batchRestorer;
+    this.isActive = isActive;
+  }
+
+  @Override
+  public void restoreBatch(final Collection<ConsumerRecord<byte[], byte[]>> records) {
+    if (isActive.get()) {
+      batchRestorer.accept(records);
+    } else {
+      throw new IllegalStateException("Should not attempt to restore records to standby stores");
```
agavra, line 42: can we add into this exception the topic/partition that was attempted to be restored?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveGlobalStore.java
```diff
@@ -54,8 +54,6 @@ public class ResponsiveGlobalStore implements KeyValueStore<Bytes, byte[]> {
   private StateStoreContext context;
   private RemoteKeyValueSchema schema;
 
-  // TODO: instead of splitting this out into a separate store implementation, we should just
```
agavra, line 57: 😅 🧹

kafka-client/src/main/java/dev/responsive/kafka/store/ActiveFilteringRestoreCallback.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Collection;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.streams.processor.internals.RecordBatchingStateRestoreCallback;
+
+public class ActiveFilteringRestoreCallback implements RecordBatchingStateRestoreCallback {
+
+  private final Consumer<Collection<ConsumerRecord<byte[], byte[]>>> batchRestorer;
+  private final Supplier<Boolean> isActive;
```
agavra, line 27:
```suggestion
  private final BooleanSupplier isActive;
```
Looks like you already use that elsewhere ;)

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
```diff
@@ -130,22 +142,100 @@ public void init(final StateStoreContext context, final StateStore root) {
...
+
+  private boolean isActive() {
+    return context.taskType() == TaskType.ACTIVE;
+  }
+
+  /**
+   * The current stored offset in the remote table. Guaranteed to be less than or equal to the
+   * committed offset of the corresponding changelog topic partition in Kafka
+   * <p>
+   * NOTE: this always makes a remote call, should be cached eg when used to obtain starting offset
```
agavra, line 183: I think it's better not to cache this as it can be updated by other writers - it's a pretty lightweight call so unless it's called in a tight loop we should be good

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,111 @@
...
+          startingOffset,
+          endingOffset);
+    } else {
+      LOG.info("Beginning restoration from offset {} to {} for partition {} of state store {}",
+               startingOffset, endingOffset, topicPartition, storeName);
+    }
+  }
+
+  @Override
+  public void onBatchRestored(
```
agavra, line 76: do these need to be synchronized as well?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,111 @@
...
+ * this Streams application.
+ */
+public class ResponsiveRestoreListener implements StateRestoreListener {
+  private static final Logger LOG = LoggerFactory.getLogger(ResponsiveRestoreListener.class);
+
+  private final Set<ResponsiveRestoreConsumer<?, ?>> restoreConsumers = new LinkedHashSet<>();
+  private StateRestoreListener userListener;
+
+  public void registerUserRestoreListener(final StateRestoreListener restoreListener) {
+    userListener = restoreListener;
```
agavra, line 37: should we throw an exception if `userListener` is already set?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+    restoreConsumers.forEach(rc -> rc.markChangelogAsActive(topicPartition));
+
+    if (userListener != null) {
+      userListener.onRestoreStart(
+          topicPartition,
+          storeName,
+          startingOffset,
+          endingOffset);
+    } else {
```
agavra, line 69: for all of these, why are we only logging in the else clause?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,111 @@
...
+  // TODO: once we split up the global ResponsiveStoreRegistry by StreamThread, we can use that
+  //  to direct the call to the correct restore consumer
+  @Override
+  public synchronized void onRestoreStart(
+      final TopicPartition topicPartition,
+      final String storeName,
+      final long startingOffset,
+      final long endingOffset
+  ) {
+    restoreConsumers.forEach(rc -> rc.markChangelogAsActive(topicPartition));
```
agavra, line 61: I see the comment in your PR description that `this listener is not invoked for standbys` - can we add that comment inline here?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStoreRegistration.java
```diff
@@ -1,42 +1,142 @@
...
   private final String name;
+  private final TopicPartition changelogTopicPartition;
   private final Consumer<Long> onCommit;
 
+  private final LongSupplier storedOffset; // RPC to fetch offset committed in remote table
+  private final BooleanSupplier maybeInitialize; // returns true if/when initialization is done
+  private long startOffset; // stored offset as of initialization, ie where restore should start
+
+
+  @VisibleForTesting
```
agavra, line 41: it looks like this was already here, but we should not use this annotation since it's pulled from a shaded guava - instead let's just add a comment

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
```diff
@@ -130,22 +142,100 @@ public void init(final StateStoreContext context, final StateStore root) {
...
+            buffer::flush,
+            this::storedOffset
+        );
+      } else {
+        registration = new ResponsiveStoreRegistration(
+            name.kafkaName(),
+            topicPartition,
+            buffer::flush,
+            this::storedOffset,
+            this::maybeTransitionToActive
```
agavra, line 159: [major] am I reading this correctly in that if we started off with `isActive() == true` that we'll never initialize the buffer? (it seems like it only happens in the `maybeTransitionToActive` callback

agavra (commented):
comments inline, seems like a cleaner approach to me over the previous init inline with put/get

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStoreRegistration.java
```diff
@@ -1,42 +1,142 @@
...
-  private final long committedOffset;
+  public static final long NO_COMMITTED_OFFSET = -1L; // buffer is initialized but no prior offset
+  public static final long UNINITIALIZED_OFFSET = -2L; // buffer is not yet initialized
+
+  private final Logger log;
   private final String name;
+  private final TopicPartition changelogTopicPartition;
   private final Consumer<Long> onCommit;
 
+  private final LongSupplier storedOffset; // RPC to fetch offset committed in remote table
```
rodesai, line 36: nit: it might be better to have the stored offset supplier (or just the stored offset) returned by the initialize callback iff the store is active. That way you're guaranteed at compile time that we never query the offset on a standby store.

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -1,40 +1,65 @@
...
+ * 2. restore active/standby state stores:
+ *     i. poll for new record batch
+ *     ii. restore batch through StateRestoreCallback
+ *     iii. for each ACTIVE changelog in batch: call StateRestoreListener#onBatchRestored
+ *     iv. update current offset based on highest offset in batch
  *
  * @param <K> Type of the record keys
  * @param <V> Type of the record values
  */
 public class ResponsiveRestoreConsumer<K, V> extends DelegatingConsumer<K, V> {
```
rodesai, line 45: I'm not sure how confident you are about the assumptions around what state the task is in (active vs standby) at different points. But we could wire that information in via the store and assert on it here.

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
```diff
@@ -130,22 +142,100 @@ public void init(final StateStoreContext context, final StateStore root) {
...
+   * to the new task without going through #close or #init again. We use the restore consumer and a
+   * restore listener to infer when a task transitions from standby to active restoration. We also
+   * block all restoration while in standby mode, and wait to set the starting offset until it's
+   * recycled to ensure we don't overcount or undercount any changelog records
+   *
+   * @return whether the store finished initialization and is ready to restore as an active task
+   */
+  // TODO extract this and below methods into common superclass to share with Window/Session stores
+  private boolean maybeTransitionToActive() {
+    if (isActive()) {
```
rodesai, line 219: consider making this idempotent or throwing if it's called twice

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -49,8 +74,14 @@ private List<TopicPartition> seekInitial(final Collection<TopicPartition> partit
 
   @Override
   public void assign(Collection<TopicPartition> partitions) {
+    pausedChangelogPartitions.retainAll(partitions);
     super.assign(partitions);
```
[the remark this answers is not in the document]
ableegoldman: I'm like 99.9999% sure there will always be a `seek` on new partitions added to the restore consumer's assignment in Streams -- see my response about this on Slack. But, we should still minimize any reliance on assumptions like that. I'm splitting this part out into a separate PR so I'll address this feedback there

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
ableegoldman: Well right now it's only ever used to initialize the offset at the start of restoration -- in the current code it's essentially cached as the "committed offset" in ResponsiveStoreRegistration. But yeah, if you actually need the updated current stored offset, you should definitely not use a cached value 🙂

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Note: this is a global object which is shared across all StreamThreads and state stores in
+ * this Streams application.
+ */
+public class ResponsiveRestoreListener implements StateRestoreListener {
+  private static final Logger LOG = LoggerFactory.getLogger(ResponsiveRestoreListener.class);
+
+  private final Set<ResponsiveRestoreConsumer<?, ?>> restoreConsumers = new LinkedHashSet<>();
+  private StateRestoreListener userListener;
```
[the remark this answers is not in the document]
ableegoldman: I actually did use Optional at first, but I got annoyed by the IDE warning against Optional fields after a while and took it out. Don't personally have a strong preference either way

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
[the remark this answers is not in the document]
ableegoldman: I added some javadocs to our listener to hopefully clear up the semantics a bit

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStoreRegistration.java
ableegoldman: Bruno was always the main proponent of this philosophy, but over time I've come to heartedly agree with him: that we should never add "//visible for testing" comments because they can't ever be trusted to be kept up to date.

My personal philosophy is, it's not worth making the distinction between public vs package-private for testing purposes, especially if it's not meant to be user-facing.

If you bumped something up from private to package-private  or public for testing purposes, then it's important to leave a comment -- but ideally you don't ever do that in the first place so it should be a non-issue

Will definitely remove the annotation though

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -1,40 +1,65 @@
...
+ * This consumer is used by Kafka Streams for both active and standby task restoration, though
+ * only one or the other at a given time. This may be subject to change, but the general process
+ * follows this pattern:
+ * 1. initialize changelogs for newly assigned tasks:
+ *     i. restoreConsumer.assign(current assignment + new changelogs)
+ *     ii. if STANDBY_MODE: restoreConsumer.pause(standby changelogs)
+ *     iii. restoreConsumer.seekToBeginning(new changelogs)
+ *     iv. startOffset = restoreConsumer.position(changelog)
+ *     iv. for each ACTIVE changelog: call StateRestoreListener#onRestoreStart
+ * 2. restore active/standby state stores:
```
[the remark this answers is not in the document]
ableegoldman: well no, all the below steps will still be performed, always, because they are part of the Kafka Streams code -- we can't stop it from assigning changelogs from standby tasks and trying to poll for records to restore, but we can make sure it never actually restores to the remote or even fetches data from the changelogs by calling `pause` on the restoreConsumer for all standby partitions.

A paused partition is still part of the consumer's assignment, but it essentially freezes the consumer's background fetcher thread and prevents `poll` from returning any records for those partitions that may have already been fetched

ableegoldman:
Closing this out since we upgraded to a version with the Streams-side fix and should not receive standby tasks.

PR #98, opened 2023-08-09
use dedicated writers for different schemas

agavra:
this does a few things:
1. removes the `FencingToken` and introduces a `WriterFactory` so that all decisions around fencing are contained within an implementation of `RemoteWriter` (easier to review and verify that it works)
2. Renames `AtomicWriter` to `LwtWriter` and introduces `FactWriter` which does no fencing at all
3. Changes `FactWriter` to also avoid batching


kafka-client/src/main/java/dev/responsive/kafka/store/RemoteWriter.java
```diff
@@ -14,17 +14,19 @@
  * limitations under the License.
  */
 
-package dev.responsive.db;
```
rodesai, line 17: to me this actually feels like it fits better in the db package alongside the factory

kafka-client/src/main/java/dev/responsive/kafka/store/FactSchemaWriter.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Override
+  public void delete(final K key) {
+    statements.add(schema.delete(name, partition, key));
+  }
+
+  @Override
+  public CompletionStage<RemoteWriteResult> flush() {
+    final List<CompletionStage<RemoteWriteResult>> results = statements.stream()
+        .map(this::executeAsync)
```
rodesai, line 64: This doesn't guarantee any ordering right? I think this is safe for two reasons:
1. in the current store implementation, we'll never get two writes for the same key
2. even if that was not the case, the fact schema assumes that there are no updates
Might want to leave a comment clarifying that.

rodesai (approved):
LGTM

kafka-client/src/main/java/dev/responsive/kafka/store/FactSchemaWriter.java
agavra: correct, but also the commit buffer will never contain two values for the same key so this would be safe for any implementation. i'll add a comment

kafka-client/src/main/java/dev/responsive/kafka/store/RemoteWriter.java
agavra: hah... weird effect of git, this file wasn't actually renamed I created a new one. Happy to move packages though.

PR #100, opened 2023-08-11
metrics: set up a StateRestoreListener 

ableegoldman:
Sets a StateRestoreListener that logs or delegates to the user's listener if one exists. Also includes some minor refactoring to pass the `Metrics` in, though no actual metrics are added just yet

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
@@ -51,15 +51,15 @@ public class ResponsiveRestoreListener implements StateRestoreListener {
       final long startingOffset,
       final long endingOffset
   ) {
+    LOG.info("Beginning restoration from offset {} to {} for partition {} of state store {}",
+             startingOffset, endingOffset, topicPartition, storeName);
+
     if (userListener != null) {
       userListener.onRestoreStart(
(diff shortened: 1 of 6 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,102 @@
...
+      final String storeName,
+      final long totalRestored
+  ) {
+    if (userListener != null) {
+      userListener.onRestoreEnd(
+          topicPartition,
+          storeName,
+          totalRestored);
+    } else {
+      LOG.info("Finished restoration of {} total records for partition {} of state store {}",
```
rodesai, line 97: why are these log lines in the `else` clause?

rodesai (commented):
Overall LGTM. One question inline

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveRestoreListener.java
ableegoldman: I was just thinking that any/every time I've seen a user implement their own listener, it was pretty much to log exactly this.

Although there's no harm to logging something twice just to be sure. Plus, if/when we set up our own logging pipeline we'll want to make sure to get this...

I'll remove it from the else

PR #102, opened 2023-08-12
disable window stores and fix compiler warning

ableegoldman:


operator/src/main/java/dev/responsive/k8s/crd/ResponsivePolicy.java
```diff
@@ -29,4 +29,6 @@
 @Singular("responsivepolicy")
 public class ResponsivePolicy extends CustomResource<ResponsivePolicySpec, ResponsivePolicyStatus>
     implements Namespaced {
+
+  private static final long serialVersionUID = 0;
```
agavra, line 33: we should generate a real serial version UID (I think IntelliJ can do this for you)

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -161,12 +161,15 @@ public static WindowBytesStoreSupplier windowStoreSupplier(
       throw new IllegalArgumentException("Retention period cannot be less than window size");
     }
 
-    return new ResponsiveWindowedStoreSupplier(
+    final WindowBytesStoreSupplier ret = new ResponsiveWindowedStoreSupplier(
```
ableegoldman, line 164: just stashed these in a temp return variable so it's easy to re-enable these when ready

agavra (approved):

operator/src/main/java/dev/responsive/k8s/crd/ResponsivePolicy.java
ableegoldman: oh yeah, it can. nice! done

PR #103, opened 2023-08-12
fix: initialize restore consumer offsets once

ableegoldman:
Main change here is a fix in the ResponsiveRestoreConsumer, to avoid accidentally resetting any actively restoring changelogs back to their starting offset when new restoring tasks are added.

Cleaned up a few miscellaneous things I happened to notice in the various kafka clients on the side.

Also includes a few related (but minor) refactors from https://github.com/responsivedev/responsive-pub/pull/97/ to help break that up

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/DelegatingConsumer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/DelegatingConsumer.java
@@ -37,7 +37,7 @@ import org.apache.kafka.common.TopicPartition;
 
 public abstract class DelegatingConsumer<K, V> implements Consumer<K, V> {
 
-  protected final Consumer<K, V> delegate;
+  private final Consumer<K, V> delegate;
 
   public DelegatingConsumer(final Consumer<K, V> delegate) {
     this.delegate = delegate;
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
@@ -22,69 +22,56 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.regex.Pattern;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.clients.consumer.OffsetCommitCallback;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.utils.LogContext;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ResponsiveConsumer<K, V> extends DelegatingConsumer<K, V> {
   private final List<Listener> listeners;
-  private final Logger logger;
+  private final Logger log;
 
   private static class RebalanceListener implements ConsumerRebalanceListener {
-    private final Optional<ConsumerRebalanceListener> wrappedRebalanceListener;
+    private final ConsumerRebalanceListener wrappedRebalanceListener;
     private final List<Listener> listeners;
-    private final Logger logger;
-
-    public RebalanceListener(final List<Listener> listeners, final Logger logger) {
-      this(Optional.empty(), listeners, logger);
-    }
+    private final Logger log;
 
     public RebalanceListener(
         final ConsumerRebalanceListener wrappedRebalanceListener,
         final List<Listener> listeners,
-        final Logger logger
-    ) {
-      this(Optional.of(wrappedRebalanceListener), listeners, logger);
-    }
-
-    private RebalanceListener(
-        final Optional<ConsumerRebalanceListener> wrappedRebalanceListener,
-        final List<Listener> listeners,
-        final Logger logger
+        final Logger log
     ) {
       this.wrappedRebalanceListener = wrappedRebalanceListener;
       this.listeners = listeners;
-      this.logger = logger;
+      this.log = log;
     }
 
     @Override
     public void onPartitionsLost(final Collection<TopicPartition> partitions) {
-      wrappedRebalanceListener.ifPresent(l -> l.onPartitionsLost(partitions));
       for (final var l : listeners) {
-        ignoreException(logger, () -> l.onPartitionsLost(partitions));
+        ignoreException(log, () -> l.onPartitionsLost(partitions));
       }
+      wrappedRebalanceListener.onPartitionsLost(partitions);
     }
 
     @Override
     public void onPartitionsRevoked(final Collection<TopicPartition> partitions) {
-      wrappedRebalanceListener.ifPresent(l -> l.onPartitionsRevoked(partitions));
       for (final var l : listeners) {
-        ignoreException(logger, () -> l.onPartitionsRevoked(partitions));
+        ignoreException(log, () -> l.onPartitionsRevoked(partitions));
       }
+      wrappedRebalanceListener.onPartitionsRevoked(partitions);
     }
 
     @Override
     public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
-      wrappedRebalanceListener.ifPresent(l -> l.onPartitionsAssigned(partitions));
       for (final var l : listeners) {
-        ignoreException(logger, () -> l.onPartitionsAssigned(partitions));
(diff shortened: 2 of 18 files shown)


kafka-client/src/main/java/dev/responsive/kafka/clients/DelegatingConsumer.java
```diff
@@ -37,7 +37,7 @@
 
 public abstract class DelegatingConsumer<K, V> implements Consumer<K, V> {
 
-  protected final Consumer<K, V> delegate;
+  private final Consumer<K, V> delegate;
```
ableegoldman, line 40: We were haphazardly invoking the `delegate` directly in some places, and calling the `super` method in others. I'm pretty sure we just always want to delegate to the super and keep this private to protect us from ourselves

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -22,69 +22,56 @@
...
-import org.slf4j.LoggerFactory;
 
 public class ResponsiveConsumer<K, V> extends DelegatingConsumer<K, V> {
   private final List<Listener> listeners;
-  private final Logger logger;
+  private final Logger log;
 
   private static class RebalanceListener implements ConsumerRebalanceListener {
-    private final Optional<ConsumerRebalanceListener> wrappedRebalanceListener;
+    private final ConsumerRebalanceListener wrappedRebalanceListener;
```
ableegoldman, line 39: Kafka Streams should always pass in a rebalance listener when the main consumer subscribes to topics

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
```diff
@@ -22,69 +22,56 @@
...
     }
 
     @Override
     public void onPartitionsLost(final Collection<TopicPartition> partitions) {
-      wrappedRebalanceListener.ifPresent(l -> l.onPartitionsLost(partitions));
       for (final var l : listeners) {
-        ignoreException(logger, () -> l.onPartitionsLost(partitions));
+        ignoreException(log, () -> l.onPartitionsLost(partitions));
       }
+      wrappedRebalanceListener.onPartitionsLost(partitions);
```
ableegoldman, line 58: Changed the order here to make sure we always invoke our listeners, since the rebalance callbacks may occasionally throw exceptions. I figured this makes sense since if, for example, we keep hitting `onPartitionsLost` over and over because it keeps throwing some exception, then we'll want to know about it from our listeners here

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -130,7 +131,7 @@ public Consumer<byte[], byte[]> getConsumer(final Map<String, Object> config) {
     );
     // TODO: the end offsets poller call is kind of heavy for a synchronized block
     return factories.createResponsiveConsumer(
```
ableegoldman, line 133: @rodesai sorry if this is a dumb question or if I asked it before, because I' remember being confused by this before, but I really don't understand the point of the `factories` approach in this class -- AFAICT there's just the one, default/empty implementation of the interface and everything just passes the input parameters directly into the constructors. Is this intended for some future feature?

Followup question: assuming there is a reason for this, why don't we have a factory for all the clients? eg why the main consumer but not the restore?

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -1,61 +1,111 @@
...
 
   @Override
   public void assign(Collection<TopicPartition> partitions) {
+    for (final TopicPartition changelog : partitions) {
+      if (!super.assignment().contains(changelog)) {
+        uninitializedOffsets.add(changelog);
+      }
+    }
     super.assign(partitions);
-    seekInitial(partitions);
```
ableegoldman, line 53: This is the main fix -- keep `assign` as a lightweight metadata operation and just track any new partitions we received that may not yet have been initialized, then remove them from this set as we seek to the starting offset for each changelog partition.

During `poll` this gets checked and an exception is thrown if we find any uninitialized changelogs, to validate the assumptions about how the restore consumer is used within Streams.

There are a lot of alternative ways we could handle this, @rodesai I know you had another suggestion, and I'm not married to this "track offset initialization" strategy -- just picked it in the end because I think it will go well with the fix for standby tasks, where we'll be tracking offset initialization anyway. Happy to consider other options though

agavra (approved):
LGTM, I'm OK with the offset tracking mechanism. (please add a test)

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -1,61 +1,114 @@
...
     this.startOffsets = Objects.requireNonNull(startOffsets);
+    this.log = new LogContext(
+        String.format("responsive-restore-consumer [%s]", Objects.requireNonNull(clientId))
+    ).logger(ResponsiveConsumer.class);
   }
 
-  private List<TopicPartition> seekInitial(final Collection<TopicPartition> partitions) {
-    final List<TopicPartition> notFound = new LinkedList<>();
+  private Set<TopicPartition> initializeOffsets(final Collection<TopicPartition> partitions) {
+    final Set<TopicPartition> notFound = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
ableegoldman: Ah, yeah this can just be a regular HashSet. I was trying to keep it List-like because that was the original type, but looking at how it's used, HashSet will be fine too

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
ableegoldman: Added some unit tests -- was struggling a bit to find a way to test this end-to-end and could only come up with some overly complex schemes, are the unit tests good enough for now?

I might be able to write better tests for this once we have the restore metrics in place, which I'm working on now/after this

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveKafkaClientSupplier.java
rodesai: It's for letting me inject mocks when unit testing to make sure that the right listeners get installed for the right threads' clients, and to test that the ref counting works.

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveConsumer.java
rodesai: Can you leave a comment to this effect and add a test?

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
rodesai: > There are a lot of alternative ways we could handle this, @rodesai I know you had another suggestion, and I'm not married to this "track offset initialization" strategy -- just picked it in the end because I think it will go well with the fix for standby tasks, where we'll be tracking offset initialization anyway. Happy to consider other options though

I'm happy with it as long as we're checking the assumption

kafka-client/src/main/java/dev/responsive/kafka/clients/ResponsiveRestoreConsumer.java
```diff
@@ -71,16 +123,18 @@ public void seek(final TopicPartition partition, final OffsetAndMetadata offsetA
...
   @Override
   public void seekToBeginning(final Collection<TopicPartition> partitions) {
-    final Collection<TopicPartition> withUncachedPosition = seekInitial(partitions);
+    final Set<TopicPartition> withUncachedPosition = initializeOffsets(partitions);
     if (withUncachedPosition.size() > 0) {
       // Make sure to only do this if the list of partitions without cached start positions
       // is not empty. If it's empty, KafkaConsumer will seek to the actual beginning of _all_
       // partitions.
       super.seekToBeginning(withUncachedPosition);
+      uninitializedOffsets.removeAll(withUncachedPosition);
```
rodesai, line 137: I think this call won't actually do anything, because `initializeOffsets` clears all the partitions it gets. Consider clearing just the ones with cached partitions from `initializeOffsets` or just clearing them all here (just for cleanliness sake - I don't think we're doing anything functionally wrong).

PR #106, opened 2023-08-14
limit concurrent requests for fact writer

agavra:
I tested this with a large number of concurrent threads/nodes to ensure that we did not get timeouts on a basic cluster with this configuration.


ableegoldman (approved):
LGTM

PR #107, opened 2023-08-15
support TTL and TWCS

agavra:
this commit adds support for TTL to be set on all stores, and changes Fact stores to use TWCS instead of the default compaction strategy.

In order to introduce this in the API, I used @ableegoldman's suggestions of parameterizing the store suppliers with a parameters class and added a single method to `ResponsiveStores` to take advantage of this new API. This way we get the best of both worlds (we get the migration path that's straightforward) but we also get the composable APIs.


kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -70,17 +75,31 @@ public CassandraFactSchema(final CassandraClient client) {
...
+
+    if (ttl.isPresent()) {
+      final int ttlSeconds = Math.toIntExact(ttl.get().getSeconds());
+      // 20 is a magic number recommended by scylla for the number of buckets
+      final Duration compactionWindow = Duration.ofSeconds(ttlSeconds / 20);
+      return createTable.withDefaultTimeToLiveSeconds(ttlSeconds)
+          .withCompaction(
+              SchemaBuilder.timeWindowCompactionStrategy()
+                  .withCompactionWindow(
+                      compactionWindow.toMinutes(),
```
ableegoldman, line 97: uh...are we converting from Duration to `int (seconds)` back to Duration and then finally to `int (minutes)`??

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -210,6 +229,7 @@ public BoundStatement delete(
       final int partitionKey,
       final Bytes key
   ) {
+    // TODO: consider throwing an exception here as TWCS doesn't work well with deletes
```
ableegoldman, line 232: this seems like it would be extreme -- but perhaps we could/should log a warning?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKeyValueParams.java
```diff
@@ -0,0 +1,86 @@
...
+  public static ResponsiveKeyValueParams keyValue(final String name) {
+    return new ResponsiveKeyValueParams(name, false, SchemaType.KEY_VALUE);
+  }
+
+  public static ResponsiveKeyValueParams timestamped(final String name) {
+    return new ResponsiveKeyValueParams(name, true, SchemaType.KEY_VALUE);
+  }
+
+  public ResponsiveKeyValueParams timestamped() {
+    this.timestamped = true;
```
ableegoldman, line 60: If we already have distinct static constructors for the timestamped variants, I'd rather not have a separate setter API -- can we make `timestamped` final` and remove this?

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -26,6 +27,10 @@
  */
 public final class ResponsiveStores {
 
+  public static KeyValueBytesStoreSupplier keyValueStore(final ResponsiveKeyValueParams params) {
```
ableegoldman, line 30: nit: javadocs? Can pretty much just copy/paste from the others below but it would be good to clarify why/how this API is different, when to use it vs the others, etc

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -241,10 +245,7 @@ public static <K, V> StoreBuilder<TimestampedWindowStore<K, V>> timestampedWindo
   public static <K, V> Materialized<K, V, KeyValueStore<Bytes, byte[]>> materialized(
       final String name
```
ableegoldman, line 246: Should we change this to `ResponsiveKeyValueParams`? Since it's not directly analogous to any Streams APIs and no one is using it yet, I think we can just change it rather than adding a new overload. Your call though

ableegoldman (commented):
Nice -- this is a step in the right direction for sure. Few minor suggestions but otherwise LGTM

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: the TTL setting needs to be in settings (see `withDefaultTimeToLiveSeconds`) while the TWCS window needs to be in either MINUTES, HOURS or DAYS.

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: I originally had a warn log, but didn't want to log a warning on every delete as that could quickly spam the logs. I think disabling deletes and forcing a TTL could be reasonable. This is a quote from the [docs](https://opensource.docs.scylladb.com/stable/kb/compaction.html#time-window-compactionstrategy-twcs):
> Avoid overwriting data and deleting data explicitly at all costs, as this can potentially block an expired SSTable from being purged, due to the checks that are performed to avoid data resurrection.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
agavra: yup! that was on my TODO list and I forgot 🙈

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
agavra: I will go ahead and change it 👍

ableegoldman (approved):
here's your real "LGTM"

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
ableegoldman: Wait, it can't handle overwrites?? That's much different than just saying "avoid deletes", I'm a bit surprised/concerned to see that included in their warning 🤔 

W.r.t how we handle things on our end, do you think even a pattern of extremely rare deletes would still blow up the cluster significantly? Because if so then yeah, we should just throw an exception here. But if there's a chance of it being viable for some use cases (eg very low traffic in general with even more rare deletes) then I think logging a warning would be appropriate.
It would still be annoying enough that anyone abusing the ttl store with frequent deletes would be unable to ignore the flood of warnings blowing up their logs, but advanced users who truly understand the system and know what they're doing could still use it and just disable the Logger. Either way it makes them stop and think about whether they're doing the right thing here.

Of course this really hinges on there being any kind of valid use case -- if we believe any deletes whatsoever are essentially fatal then we can/should block it

ableegoldman: Like I understand _why_ this is so bad, but I'm curious _how_  bad it can get -- I guess this comes down to the size of SST tables vs frequency of deletes --> how many "dead" tables need to be kept around longer just for tombstones? I don't suppose you can get around this problem by configuring the tombstones with a very very short additional retention time (assuming that's basically how it works)

agavra: > Wait, it can't handle overwrites

it can, but it's not the intended use case and the performance can be a struggle. clearly they found it a common enough pattern to provide it OOTB and recommend it for some use cases

> W.r.t how we handle things on our end, do you think even a pattern of extremely rare deletes would still blow up the cluster significantly?

the answer to that is probably "it depends" 😕 I have to test it to get a better sense 

>  just disable the Logger.

that was also my line of thinking, but I think there may be other valuable log statements at WARN eventually from this class. we can play it by ear then or extract this into a utility class to make it easier to disable the log

> but I'm curious how bad it can get

me too! I think there's essentially two possible outcomes: leaked disk and increased CPU. We can handle the first, the second could be fatal. It's probably worth running some kind of test.

PR #108, opened 2023-08-16
Finish Responsive TTD with stubbed clients and store schemas

ableegoldman:
Fills out the TopologyTestDriver ecosystem of stubbed clients, schemas, writers, and stores with a basic in-memory map for use in a TTD test app

PR #109, opened 2023-08-16
Add ttl support to ResponsiveTTD

ableegoldman:
Followup to https://github.com/responsivedev/responsive-pub/pull/108

Finish implementing ttl for the KV store stub, and add test coverage


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStoreRegistry.java
```diff
@@ -54,6 +54,10 @@ public synchronized void deregisterStore(final ResponsiveStoreRegistration regis
     stores.remove(registration);
   }
 
+  public synchronized List<ResponsiveStoreRegistration> stores() {
+    return stores;
```
ableegoldman, line 58: I don't love having to expose this just for the TTD, we might want to take stock and try to clean up the commit logic at some point but I think it's fine for now

responsive-test-utils/src/main/java/dev/responsive/kafka/api/KVStoreStub.java
ableegoldman, line 1: not actually deleted, I just moved all the internal TTD support classes to a separate package

ableegoldman: idk why git is showing this one as deleted and re-added while it seems to have figured out that all the other classes just changed package 🤷‍♀️

responsive-test-utils/src/main/java/dev/responsive/kafka/store/KVStoreStub.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.TreeMap;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.common.utils.Time;
+import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.state.KeyValueIterator;
+
+public class KVStoreStub {
+
+  private final NavigableMap<Bytes, byte[]> records = new TreeMap<>();
+  private final Map<Bytes, Long> expiryTimes = new LinkedHashMap<>();
```
ableegoldman, line 33: The class itself is not new, but there are some significant changes here. Mainly the difference is going from a single, simple key-value map to a navigable key-value map (to support `range` and `all`) and a separate ordered map with the expiration deadline for each key.

It's pretty much a quick version of Confluent's (in)famous interview question on window processing

agavra (approved):
LGTM, I'm a bit anxious about drift (users believing they're testing production behavior for something like TTL but they're actually just testing that the test driver works). I don't have any better ideas though (well, I guess we could encourage people to do what I did and use a Cassandra/Scylla container if they really want to test production behavior).

responsive-test-utils/src/main/java/dev/responsive/kafka/api/KVStoreStub.java
agavra: I think it's enough changed content of the file as well

ableegoldman:
@agavra yeah that's fair, I had the same concerns but in the end this is how I think we should frame things:
1. if you want to test _your own_ application logic/code, use the RTTD
2. if you want to test the _Responsive infra_, you should spin up a test container or whatever
3. (if you want to test ttl behavior specifically, just don't: because we don't really provide strict semantic guarantees)

PR #110, opened 2023-08-17
Separate metadata into its own table for Fact Tables

agavra:
TWCS doesn't perform well if there are rows that are replaced. This assumption was not respected with the metadata (offset) rows, that were replaced regularly potentially messing up compaction.

The only meaningful change is in `kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java`

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
+++ b/kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
@@ -21,6 +21,7 @@ import static dev.responsive.db.ColumnName.DATA_KEY;
 import static dev.responsive.db.ColumnName.DATA_VALUE;
 import static dev.responsive.db.ColumnName.OFFSET;
 import static dev.responsive.db.ColumnName.PARTITION_KEY;
+import static dev.responsive.db.ColumnName.ROW_TYPE;
 
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.PreparedStatement;
@@ -100,6 +101,7 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
     final CreateTableWithOptions createMetadataTable = SchemaBuilder
         .createTable(metadataTable(tableName))
         .ifNotExists()
+        .withPartitionKey(ROW_TYPE.column(), DataTypes.TINYINT)
         .withPartitionKey(PARTITION_KEY.column(), DataTypes.INT)
         .withColumn(OFFSET.column(), DataTypes.BIGINT);
 
@@ -111,6 +113,7 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
     return SchemaBuilder
         .createTable(tableName)
         .ifNotExists()
+        .withPartitionKey(ROW_TYPE.column(), DataTypes.TINYINT)
         .withPartitionKey(DATA_KEY.column(), DataTypes.BLOB)
         .withColumn(DATA_VALUE.column(), DataTypes.BLOB);
   }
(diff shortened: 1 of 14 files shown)


kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -76,30 +73,43 @@ public CassandraFactSchema(final CassandraClient client) {
 
   @Override
   @CheckReturnValue
-  public SimpleStatement create(final String tableName, Optional<Duration> ttl) {
+  public void create(final String tableName, Optional<Duration> ttl) {
     LOG.info("Creating fact data table {} in remote store.", tableName);
-    final CreateTable createTable = SchemaBuilder
+    CreateTableWithOptions createTable = SchemaBuilder
```
ableegoldman, line 78: nit: maybe this is paranoid but I'm a bit nervous about non-final variables like this...could we maybe wrap the shared options in a `createTable()` method and have something like
```
final CreateTable createTable;
if (ttl.isPresent()) {
 createTable = createTable().withTTl(...).withCompaction(...);
} else {
  createTable = createTable().withCompaction(...);
}
```

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -76,30 +73,43 @@ public CassandraFactSchema(final CassandraClient client) {
...
+    // separate metadata from the main table for the fact schema, this is acceptable
+    // because we don't use the metadata at all for fencing operations and writes to
+    // it do not need to be atomic (transactional with the original table). we cannot
+    // effectively use the same table (as we do with the normal KeyValueSchema) because
+    // TWCS cannot properly compact files if there are any overwrites, which there are
+    // for the metadata columns
+    final CreateTableWithOptions createMetadataTable = SchemaBuilder
+        .createTable(metadataTable(tableName))
+        .ifNotExists()
+        .withPartitionKey(PARTITION_KEY.column(), DataTypes.INT)
```
ableegoldman, line 108: Are we going from a TINYINT to an INT here? Is that intentional?

kafka-client/src/main/java/dev/responsive/db/RemoteSchema.java
```diff
@@ -34,7 +33,7 @@ public interface RemoteSchema<K> {
    * desired schema.
    */
   @CheckReturnValue
```
ableegoldman, line 35: this should probably be removed, hm?

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -308,11 +313,9 @@ public KeyValueIterator<Bytes, byte[]> all(
...
   }
 
-  private static ByteBuffer metadataKey(final int partition) {
-    final ByteBuffer key = ByteBuffer.wrap(new byte[METADATA_KEY.get().length + Integer.BYTES]);
-    key.put(METADATA_KEY.get());
-    key.putInt(partition);
-    return key.position(0);
+  // Visible for Testing
+  static String metadataTable(final String tableName) {
+    return tableName + "_md";
```
ableegoldman, line 318: This should never change, right? Could we make this private and then hard-code this expectation in the tests to verify this?

agavra: yeah we could do that

kafka-client/src/main/java/dev/responsive/db/RemoteSchema.java
agavra: oops thanks!

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: yes, TINYINT was for the "row type", the partition key on the other hand is the Kafka Partition which is technically an integer type.

rodesai (commented):
LGTM. One question inline

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: 👍

PR #112, opened 2023-08-18
respect TTL semantically even if data is not physically cleared

agavra:
this PR changes the table schema to include a `TIMESTAMP` clustering column so that we can semantically respect the TTL based on event time even if wall clock TTL hasn't kicked in.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
+++ b/kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
@@ -22,6 +22,7 @@ import static dev.responsive.db.ColumnName.DATA_VALUE;
 import static dev.responsive.db.ColumnName.OFFSET;
 import static dev.responsive.db.ColumnName.PARTITION_KEY;
 import static dev.responsive.db.ColumnName.ROW_TYPE;
+import static dev.responsive.db.ColumnName.TIMESTAMP;
 import static dev.responsive.kafka.store.ResponsiveStoreRegistration.NO_COMMITTED_OFFSET;
 
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
@@ -35,6 +36,7 @@ import com.datastax.oss.driver.api.querybuilder.schema.compaction.TimeWindowComp
 import dev.responsive.db.partitioning.SubPartitioner;
 import java.nio.ByteBuffer;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -116,6 +118,7 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
         .ifNotExists()
         .withPartitionKey(ROW_TYPE.column(), DataTypes.TINYINT)
         .withPartitionKey(DATA_KEY.column(), DataTypes.BLOB)
+        .withClusteringColumn(TIMESTAMP.column(), DataTypes.TIMESTAMP)
         .withColumn(DATA_VALUE.column(), DataTypes.BLOB);
   }
 
@@ -187,6 +190,7 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
             .insertInto(tableName)
             .value(ROW_TYPE.column(), RowType.DATA_ROW.literal())
             .value(DATA_KEY.column(), bindMarker(DATA_KEY.bind()))
+            .value(TIMESTAMP.column(), bindMarker(TIMESTAMP.bind()))
             .value(DATA_VALUE.column(), bindMarker(DATA_VALUE.bind()))
             .build()
     ));
@@ -197,6 +201,7 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
             .columns(DATA_VALUE.column())
             .where(ROW_TYPE.relation().isEqualTo(RowType.DATA_ROW.literal()))
             .where(DATA_KEY.relation().isEqualTo(bindMarker(DATA_KEY.bind())))
+            .where(TIMESTAMP.relation().isGreaterThanOrEqualTo(bindMarker(TIMESTAMP.bind())))
             .build()
     ));
 
@@ -258,14 +263,14 @@ public class CassandraFactSchema implements RemoteKeyValueSchema {
    * Inserts data into {@code table}. Note that this will overwrite
    * any existing entry in the table with the same key.
    *
-   * @param table         the table to insert into
-   * @param partitionKey  the partitioning key
-   * @param key           the data key
-   * @param value         the data value
-   *
+   * @param table        the table to insert into
+   * @param partitionKey the partitioning key
+   * @param key          the data key
+   * @param value        the data value
+   * @param timestamp    the event time of the data
    * @return a statement that, when executed, will insert the row
-   *         matching {@code partitionKey} and {@code key} in the
-   *         {@code table} with value {@code value}
+   * matching {@code partitionKey} and {@code key} in the
+   * {@code table} with value {@code value}
    */
   @Override
   @CheckReturnValue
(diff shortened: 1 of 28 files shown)


kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
```diff
@@ -169,7 +169,7 @@ public boolean persistent() {
 
   @Override
   public void put(final Bytes key, final byte[] value) {
-    buffer.put(key, value);
+    buffer.put(key, value, context.timestamp());
```
agavra, line 172: cc @ableegoldman - is this timestamp valid? this way we won't rely on the timestamp KV store functionality at all

kafka-client/src/main/java/dev/responsive/utils/Iterators.java
```diff
@@ -284,6 +291,42 @@ public T next() {
     }
   }
 
+  private static class KeyMappingIterator<I, O, V> implements KeyValueIterator<O, V> {
```
agavra, line 294: not used - I used this in the original implementation when I was client-side filtering but dropped that approach. Either way, this is a worthwhile utility iterator so I'm keeping it

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
ableegoldman: Ah...sort of? I believe this is the timestamp of the record that was read from the input topic, which is probably good enough but definitely not strictly-speaking correct. 

Some additional context in this KIP (the relevance being that we used to use this method to get a timestamp when one was not provided to a `WindowStore#put` API, but we deprecated/removed  that API entirely because this wasn't semantically sound -- https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=115526545

ableegoldman: Like I said though, for our purposes this is probably ok. If someone ever complains about it we can consider extracting it from the `TimestampAndValue` for timestamped stores to improve the semantics

The one thing I'm not certain of is what happens if you try to invoke this from a punctuator (where there is no "input record"). I'd recommend writing a quick test for this just to make sure it doesn't throw an exception or return complete garbage

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -196,6 +200,10 @@ public void prepare(final String tableName) {
             .columns(DATA_VALUE.column())
             .where(ROW_TYPE.relation().isEqualTo(RowType.DATA_ROW.literal()))
             .where(DATA_KEY.relation().isEqualTo(bindMarker(DATA_KEY.bind())))
+            .where(TIMESTAMP.relation().isGreaterThanOrEqualTo(bindMarker(TIMESTAMP.bind())))
+            // ALLOW FILTERING is OK b/c the query only scans one partition (it actually  only
+            // returns a single value)
+            .allowFiltering()
```
ableegoldman, line 206: Can you elaborate on this comment -- what does `.allowFiltering()` do/mean and why would it not be "ok"?

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -257,44 +265,47 @@ public BoundStatement delete(
...
    */
   @Override
   @CheckReturnValue
   public BoundStatement insert(
       final String table,
       final int partitionKey,
       final Bytes key,
-      final byte[] value
+      final byte[] value,
+      long timestamp
```
ableegoldman, line 284:
```suggestion
      final long timestamp
```

ableegoldman: Don't want to leave a comment on every single one but same goes for all the below

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsivePartitionedStore.java
ableegoldman: If it helps, I just checked the code and this is what we use for the changelog timestamp, so it's actually probably more correct than the "real" message timestamp (for our purposes)

kafka-client/src/main/java/dev/responsive/kafka/store/SizeTrackingBuffer.java
```diff
@@ -41,6 +41,8 @@ public NavigableMap<K, Result<K>> getReader() {
   }
 
   private long sizeOf(final K key, final Result<K> value) {
-    return extractor.bytes(key).get().length + (value.isTombstone ? 0 : value.value.length);
+    return extractor.bytes(key).get().length
+        + (value.isTombstone ? 0 : value.value.length)
+        + Long.BYTES; // timestamp size
```
ableegoldman, line 46: nit: can we move this method to the `Result` class? Feels way too easy to miss updating the size computation if/when the format changes

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: https://www.datastax.com/blog/allow-filtering-explained - if you don't have this you can't use `WHERE` clauses on non-clustering/primary key columns. I don't think it makes sense to elaborate too much in the comment, as this class in general assumes C* knowledge

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: oops... I need to change my IntelliJ settings back after some plugin messed it up

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -197,6 +201,10 @@ public void prepare(final String tableName) {
             .columns(DATA_VALUE.column())
             .where(ROW_TYPE.relation().isEqualTo(RowType.DATA_ROW.literal()))
             .where(DATA_KEY.relation().isEqualTo(bindMarker(DATA_KEY.bind())))
+            .where(TIMESTAMP.relation().isGreaterThanOrEqualTo(bindMarker(TIMESTAMP.bind())))
```
rodesai, line 204: are you sure bloom filtering still gets used when doing scans? I think it should, but maybe good to double check.

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -258,44 +266,47 @@ public BoundStatement delete(
...
    */
   @Override
   @CheckReturnValue
   public BoundStatement insert(
       final String table,
       final int partitionKey,
       final Bytes key,
-      final byte[] value
+      final byte[] value,
+      final long timestamp
```
rodesai, line 285: nit: rename to epochMillis and clarify that's what it is in the javadoc

rodesai (commented):
What do we want to happen if a user turns TTL on on an existing store that doesn't use TTL? I  think given the implementation here we'd just not return anything for the existing rows. It's probably worth adding a check when initializing that ensures the table configuration in scylla matches the one expected by code.

rodesai:
> What do we want to happen if a user turns TTL on on an existing store that doesn't use TTL? I think given the implementation here we'd just not return anything for the existing rows. It's probably worth adding a check when initializing that ensures the table configuration in scylla matches the one expected by code.

Actually I think it will just fail because the TIMESTAMP column won't exist for queries. Still might be nice to have more explicit handling for this kind of thing (a table config changing after creation) in a follow-up.

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
agavra: I ran a perf test over the weekend and it did pretty well, I'm pretty sure the bloom filtering is still applied as BF is applied just on the partition key to see whether the partition is present or not in an SST and we still have a full partition key clause. Unfortunately I don't see any docs indicating one way or another...

agavra:
> Actually I think it will just fail because the TIMESTAMP column won't exist for queries. Still might be nice to have more explicit handling for this kind of thing (a table config changing after creation) in a follow-up.

The timestamp column will be there either way (assuming they created the table on this version, which I think is fine). The physical expiration TTL, OTOH, won't be applied... we can check and log a warning for this. I'll do that in the follow-up.

PR #113, opened 2023-08-30
refactor: change some of the names of our policy objects and un-nest:

rodesai:
- Add a "Spec" suffix to all the proto and policy models to differentiate the specifications from their implementations (which are in the controller).
- Move the diagnoser stuff out from inside the policy model object in the crd model (just for cleanliness sake)

The obvious concern here is compatibility. Lets look at a couple different compatibility scenarios and see how they are handled:

- An old operator talking to a controller with the new policy protos. The new proto spec is wire-compatible with the spec that uses the old names. The messages themselves rely on the field numbers to identify fields. So the controller can read messages from an Operator that is using code generated using the old names.
- An old controller getting messages from a new operator. Works for the same reason as above.


controller-api/src/main/proto/responsive/controller/v1/controller.proto
```diff
@@ -73,32 +73,34 @@ enum PolicyStatus {
 
 // TODO(rohan): move this to a a common location where we specify the policy
 //              also, see if we can combine with the CRD somehow
-message ApplicationPolicy {
+message ApplicationPolicySpec {
```
agavra, line 76: can we encode a version somewhere in the spec? I'd feel better if we could throw exceptions if the versions mismatch

agavra (approved):
LGTM

controller-api/src/main/proto/responsive/controller/v1/controller.proto
agavra: actually, I think it makes more sense to have it at the top level (wherever that schema is) - let me know if we already have one.

rodesai: yeah the proto file itself is versioned. So if we want to change the protocol we can just add a new set of protos under `controller-api/src/main/proto/responsive/controller/v2` and name the service something like `ControllerV2`

PR #114, opened 2023-08-30
add controller models for new policies

rodesai:
add models for mean sojourn time and saturation diagnosers

mean sojourn time diagnoser: this diagnoser takes a target for "mean sojourn time"
   and will try to make sure the application can satisfy this target. The mean
   sojourn time is the time expected to process a record appended to a source
   topic partition.

saturation diagnoser: this diagnoser attemmpts to keep thread "saturation" over
   some target. Saturation is the percentage of time a given thread spends not in poll
   (e.g. processing data). A low value means the thread is idle.

### DIFF
--- a/operator/src/main/java/dev/responsive/k8s/crd/kafkastreams/ScaleUpStrategySpec.java
+++ b/operator/src/main/java/dev/responsive/k8s/crd/kafkastreams/ScaleUpStrategySpec.java
@@ -1,0 +1,45 @@
+package dev.responsive.k8s.crd.kafkastreams;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class ScaleUpStrategySpec {
+  public enum Type {
+    FIXED_REPLICA,
+    RATE_BASED,
+    SCALE_TO_MAX
+  }
+
+  private Type type;
+  private Optional<FixedReplicaScaleUpStrategySpec> fixedReplica;
+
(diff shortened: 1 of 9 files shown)


operator/src/main/java/dev/responsive/k8s/crd/kafkastreams/ScaleUpStrategySpec.java
```diff
@@ -0,0 +1,45 @@
+package dev.responsive.k8s.crd.kafkastreams;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class ScaleUpStrategySpec {
+  public enum Type {
```
agavra, line 7: do we need some kind of default enum for UNKNOWN for forwards compatibility?

agavra (approved):
lgtm

operator/src/main/java/dev/responsive/k8s/crd/kafkastreams/ScaleUpStrategySpec.java
rodesai: This gets run when the policy is read from k8s and loaded into the operator. So we'd hit this case if a user upgrades, specifies a policy that uses some new scaling strategy, and then downgrades operator. In this case we can either have the operator reject the policy and then the user can re-issue it, or choose some default scaling behavior. I'd rather just keep things simple and do the former for now.

PR #115, opened 2023-08-30
rename to Kafka Streams policy

rodesai:
rename Demo to KafkaStreams

Renames the Demo policy to Kafka Streams. Originally, the intent
here was to throw away the demo policies. As we've gone along, I
think a better approach is to retain the existing model of a policy
with pluggable diagnosers that each try to drive the application to
meet their own goals.

The obvious concern here is compatibility. Lets look at a few different
compatibility scenarios and see how they are handled:

- An old operator talking to a controller with the new policy protos.
  The new proto spec is wire-compatible with the spec that uses the
  demo name. The messages themselves rely on the field numbers to identify
  fields. So the controller can read messages from an Operator that is
  using code generated using the demo names.
- An old controller getting messages from a new operator. Works for the
  same reason as above.
- A new operator reading the existing policy objects in K8s. The deserializer
  constructor will still get a demo policy. It internally converts it to
  a Kafka Streams policy (see the ResponsivePolicySpec constructor). Policy
  CRDs are only deserialized by the operator. Even if they were serialized,
  the result would simply be a converted crd that is functionally equivalent.
- An old operator reading a new policy (e.g. a downgrade). This wont work.
  In this case the user could just respecify the policy.


agavra (approved):
Just skimmed

agavra (approved):
skimmed, approving based on description

PR #118, opened 2023-09-01
More pre-window store cleanup

ableegoldman:
No real fundamental changes here, just some minor refactoring to clean things up and consolidate the difference store types a bit before the main window store implementation


kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
```diff
@@ -145,19 +150,59 @@ public OptionalInt numPartitions(final String tableName) {
...
+        throw new IllegalArgumentException(params.schemaType().name());
     }
+
+    schema.create(params.name().cassandraName(), Optional.empty());
+    awaitTableAndPrepareSchema(params.name().cassandraName(), schema);
+
+    return schema;
+  }
+
+  private void awaitTableAndPrepareSchema(
```
ableegoldman, line 199: basically moved all the table creation and schema prep to the client so we don't accidentally update eg the KV store but not the window store

Eventually I'll probably do a larger refactor to pull everything else shared in the stores to a common superclass, but I'm waiting to finish the window store stuff first

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
```diff
@@ -234,29 +234,27 @@ public class ResponsiveConfig extends AbstractConfig {
...
-    public String toString() {
-      return "non-empty password";
-    }
+  /**
+   * This should generally be used over the {@link #loggedConfig(Map)} override in all
+   * cases outside the initial {@link dev.responsive.kafka.api.ResponsiveKafkaStreams#create}.
+   * Prefer this API whenever you need to construct a {@code ResponsiveConfig} mid-application,
+   * for example during state store init.
+   */
+  public static ResponsiveConfig quietConfig(final Map<?, ?> originals) {
```
rodesai, line 252: nit: maybe just call this `config` so that it's perceived as the default to use

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
```diff
@@ -43,7 +46,9 @@
  */
 public class CassandraClient {
 
+  private final ScheduledExecutorService executor = new ScheduledThreadPoolExecutor(2);
```
rodesai, line 49: why keep the thread pool isolated to the cassandra client? It's probably not a big deal, but since we're embedding ourselves into customers' apps we should probably be a little conservative about creating thread pools.

rodesai (commented):
Mostly LGTM. A couple thoughts inline

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
ableegoldman: We don't use it anywhere else -- at the moment it's all internal to the CassandraClient. I moved it here as part of consolidating the schema creation/prep in each store to the CassandraClient which is where/why the executor is needed. Seemed silly to pass it around through the configs and extract it to the StateStores just to ultimately end up only needing it for the CassandraClient, then we might as well just move it there.

As an aside, we do create separate executors for the EndOffsetsPoller, but that's different.

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
ableegoldman: good call

kafka-client/src/main/java/dev/responsive/db/CassandraClient.java
ableegoldman: Going to go ahead and merge this once the tests pass, but if you're worried about this I can do a quick patch to move it back

That said, I'm envisioning that if we do end up needing an executor outside the CassandraClient class, we can/should still just access it via a standard getter on CassandraClient so that it's still clear who "owns" the executor and its lifecycle/scope. But like I said, if you're concerned we might accidentally forget about the executor here and end up creating more and blowing up the thread count, I'm happy to do a short followup

PR #119, opened 2023-09-02
Support null-value deletes 

ableegoldman:
Both KV and Window stores allow deletion via put with a null value in Kafka Streams. We need to handle these properly as tombstones and enable deletes for window stores

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/db/CassandraWindowedSchema.java
+++ b/kafka-client/src/main/java/dev/responsive/db/CassandraWindowedSchema.java
@@ -72,6 +72,7 @@ public class CassandraWindowedSchema implements RemoteWindowedSchema {
   // if the key is absent, else not at all. this guarantee is not present
   // in all implementations of ConcurrentMap
   private final ConcurrentHashMap<String, PreparedStatement> insert;
+  private final ConcurrentHashMap<String, PreparedStatement> delete;
   private final ConcurrentHashMap<String, PreparedStatement> fetchSingle;
   private final ConcurrentHashMap<String, PreparedStatement> fetch;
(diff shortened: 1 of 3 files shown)


kafka-client/src/main/java/dev/responsive/db/CassandraWindowedSchema.java
```diff
@@ -98,7 +100,7 @@ public CassandraWindowedSchema(final CassandraClient client) {
 
   @Override
   public void create(final String name, Optional<Duration> ttl) {
-    // TODO: explore better data models for fetchRange/fetchAll
+    // TODO(window): explore better data models for fetchRange/fetchAll
```
rodesai, line 103: ultra-nit: usually this notation is used to make note of who owns the todo (e.g. TODO(sophie))

rodesai (commented):
The code change LGTM. We should add unit tests for this stuff or its going to be hard to confidently refactor/change things later. Definitely don't want to block this change on that though.

kafka-client/src/main/java/dev/responsive/db/CassandraWindowedSchema.java
ableegoldman: yeah, I guess it's technically targeted at myself but I just wanted to mark everything I want to hit in the next window store PR. so it should be very temporary

ableegoldman:
Thanks, I'm going to merge it for now but totally agree on the tests. I was hoping to just throw in a quick unit test for this but found out we don't really have a unit test framework for the state stores yet, so it's kind of a bigger task than I wanted to take on just yet. I'll file a ticket for that though

PR #124, opened 2023-09-07
fix: initialize TTDMockAdmin with the changelogs to create up front

ableegoldman:
The main fix here is in how we "create" topics within the MockAdmin used by the RTTD. Because no real topics are created in the TTD context, everything has to mock its own internal view of the "cluster" which in the admin case, means making sure that it adds any topics that we will later make requests for. For us, this is just changelogs (and should generally only ever be changelogs, as the input topics are completely abstracted away by the TTD entirely including everything like the KafkaClientSupplier)

Before this, we would lazily create topics when requested for in the #describeTopics API. However for FACT tables we skip this particular call, and so incur an error when we next hit the mock admin in the #describeConfigs call.

The fix is to remove any topic-creation inside Admin overrides (since this is extremely fragile and relies on us either overriding every method that is ever added to it or else being super diligent about keeping the TTDMockAdmin up to date with any admin calls ever added to our codebase). Instead, we pass in the appId and topology to the TTDMockAdmin's constructor so that it can derive the changelog topic names from the list of state store names, and add everything up front. Note that we can't get a handle on the changelog topic names directly until after the topology is initialized, and unfortunately this does not occur until we're already inside the TopologyTestDriver constructor, where the exception in this bug originates.

However, as a bit of an added safety measure, I also added a check after we exit the TTD constructor and can make sure that all the changelogs were created as expected. This doesn't really do much, but it'll be nice to have as a warning in case something does change that invalidates any assumption in this fix

### DIFF
--- a/responsive-test-utils/src/main/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriver.java
+++ b/responsive-test-utils/src/main/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriver.java
@@ -16,31 +16,23 @@
 
 package dev.responsive.kafka.api;
 
+import dev.responsive.kafka.clients.TTDCassandraClient;
+import dev.responsive.kafka.clients.TTDMockAdmin;
 import dev.responsive.kafka.config.ResponsiveConfig;
-import dev.responsive.kafka.store.CassandraClientStub;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Collection;
-import java.util.Collections;
 import java.util.Properties;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
-import org.apache.kafka.clients.admin.DescribeTopicsResult;
-import org.apache.kafka.clients.admin.MockAdminClient;
-import org.apache.kafka.common.Node;
-import org.apache.kafka.common.TopicPartitionInfo;
-import org.apache.kafka.common.config.TopicConfig;
-import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.MockTime;
 import org.apache.kafka.streams.StreamsConfig;
-import org.apache.kafka.streams.TestInputTopic;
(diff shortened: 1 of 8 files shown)


responsive-test-utils/src/main/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriver.java
```diff
@@ -16,31 +16,23 @@
...
 import org.apache.kafka.common.utils.MockTime;
 import org.apache.kafka.streams.StreamsConfig;
-import org.apache.kafka.streams.TestInputTopic;
 import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.TopologyDescription;
 import org.apache.kafka.streams.TopologyTestDriver;
-import org.apache.kafka.streams.test.TestRecord;
 
 public class ResponsiveTopologyTestDriver extends TopologyTestDriver {
+  public static final String RESPONSIVE_TTD_ID = "Responsive_TopologyTestDriver";
```
agavra, line 33: will there be any issues if this is reused? would it make sense to add a random id to it like the TTD does?

agavra (commented):
LGTM

responsive-test-utils/src/main/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriver.java
ableegoldman: Nah, I can't imagine how there could be -- there's no real "cluster", everything is internal to the TTD. Even the StateDirectory is wiped when the TTD is closed

I guess if you have really poor test practices and don't ensure the TTD is closed, have a persistent/rocksdb store somewhere in your topology alongside the Responsive stores, AND set the state directory yet not the app id (because default state.dir involves randomly generated name IIRC), then this could be a "regression" -- but at that point, no one can help you lol

PR #125, opened 2023-09-08
Align kafka-client and responsive-test-utils versions

ableegoldman:
Main fix is to define a "module group" and add some github actions so that we can execute parts of the `Publish Artifacts` workflow on each module in the group.


rodesai (approved):
LGTM! Nice find on the actions

PR #126, opened 2023-09-08
Extract AWS/ECR login steps to separate Github action

ableegoldman:
While introducing some actions to tackle the module groups/versioning stuff, I happened to notice that there are quite a few other kinds of highly repetitive procedures within our three workflows. It would be nice to extract as much of these as possible into separate actions that can be called from workflow jobs, whether new or while updating the existing ones.

Should also help to make the workflows a bit easier to read/comprehend

### DIFF
--- a/.github/workflows/github-pr.yaml
+++ b/.github/workflows/github-pr.yaml
@@ -20,25 +20,13 @@ jobs:
           java-version: '11'
           distribution: 'temurin'
 
-      - name: Configure AWS Credentials for Public ECR Login
-        uses: aws-actions/configure-aws-credentials@v2
+      - name: "Login to AWS & ECR"
+        uses: ./.github/actions/ecr-login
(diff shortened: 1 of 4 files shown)


.github/workflows/github-pr.yaml
```diff
@@ -53,9 +41,12 @@ jobs:
           ref: refs/heads/master
           token: ${{ secrets.TOOLS_GHA_TOKEN }}
 
-      - name: Login to Amazon ECR
-        id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
+      - name: "Login to AWS & ECR"
+        uses: ./.github/actions/ecr-login
```
rodesai, line 45: nit: rename the action to `aws-login` or `aws-and-ecr-login`. `ecr` specifically refers to the container registry

PR #129, opened 2023-09-12
Fix publishing workflow with module groups

ableegoldman:


ableegoldman:
Got a successful build of the `kafka-client` group with aligned artifacts (kafka-client and responsive-test-utils)

Screenshot 2023-09-12 at 4 46 01 PM

PR #130, opened 2023-09-14
Cache partition counts for sub-partitioner and set definitions/naming conventions

ableegoldman:
The main change here is the introduction of the `CachedMetadata` class, which as the name suggests caches some of the nontrivial repeated requests for things like the number of kafka partitions or sub-partitions in Cassandra. Also cleans up the sub-partitioner and related APIs a bit, mainly with regards to naming. We should try to stick to the conventions outlined in the SubPartitioner javadocs, as it gets a bit confusing to keep track of which "partition" is being referred to when.

This is related to the refactoring I need for the window store implementation, but  indirectly enough that it makes sense to pull out into its own patch.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/utils/CachedMetadata.java
+++ b/kafka-client/src/main/java/dev/responsive/utils/CachedMetadata.java
@@ -1,0 +1,68 @@
+package dev.responsive.utils;
+
+import static dev.responsive.db.ColumnName.PARTITION_KEY;
+
+import com.datastax.oss.driver.api.core.cql.ResultSet;
+import dev.responsive.db.CassandraClient;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.OptionalInt;
+import java.util.concurrent.ExecutionException;
+import org.apache.kafka.clients.admin.Admin;
+
+/**
+ * Use this any time we need to repeatedly look up metadata that doesn't change, so we can
+ * cache the results rather than make admin/client requests every single time
+ */
+public class CachedMetadata {
+
+  private final Admin admin;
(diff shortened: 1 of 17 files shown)


responsive-test-utils/build.gradle.kts
```diff
@@ -21,11 +21,7 @@ plugins {
 version = project(":kafka-client").version
 
 dependencies {
-    // TODO: we should make sure to fail the release of kafka-client
-    // if we don't bump this up - or have some other mechanism to
-    // ensure they are released in tandem - otherwise it is possible
-    // that the latest release of kafka-client will be untestable
-    implementation("dev.responsive:kafka-client:0.7.1")
+    implementation(project(":kafka-client"))
```
ableegoldman, line 24: Just FYI, the responsive-test-utils module should now be fully tied to the kafka-client module, both in terms of dependencies and releases.

kafka-client/src/main/java/dev/responsive/utils/CachedMetadata.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map;
+import java.util.OptionalInt;
+import java.util.concurrent.ExecutionException;
+import org.apache.kafka.clients.admin.Admin;
+
+/**
+ * Use this any time we need to repeatedly look up metadata that doesn't change, so we can
+ * cache the results rather than make admin/client requests every single time
+ */
+public class CachedMetadata {
```
rodesai, line 18: This class seems to be shared by all the stores, so it needs to be synchronized.

rodesai (commented):
Usage of the cache looks good to me, but the cache itself needs concurrency control.

PR #133, opened 2023-09-15
Make changelog truncation configurable

ableegoldman:
This PR makes changelog truncation a store-level config, rather than something we implicitly derive/guess at the user intention based on the topic configs. This solves a few things that have been on my mind for a while

1. Don't accidentally wipe out users don't when they weren't expecting it 💀 
2. Make it a top-level config so it's easier to notice for people who _do_ want it
3. Since we currently aren't compatible with the source-topic optimization or disabling the changelog, we can point users who are concerned about changelog topic space/growth to use changelog truncation instead
4. Best of all, no need to make so many admin calls during state store initialization (with this and https://github.com/responsivedev/responsive-pub/pull/130, we now have no superflous admin calls)


kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveStores.java
```diff
@@ -274,21 +293,6 @@ public static <K, V> StoreBuilder<TimestampedWindowStore<K, V>> timestampedWindo
...
-  /**
-   * Create a {@link Materialized} that can be used to build a Responsive {@link KeyValueStore}
-   * and materialized in the DSL. If using the low-level Processor API, use
-   * {@link #keyValueStoreBuilder} instead.
-   *
-   * @param params the store parameters
-   * @return a Materialized configuration that can be used to build a key value store with the
-   *         given options that uses Responsive's storage for its backend
-   */
-  public static <K, V> Materialized<K, V, KeyValueStore<Bytes, byte[]>> materialized(
```
ableegoldman, line 286: no actual changes here, I just moved all the kv store stuff into one place

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveMaterialized.java
```diff
@@ -17,14 +17,27 @@ public ResponsiveMaterialized(
...
     super(materialized);
     this.truncateChangelog = truncateChangelog;
+
+    // Truncation uses the admin's delete request, which seems to require the cleanup.policy
+    // to contain 'delete'
     if (truncateChangelog) {
-      topicConfig.put(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_DELETE);
+      topicConfig.put(
+          TopicConfig.CLEANUP_POLICY_CONFIG,
+          TopicConfig.CLEANUP_POLICY_COMPACT + "," + TopicConfig.CLEANUP_POLICY_DELETE
```
ableegoldman, line 26: Small change here -- when truncation _is_ enabled, I think it makes sense to configure it with `compact, delete` rather than just `delete`, since the default is `compact`. And at this point, only the default configs are enabled -- to actually change any configs you have to go through the `withLoggingEnabled` API, where we have a separate validation but otherwise let users override it as they please

Note that these configs are only applied if the changelog is created anew. For existing changelogs, if they are not `delete` enabled, we will just fail the `deleteRecords` request in the CommitBuffer and then turn off further truncation attempts.

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
```diff
@@ -402,13 +405,19 @@ private RemoteWriteResult drain(final Collection<RemoteWriter<K>> writers) {
...
       try {
         admin.deleteRecords(Map.of(changelog, RecordsToDelete.beforeOffset(offset))).all().get();
         log.info("Truncated changelog topic {} before offset {}", changelog, offset);
       } catch (final ExecutionException e) {
-        log.warn("Could not truncate changelog topic-partition {}.", changelog, e);
+        log.warn("Could not truncate changelog topic-partition " + changelog, e);
+        if (e.getCause() instanceof PolicyViolationException) {
+          log.info("Disabling further changelog truncation attempts due to topic configuration "
+                       + "being incompatible with deleteRecord requests", e);
+          isNotDeleteEnabled = true;
```
rodesai, line 417: is the reason for doing this in the exception handler to avoid admin calls during store initialization?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveMaterialized.java
rodesai: if there is not existing topic why not just do the most sensible thing for our store? with just `delete` there's zero risk of losing tombstones right? What's the benefit of staying consistent with how vanilla streams would configure the changelog?

rodesai (approved):
Just a couple questions inline. Nothing I'd block merging on.

kafka-client/src/main/java/dev/responsive/kafka/store/CommitBuffer.java
ableegoldman: Yeah, basically. Most cases I think it should be unnecessary to check at all, so doing it on every store init is pretty costly. Ideally we could just check once upfront, when the app first starts up, but atm there's not really a good way/place to do this, because initialization is the first place where we have all the necessary ingredients all together: the changelog name and the store params plus the admin (which is extracted through the configs that only get passed in to #init)

Anyways, I think it's always possible for users to change their topic cleanup.policy on-the-fly, so they could hit this exception even if they passed any startup checks. I think it's better to just permanently disable changelog truncation in a case like that, rather than cripple the application with all these failing requests. If they want to re-enable changelog truncation after changing the topic configs back, they can simply bounce the app. That kind of double-turnaround seems particularly unlikely so I feel like a restart is a warranted solution

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveMaterialized.java
ableegoldman: I actually do think `delete, compact` _is_ what we want for our stores, with changelog truncation enabled. Here's how I was reasoning about it, but definitely think about it and whether you agree:

1. I think the benefit is undeniable, though perhaps negligible outside of certain edge cases. But imagine an app with mostly all overwrites of the same key. If we end up crashing/failing the flush/getting fenced/etc and there's a large buildup of single-key overwrites, with just `delete` we'll have to consume and write all of these as the restoration path doesn't deduplicate records in any way (like our CommitBuffer does). Whereas with `delete, compact` we only have to restore the latest value (per key), which is the same as what our CommitBuffer would be trying to flush had it gone through normal processing instead of restoration. I'm worried we could get cascading failures (maybe a bit extreme but you get the idea) if the data to be written via restoration is potentially orders of magnitude larger than the write from a CommitBuffer flush
Similarly, in the case of repetitive commit failures under ALOS, we'll rewind to the point of last flush/commit and then rewrite everything in the changelog each time before attempting the (doomed) commit flush. In that case, without compaction we'd be digging a deeper and deeper grave as each failure meant another full set of data that need to be restored and written to Cassandra (even though it's actually just repeated sequences)
2. I believe the risk is negligible, because we aggressively truncate after each flush which means the changelog only ever contains the current "working set" of data since the last successful flush/commit. It's hard to imagine an application that could be successful with a flush interval on the order of 24 hours, which is how long tombstones are kept around for by default (and we could override/validate this config to be sure, or even set it to 7 days to feel extra safe).
3. Somewhat orthogonal, I'm a little worried that users might enable changelog truncation without realizing that it means turning on time-based topic retention, and could lose data if their application has long downtimes beyond the default of 7 days. I'm wondering if we should actually set an infinite retention time, and simply rely on the truncation and compaction semantics to keep the changelog topic growth in check

PR #134, opened 2023-09-16
Convert to pure ResponsiveKafkaStreams constructor setup

ableegoldman:
This PR changes the way applications are created from the somewhat awkward static `ResponsiveKafkaStreams.connect` constructors, to something that is fully analogous with vanilla KafkaStreams basic constructors (and thus easily migrated via find-and-replace).

It was a huge hassle and takes a really long sequence of constructors with iterative parameters, but it was worth it I think.
I did leave in the one static constructor that I believe is currently being used, and will remove it in a separate PR I have planned with other changes that will break our current user's code and need to be updated, so we can do that when everyone is ready


ableegoldman:
Tests passed locally and I'm a bit blocked on this, so going to go ahead and merge. Feel free to leave any feedback for me to address in a followup (although I think everyone will be happy, especially Almog who wanted this all along 😄 )

PR #136, opened 2023-09-18
Finalize ResponsiveKafkaStreams API and clean up StreamsConfig

ableegoldman:
Get this ready to publish the API Reference docs. Fills in a few niche missing constructors (KafkaStreams allows you to pass in a `Time` object), adds javadocs for everything, and cleans up the way we handle StreamsConfigs and validate the properties.

Introduces a new ResponsiveStreamsConfig class so we can do "quiet logging". Will follow up with the same for the other kafka clients that we're excessively logging


ableegoldman:
Ran the tests locally and they succeeded so I'm just going to merge so I can publish a new artifact. Will follow up on any comments if necessary

PR #138, opened 2023-09-19
emit -1 as the committed offset metric for newly added tasks

rodesai:
Changes MetricsPublishingCommitListener to emit -1 for for the committed offset before the first commit of a newly added task. This lets the controller distinguish between the application not knowing the committed offset and the application not sending metrics.

As part of this change, we pass in the consumer group ID to the listener rather than using the callback to get it


agavra (approved):
LGTM

PR #142, opened 2023-09-21
Log version and git commit id during startup

ableegoldman:
Added a task to the gradle build that stashes some various useful git properties in a file along with the main build. We can then read this out to log valuable debugging info such as the version and specific commit id of the dependency.

Tested this out both in a local kafka-client integration test, and by pulling in a local build with the fix while running a test in the soak repo. Works with both, and logs for the TTD as well as real applications.

Useful for sanity checks, both in the soak test and future customer logs where we want to verify exactly what version they're running. I only put it in the kafka-client module since that's where it seems most relevant, but we could extend it to others as well if that feels right.

### DIFF
--- a/kafka-client/build.gradle.kts
+++ b/kafka-client/build.gradle.kts
@@ -50,34 +50,6 @@ tasks.register(writeVersionPropertiesFile) {
     }
 }
 
-// Need to add all generated resource directories as outputs of the main sourceSet
-sourceSets {
-    main {
-        output.dir(file("$buildDir/resources"))
-    }
-}
-
-
-// TODO: there has GOT to be a more concise way to do this than listing it out for every task individually...
-//  might be overkill with some of these already depending on each other so we can probably clean up a bit too
-tasks.release {
-    dependsOn(tasks[writeVersionPropertiesFile])
-}
-
-tasks.publish {
-    dependsOn(tasks[writeVersionPropertiesFile])
-}
-
-tasks.publishToMavenLocal {
-    dependsOn(tasks[writeVersionPropertiesFile])
-}
-
-tasks.jar {
-    dependsOn(tasks[writeVersionPropertiesFile])
-    dependsOn(tasks.processResources)
-    dependsOn(tasks.processTestResources)
-}
-
 tasks.compileJava {
     dependsOn(tasks[writeVersionPropertiesFile])
     dependsOn(tasks.processResources)
--- a/kafka-client/build.gradle.kts
+++ b/kafka-client/build.gradle.kts
@@ -57,6 +57,21 @@ sourceSets {
     }
 }
 
+
+// TODO: there has GOT to be a more concise way to do this than listing it out for every task individually...
+//  might be overkill with some of these already depending on each other so we can probably clean up a bit too
+tasks.release {
+    dependsOn(tasks[writeVersionPropertiesFile])
+}
+
+tasks.publish {
+    dependsOn(tasks[writeVersionPropertiesFile])
+}
+
+tasks.publishToMavenLocal {
+    dependsOn(tasks[writeVersionPropertiesFile])
+}
+
 tasks.jar {
     dependsOn(tasks[writeVersionPropertiesFile])
     dependsOn(tasks.processResources)
--- a/kafka-client/build.gradle.kts
+++ b/kafka-client/build.gradle.kts
@@ -16,7 +16,19 @@
 
 plugins {
     id("responsive.java-library-conventions")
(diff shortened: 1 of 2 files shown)


ableegoldman:
One issue to resolve before this can/should be merged:  I added a plugin that writes the versioning info to a file in the build (so we can log it). When running some integration tests to verify that, I noticed that multiple classpath.index files are now created: once under kafka-client/build/resources/test alongside some other test-specific resources (eg log4j config), but now also under kafka-client/build/resources/main where the new version file is created. Somehow this seems to cause a classpath.index file to show up when accessing any of these resources, and since we now have both a resources/test and a resources/main directory being created, we get two classapath.index files

This is a problem because if you try to run ./gradlew install (or whatever) after these are created by a test run, it fails with
```
Execution failed for task ':kafka-client:jar'.
> Entry classpath.index is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/8.1.1/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
```

it seems like maybe we can just tell it how to resolve the multiple strategies, but I just want to make sure I'm not accidentally hurting the caching or build in some way by doing that (assuming it even works). Otherwise we'd have to manually clear these files on every build (which might also hurt the caching? Although if it's just about looking for files then that's pretty fast, so...)

Anyways is anyone familiar with the classpath.index file? Is this a gradle or IntelliJ or plain java thing? I think maybe it has to do with caching rather than any actual build/classpath functionality, but I can't seem to find any documentation on it -- just people complaining it caused issues.

kafka-client/build.gradle.kts
```diff
@@ -16,8 +18,79 @@
...
 
+// Need to add all generated resource directories as outputs of the main sourceSet
+sourceSets {
+    main {
+        output.dir(file("$buildDir/resources"))
+    }
+}
+
+
+// TODO: there has GOT to be a more concise way to do this than listing it out for every task individually...
```
agavra, line 61: I think these all depend on either compileJava or compileTestJava - we should be able to put it just in those two (I don't know whether compileTestJava depends on compileJava or not, I think it probably does)

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
```diff
@@ -252,13 +258,35 @@ public static ResponsiveConfig responsiveConfig(final Map<?, ?> originals) {
    * mid-application, for example during state store init.
    */
   public static ResponsiveConfig loggedConfig(final Map<?, ?> originals) {
+
+    final Properties props = loadVersionPropertiesFromFile();
+    final String version = props.getProperty("git.build.version", "").trim();
```
agavra, line 263: is it possible to create constants here that we can share with the kotlin script?

agavra (approved):
Thanks @ableegoldman! This will be quite helpful

kafka-client/src/main/java/dev/responsive/kafka/config/ResponsiveConfig.java
ableegoldman: After much digging around on this question, I have to conclude that the answer is "no" -- at least, it's not possible to create/share constants in the way you mean or would make sense here. My understanding is that the only way to "share" constants between the gradle/kotlin scripts and our compiled java classes is basically to pass them from gradle to the JVM via environment variables or properties files. That doesn't really help us keep them in sync because we still need to know the format/name of the env/file property -- which is basically exactly what these constants are used for in the first place. ie we'd just be adding another layer of equally fragile strings to keep track of 

There do seem to be some 3rd party plugins that _might_ achieve this, but the whole reason for doing this manually was to avoid those in the first place (not to mention, they probably are just syntactic sugar over basically the same method of passing around constants 🤷‍♀️ )

ableegoldman: All that said, I'm still relatively a complete noob to kotlin for gradle, so I'd be very happy to be proven wrong and learn how to do this. If nothing else this whole patch has been an excellent crash course in kotlin-based gradle build scripts, so I appreciate the question even if I couldn't work out a way to do it that made sense

kafka-client/build.gradle.kts
ableegoldman: Yeah I had that originally and could not for the life of me figure out why it was still complaining -- turns out I had added something to the build file that needed to be explicitly defined for each task, hence this whole mess, but wasn't necessary to begin with. Anyways reduced it down to just the compileJava task

PR #143, opened 2023-09-21
Add `assigned-partitions` consumer metric

ableegoldman:
Could have sworn I added this already but I don't see it, so...

Good to have


ableegoldman:
I'm just going to merge this since it's only adding a single metric

PR #144, opened 2023-09-22
fix and re-enable global stores

agavra:
This PR does three things (let me know if you'd rather me split it up!):

1. It fixes an issue with time-travel where offsets are not committed after restoration but an `unsubscribe` is called, meaning those offsets get re-processed
2. It fixes a potential time-travel issue where restore reprocesses records during an unclean shutdown (essentially the same situation as above, but if a commit isn't called during normal processing) by persisting the last written offset to a metadata table
3. It improves the implementation to use Fact schema instead of the normal schema so we don't push everything into a single partition

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveGlobalConsumer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveGlobalConsumer.java
@@ -90,6 +90,18 @@ public class ResponsiveGlobalConsumer extends DelegatingConsumer<byte[], byte[]>
     );
   }
 
+  @Override
+  public void unsubscribe() {
+    // since this consumer has ENABLE_AUTO_COMMIT_CONFIG set to true,
+    // it is not guaranteed that any commits will happen before calling
+    // unsubscribe - the GlobalStreamThread will call unsubscribe
+    // when it's finished restoring, so we should make sure to commit
+    // any offsets at this point so that when it resumes normal operation
+    // it doesn't experience any time travel
+    commitSync();
+    super.unsubscribe();
+  }
+
   @Deprecated
   public ConsumerRecords<byte[], byte[]> poll(final long timeoutMs) {
     return poll(Duration.ofMillis(timeoutMs));
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraClient.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraClient.java
@@ -53,6 +53,7 @@ public class CassandraClient {
   private final RemoteKeyValueSchema kvSchema;
   private final RemoteKeyValueSchema factSchema;
   private final RemoteWindowedSchema windowedSchema;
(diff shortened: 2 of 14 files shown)


kafka-client/src/main/java/dev/responsive/db/CassandraGlobalKeyValueSchema.java
```diff
@@ -0,0 +1,40 @@
...
+ * strategy instead of time windowed.
+ */
+public class CassandraGlobalKeyValueSchema extends CassandraFactSchema {
+
+  public CassandraGlobalKeyValueSchema(final CassandraClient client) {
+    super(client);
+  }
+
+  @Override
+  protected CompactionStrategy<?> compactionStrategy(final Duration compactionWindow) {
```
agavra, line 37: this is a tiny bit hacky, but I figured it was better than copying a bunch of code. originally I had copied everything, but thought that was worse

happy to hear your thoughts @ableegoldman

kafka-client/src/main/java/dev/responsive/db/CassandraFactSchema.java
```diff
@@ -147,10 +152,12 @@ public MetadataRow metadata(final String table, final int partition) {
         .setInt(PARTITION_KEY.bind(), partition);
     final List<Row> result = client.execute(bound).all();
 
-    if (result.size() != 1) {
+    if (result.size() > 1) {
       throw new IllegalStateException(String.format(
-          "Expected exactly one offset row for %s[%s] but got %d",
+          "Expected at most one offset row for %s[%s] but got %d",
```
agavra, line 157: we don't initialize the offsets in the case of global stores, so it's possible to get no rows back

ableegoldman: Will probably end up doing the same for actual WindowStores soon, so works for me

agavra:
Well clearly this fix didn't help since the test failed on GHA... will keep investigating.

ableegoldman (approved):
Looks good for now, although I'm a bit unhappy with how much the schema definition has gotten tangled up with the compaction strategy -- or how it has separated, depending on how you look at it. I think we should either lean in and define a RemoteSchema as a combination of byte format/store type (eg KV vs Window) plus a compaction strategy, or else fully decouple it. I know that sounds vague, but we can talk about it more later -- this looks ok for now. 

(for example, it's very confusing to say we're going to make global stores a fact schema in one PR while in another we define fact schema as allowing only new writes and restrict overwrites/deletes -- which we definitely should, and do, allow for global stores. Code-wise it might work, but conceptually it's hard to reason about)

agavra:
>(for example, it's very confusing to say we're going to make global stores a fact schema in one PR while in another we define fact schema as allowing only new writes and restrict overwrites/deletes -- which we definitely should, and do, allow for global stores. Code-wise it might work, but conceptually it's hard to reason about)

That's fair feedback, I think I have something in mind about how to make this clearer -- though whatever we do should tie together schemas, compaction and supported operations since they're all interconnected.

I'll merge this for now and address it in the fact schema PR, since that refactors some stuff anyway.

PR #145, opened 2023-09-26
Clean up EndOffsetsPoller logging

ableegoldman:
Noticed a flood of INFO-level logging in the soak, mainly about updating the end offsets which is done every 30s but per-partition. Can definitely be reduced to a single line, and possibly bumped down to DEBUG

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
@@ -236,7 +236,7 @@ public class EndOffsetsPoller {
         log.debug("Updated end offsets to [{}].",
                   endOffsets.entrySet().stream()
                      .map(tp -> "(" + tp.getKey() + ": " + tp.getValue() + ")")
-                     .collect(Collectors.joining(","))
+                     .collect(Collectors.joining(", "))
         );
       }
     }
--- a/kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
@@ -125,7 +125,11 @@ public class EndOffsetsPoller {
     if (poller != null) {
       throw new IllegalStateException("Poller was already initialized");
     }
-    poller = new Poller(() -> factories.createAdminClient(configs), executor, this::getThreadMetrics);
+    poller = new Poller(
+        () -> factories.createAdminClient(configs),
+        executor,
+        this::getThreadMetrics
+    );
   }
 
   private void stopPoller() {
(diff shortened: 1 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -88,10 +89,10 @@ public EndOffsetsPoller(
   }
 
   public synchronized Listener addForThread(final String threadId) {
-    LOG.info("add end offset metrics for thread {}", threadId);
+    LOG.debug("Adding end offset metrics for thread {}", threadId);
```
ableegoldman, line 92: For some context, the Metrics implementation from Kafka that we use has the "adding/removing metric X" logging all the way down to TRACE. Debug seems like a reasonable compromise imo

kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -119,19 +121,19 @@ private synchronized Collection<Listener> getThreadMetrics() {
...
 
   private void initPoller() {
-    LOG.info("init end offsets poller");
+    LOG.info("Initializing end offsets poller");
     if (poller != null) {
-      throw new IllegalStateException("poller already initialized");
+      throw new IllegalStateException("Poller was already initialized");
     }
-    poller = new Poller(factories.createAdminClient(configs), executor, this::getThreadMetrics);
+    poller = new Poller(() -> factories.createAdminClient(configs), executor, this::getThreadMetrics);
```
ableegoldman, line 128: This isn't really necessary, but I feel like having the `Poller` be responsible for creating the admin client helps  make it clear that the `Poller` is also responsible for closing the admin during shutdown. That's the only motivation behind this change

kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -160,6 +162,8 @@ private void stop() {
     }
 
     private void pollEndOffsets() {
+      LOG.info("Polling end offsets");
```
ableegoldman, line 165: This is logged once per poll, before (and after) we update the end offsets

kafka-client/src/main/java/dev/responsive/kafka/clients/EndOffsetsPoller.java
```diff
@@ -217,21 +231,22 @@ public void onPartitionsAssigned(final Collection<TopicPartition> assigned) {
...
-          );
-          return entry.getValue().offset();
-        });
+    private void update(final Map<TopicPartition, ListOffsetsResultInfo> newEndOffsets) {
+      for (final var entry : newEndOffsets.entrySet()) {
+        endOffsets.computeIfPresent(entry.getKey(), (k, v) -> entry.getValue().offset());
+      }
+
+      if (log.isDebugEnabled()) {
+        log.debug("Updated end offsets to [{}].",
```
ableegoldman, line 240: DEBUG feels more appropriate to me, but I could definitely be persuaded to bump this up to INFO if it seems useful -- once every 30s isn't too bad for a single log line, but it could get pretty long for large clusters with a lot of partitions. Anyways I'm fine either way

rodesai (approved):
LGTM, thanks!

PR #146, opened 2023-09-28
set grace_gc_seconds explicitly in the case of TWCS + TTL

agavra:


ableegoldman (approved):
Awesome -- I knew we could take advantage of not really needing additional tombstone retention time in our specific ttl case, although I completely forgot about it haha. Thanks for this

PR #147, opened 2023-09-29
Register StateListener to track KafkaStreams.State metric 

ableegoldman:


.github/workflows/github-pr.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
         with:
           repository: responsivedev/sindri
           path: sindri
-          ref: refs/heads/main
+          ref: refs/heads/SYSTEM-TEST-improve-logging
```
agavra, line 54: @ableegoldman is this intentional?

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStateListener.java
```diff
@@ -0,0 +1,65 @@
...
+      final String streamsClientId
+  ) {
+    this.metrics = metrics;
+    final Map<String, String> tags = new HashMap<>();
+    tags.put("application-id", applicationId);
+    tags.put("client-id", streamsClientId);
+
+    stateMetricName = new MetricName(
+        "state",
+        "stream-application-metrics",
```
agavra, line 35: for other streams metrics it seems we used `responsive.streams` - any reason not to use that group here?

agavra (approved):
LGTM, just one thing about the GHA yaml that's a blocker

.github/workflows/github-pr.yaml
agavra: ah I just saw this is a temporary commit in the history - I thought this was already merged but it isn't

ableegoldman: ah, shoot. didn't mean to commit that -- thanks

ableegoldman: >  I thought this was already merged but it isn't

Yeah sorry I was overconfident at first and said I was going to merge it ASAP since it was pretty late, but I always wait for the tests to pass -- and they did not. This was a remnant of my debugging because I didn't know about the system test logs in DD lol

kafka-client/src/main/java/dev/responsive/kafka/store/ResponsiveStateListener.java
ableegoldman: Well right now at least the only metrics with that seem to be topic metrics that we added -- personally, I would've just put those in the `"stream-topic-metrics"` group with the other built-in topic metrics in Streams

The idea being to keep metrics scoped to the same level/tags in the same group -- but that's just my random personal philosophy, if there's a better way to organize things I'll go along. We can keep all the responsive metrics in the same group, it's just a bit awkward when we have metrics that are really streams metrics...

that's the long answer. Short answer is, anything works

PR #149, opened 2023-09-30
Reduce INFO level logging of flushes by the CommitBuffer

ableegoldman:
Debugging the soak has been great practice for our logs and their usefulness vs sanity ratio. With the various offset pollers reduced in frequency, the CommitBuffer logging around flushes seems to be the most aggressive thing left, definitely too much for INFO

Tried to strike a good balance by bumping the high frequency logging to DEBUG with some new summary logging for us to have something at INFO.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
@@ -55,6 +55,10 @@ import org.slf4j.Logger;
 
 class CommitBuffer<K, S extends RemoteSchema<K>> implements RecordBatchingStateRestoreCallback {
 
+  // Keep INFO level logging to a reasonable frequency
+  private static final long INFO_LOGGING_OFFSET_INTERVAL = 50_000L;
+  private long lastOffsetLogged = 0L;
+
   public static final int MAX_BATCH_SIZE = 1000;
   private final Logger log;
 
@@ -270,11 +274,6 @@ class CommitBuffer<K, S extends RemoteSchema<K>> implements RecordBatchingStateR
     return remoteSchema.metadata(tableName, basePartition).offset;
   }
 
-  // Visible For Testing
-  int size() {
-    return buffer.getReader().size();
-  }
-
   private long totalBytesBuffered() {
     return buffer.getBytes();
   }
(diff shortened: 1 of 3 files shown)


ableegoldman:
@agavra take a look when you have a minute and lmk if there's anything you feel is missing -- I definitely do not want to err on the side of too little logging, so keep me in check please!

kafka-client/src/main/java/dev/responsive/internal/stores/CommitBuffer.java
```diff
@@ -274,44 +273,50 @@ private boolean shouldFlush() {
...
+      log.debug("Time since last flush {} not over trigger {}",
           Duration.between(lastFlush, now),
           now
       );
     }
-    return recordsTrigger || bytesTrigger || timeTrigger;
+    final boolean shouldFlush = recordsTrigger || bytesTrigger || timeTrigger;
+
+    if (shouldFlush) {
+      log.info("Triggering flush due to recordsTrigger={}, bytesTrigger={}, timeTrigger={}",
```
agavra, line 316: we might as well add the number of records we're about to flush here (and the current topic-partition/virtual-partition/offset) - I think all of those are helpful when debugging.

kafka-client/src/main/java/dev/responsive/internal/stores/CommitBuffer.java
```diff
@@ -385,6 +390,11 @@ private RemoteWriteResult flush(final long offset, final int batchSize) {
     );
     buffer.clear();
 
+    if (offset - lastOffsetLogged > INFO_LOGGING_OFFSET_INTERVAL) {
```
agavra, line 393: if we incorporate my comment above then we can just remove this and the offset interval altogether, right? (and just log on every flush, which we seem to be doing anyway - yes that'll be a lot when there's 4096 virtual partitions, but I did find that to be really helpful debugging stuff)

agavra (approved):

kafka-client/src/main/java/dev/responsive/internal/stores/CommitBuffer.java
ableegoldman: Fair enough -- was on the fence about whether or not to leave even one INFO level log per flush and added this back at the last minute, but if there's anything logged per flush at INFO it should at least have all the useful information.

I'll remove this one entirely and then log the summary without the offset interval logic. Also just to clarify up front in case you were going to ask, I'm intentionally leaving out the kafka topic/partition because that's static and is included automatically in the logger's prefix.

kafka-client/src/main/java/dev/responsive/internal/stores/CommitBuffer.java
ableegoldman: Fair enough, we can always reduce it further or implement the interval/summary logging if it's really overwhelming. Let's keep an eye on it for now. Maybe I'm just being paranoid after one too many on-call escalations that came down to performance degradation from excessive logging

(the other one I'm a bit concerned about is the INFO-level logging after each successful changelog truncation, though I left that alone for now as well -- of course logging should be trivial compared to the fairly heavyweight admin  requests to delete records, so...

PR #150, opened 2023-09-30
Allow all topology optimizations and disable truncation for source-topic changelogs

ableegoldman:
We should now be able to handle all topology optimizations, specifically reusing source topics as changelogs which is currently a restricted config.

We can simply remove all validation of the `topology.optimization` config, though we do want to be guard against users accidentally deleting source topic data and therefore should make sure to disable changelog truncation for any source tables with this source-changelog optimization


agavra (approved):
Thanks!

PR #152, opened 2023-10-01
Make changelog truncation happen asynchronously by CommitBuffer

ableegoldman:
Planning to step this up in the soak and have been pretty concerned about doing all these blocking requests on the relatively-hot-path of flushing. Since we don't depend on changelog truncation in any way and it's completely idempotent, retriable, etc we can just kick off a new deleteRecords request after each flush without waiting for it to complete. When the next flush comes around then we can check on the previous delete request's status and kick off a new request if it was succcessful, or else retry with the updated low-watermark offset to delete from

I hope this will help enough to call the feature production ready, but we should still be aware of admin calls and connections which have caused many issues with Streams in the past. If we still run into issues we could go a step further by consolidating all the delete requests across all the truncated changelog partitions (on the node) into a single request. Though I can't imagine that will be too critical until we start running apps with huge numbers of partitions, and we'll probably find other bottlenecks in that case far before we hit one here

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
@@ -16,6 +16,8 @@
 
 package dev.responsive.kafka.internal.stores;
 
+import static org.apache.kafka.clients.admin.RecordsToDelete.beforeOffset;
+
 import dev.responsive.kafka.api.config.ResponsiveConfig;
 import dev.responsive.kafka.internal.db.CassandraClient;
 import dev.responsive.kafka.internal.db.KeySpec;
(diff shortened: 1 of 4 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -408,23 +412,51 @@ private RemoteWriteResult drain(final Collection<RemoteWriter<K>> writers) {
...
+        deleteRecordsResult = null;
+
+      } else {
+        log.debug("Still waiting on previous changelog truncation attempt to complete");
       }
     }
+
+    if (truncateChangelog && !isNotDeleteEnabled && deleteRecordsResult == null) {
+      log.debug("Issuing new delete request to truncate {} up to offset {}", changelog, offset);
+      deleteRecordsResult = admin.deleteRecords(Map.of(changelog, beforeOffset(offset)));
```
agavra, line 458: (feel free to ignore) if you're keen to explore, I think this would be more idiomatic if you change this to use `Future#whenComplete` and pass in the logging as a method reference

then you can also initialize it to `KafkaFuture.completedFuture(null)` and avoid nulls by:
```
if (truncateChangelog && isNotDeleteEnabled && deleteRecordResult.isDone()) {
   deleteRecordResult = admin.deleteRecords().lowWatemarks().get(changelog);
}
```

agavra (approved):
LGTM, just a suggestion if you want to explore it and see if it makes the code cleaner

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
ableegoldman: ah yeah that's much nicer, good tip

PR #153, opened 2023-10-02
Consolidate common store logic into AbstractResponsiveStore

ableegoldman:
There's a lot of shared and increasingly-complex logic in the three StateStore implementations. This would be a good time to extract it all into an abstract class that can be extended by the ResponsiveGlobalStore, ResponsivePartitionedStore, and ResponsiveWindowStore implementations (as well as ResponsiveSessionStore one day)

There aren't any real logical changes here, but it's an important refactoring that will help keep things organized and consistent across the state stores. Specifically, I'd like to have this merged and ready to take advantage of for both the standby-task fix and the upcoming WindowStore implementation, as well as some smaller ongoing side efforts like the restoration metrics

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
@@ -202,11 +202,11 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
         responsiveConfig,
         responsiveKafkaClientSupplier,
         time,
-        storeRegistry,
         metrics,
         new SharedClients(
             cassandraClient,
-            responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals())
+            responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals()),
+            storeRegistry
         )
     );
   }
@@ -216,7 +216,6 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
       final ResponsiveConfig responsiveConfig,
       final ResponsiveKafkaClientSupplier clientSupplier,
       final Time time,
-      final ResponsiveStoreRegistry storeRegistry,
       final Metrics metrics,
       final SharedClients sharedClients
   ) {
@@ -225,7 +224,6 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
         propsWithOverrides(
             responsiveConfig.originals(),
             sharedClients,
-            storeRegistry,
             topology.describe()),
         clientSupplier,
         time
@@ -272,7 +270,6 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
   private static Properties propsWithOverrides(
       final Map<?, ?> configs,
       final SharedClients sharedClients,
-      final ResponsiveStoreRegistry storeRegistry,
       final TopologyDescription topologyDescription
   ) {
     final Properties propsWithOverrides = new Properties();
@@ -280,7 +277,7 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
     propsWithOverrides.putAll(new InternalConfigs.Builder()
             .withCassandraClient(sharedClients.cassandraClient)
             .withKafkaAdmin(sharedClients.admin)
-            .withStoreRegistry(storeRegistry)
+            .withStoreRegistry(sharedClients.storeRegistry)
             .withTopologyDescription(topologyDescription)
             .build());
 
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
@@ -16,7 +16,7 @@
 
 package dev.responsive.kafka.api.stores;
 
-import dev.responsive.kafka.internal.stores.ResponsiveStore;
+import dev.responsive.kafka.internal.stores.ResponsiveKeyValueStoreWrapper;
 import java.util.Locale;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.state.KeyValueBytesStoreSupplier;
@@ -37,7 +37,7 @@ public class ResponsiveKeyValueBytesStoreSupplier implements KeyValueBytesStoreS
 
   @Override
   public KeyValueStore<Bytes, byte[]> get() {
-    return new ResponsiveStore(params);
+    return new ResponsiveKeyValueStoreWrapper(params);
   }
 
   @Override
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactSchemaWriter.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactSchemaWriter.java
@@ -29,7 +29,7 @@ public class FactSchemaWriter<K> implements RemoteWriter<K> {
   private final CassandraClient client;
   private final RemoteSchema<K> schema;
   private final String tableName;
-  private final int subpartition;
+  private final int tablePartition;
 
   private final List<Statement<?>> statements;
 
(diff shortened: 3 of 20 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -202,11 +202,11 @@ private ResponsiveKafkaStreams(
         responsiveConfig,
         responsiveKafkaClientSupplier,
         time,
-        storeRegistry,
         metrics,
         new SharedClients(
             cassandraClient,
-            responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals())
+            responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals()),
+            storeRegistry
```
ableegoldman, line 209: Just moved the store registry to the SharedClients container class, since it was already being passed around via the configs and is needed/accessed from the same places that were also accessing the other client(s)

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
```diff
@@ -37,7 +37,7 @@ public String name() {
 
   @Override
   public KeyValueStore<Bytes, byte[]> get() {
-    return new ResponsiveStore(params);
+    return new ResponsiveKeyValueStoreWrapper(params);
```
ableegoldman, line 40: No change, just a renaming of `ResponsiveStore` to emphasize that it (a) is a KV store, and (b) is itself just a wrapper for the two KV sub-types, namely the global vs partitioned (regular) store

kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactSchemaWriter.java
```diff
@@ -28,33 +28,33 @@ public class FactSchemaWriter<K> implements RemoteWriter<K> {
 
   private final CassandraClient client;
   private final RemoteSchema<K> schema;
-  private final String name;
-  private final int partition;
+  private final String tableName;
+  private final int subpartition;
```
ableegoldman, line 32: Sorry for all the renaming here, I know it makes PR reviews more annoying but I kept having to jump around to understand when something was a kafka partition vs a subpartition. Hopefully this helps others follow the code the way it helps me

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/AbstractResponsiveStore.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Collection;
+import java.util.function.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.streams.processor.StateStore;
+import org.apache.kafka.streams.processor.StateStoreContext;
+import org.apache.kafka.streams.processor.internals.InternalProcessorContext;
+import org.apache.kafka.streams.query.Position;
+
+public abstract class AbstractResponsiveStore implements StateStore {
```
ableegoldman, line 32: For reviewers: the introduction of this abstract class is the main change here. Tried to pull out all/only the common logic between the stores. There might be some further adjustments to how things are broken up in `init` when I rebase the standby-task fix on top of this, but I wanted to make sure this PR stands on its own in case that fix takes any longer

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveGlobalStore.java
```diff
@@ -87,60 +74,61 @@ public void init(final ProcessorContext context, final StateStore root) {
...
-  @Override
-  public boolean persistent() {
-    return false;
-  }
-
   @Override
   public void put(final Bytes key, final byte[] value) {
-    client.execute(schema.insert(name.cassandraName(), partition, key, value, context.timestamp()));
-    StoreQueryUtils.updatePosition(position, context);
+    if (value == null) {
```
agavra, line 120: nit: thanks for the fix! let's do this in a separate PR so this one is just a refactor

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/AbstractResponsiveStore.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private boolean open;
+
+  public AbstractResponsiveStore(final TableName name) {
+    this.name = name;
+    this.position = Position.emptyPosition();
+  }
+
+  @SuppressWarnings("rawtypes")
+  protected void init(
```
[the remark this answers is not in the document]
agavra: another trick, by the way, to make sure that variables (especially non-final ones) remain under control of the super class is to pass them in to the methods you expect the subclass to override. 

Here's a quick example:

```java
class Base {
  private int foo;

  final void init() {
    foo = 3;
  }

  final int magic() {
    return doMagic(foo);
  }

  protected abstract int doMagic(final int foo);
}

class Sub {

  @Override
  protected int doMagic(final int foo) {
    return foo * 2;
  }

}
```

Generally what's nice here is that you can read each class totally independently of one another - here `doMagic` is a very clear formula, and in the super class `magic()` is clear that it delegates to some implementation, but they're both actually entirely independent.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactSchemaWriter.java
ableegoldman: I was thinking of it as a subpartition that just happened to be 1:1 with the kafka partitions, but I guess that's confusing too -- the goal was to distinguish between the kafka partition vs the table partition, regardless of the mapping between them

Idk how I didn't think of `tablePartition` vs `kafkaPartition` though, I like that a lot. Great suggestion!

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveGlobalStore.java
```diff
@@ -87,60 +74,61 @@ public void init(final ProcessorContext context, final StateStore root) {
...
     try {
       log.info("Initializing global state store {}", name);
-      this.context = (GlobalProcessorContextImpl) context;
+
+      final SharedClients sharedClients = loadSharedClients(storeContext.appConfigs());
+
       // this is bad, but the assumption is that global tables are small
-      // and can fit in a single partition - all writers will write using
-      // the same partition key. we should consider using ALLOW FILTERING
+      // and can fit in a single topicPartition - all writers will write using
```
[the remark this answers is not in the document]
ableegoldman: Hm, I don't remember changing this on purpose -- might have accidentally been swept up in the variable name change from partition -> topicPartition

I'll update it to say "table partitions" for now but let you take care of it from there

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveGlobalStore.java
ableegoldman: fair enough -- https://github.com/responsivedev/responsive-pub/pull/154

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/AbstractResponsiveStore.java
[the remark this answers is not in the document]
ableegoldman: 1. Good point about init/open, you're totally right I have the parent-child relationship completely backwards right now. Thanks for the feedback and suggestions, let me try again with a more principled approach
2. Regarding the `position` thing, it's certainly true I left the ownership/responsibility unclear, although that was somewhat intentional -- there's no contract that defines where/when it should be updated because I'm pretty sure we're not actually updating it everywhere it should be in the current implementation, and we'll need to do a thorough pass if/when we do want to enable IQ (see the TODO). But I'll try to get it into a better place for now

ableegoldman:
Closing this for now since it's pretty out of date and I think we've addressed a good chunk of the issues here already

PR #154, opened 2023-10-03
Treat null-value puts as deletes in ResponsiveGlobalStore

ableegoldman:
Extracting this fix to its own PR as suggested here https://github.com/responsivedev/responsive-pub/pull/153#discussion_r1343070475


agavra (approved):
Thanks!

PR #156, opened 2023-10-05
Consolidate metrics and group by scope

ableegoldman:
The primary goal here is to consolidate all of the metrics added by us and then group them according to their scope, so that we can attach a set of common base tags to all metrics with additional tags added based  on the specific context

To aid with reviewing, you should start with the ResponsiveMetrics class which is the main change. I also scooped up the versioning info that was randomly logged during config validation, and moved that into a new ClientVersionMetadata class which is loaded from the filesystem and passed in to provide version context tags to all metrics.

To help track tags for a given scope, I added a `MetricGroup` interface with three starter implementations: ApplicationMetrics, TopicMetrics, and StoreMetrics. The first two are already in use in this PR as I migrated the existing metrics which should help demonstrate the approach.

The StoreMetrics group will be used by the restoration metrics, though I backed those out of this PR to avoid adding too much at once. Will open a followup PR with those on top of this one shortly


kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -259,35 +253,13 @@ public static ResponsiveConfig responsiveConfig(final Map<?, ?> originals) {
    * mid-application, for example during state store init.
    */
   public static ResponsiveConfig loggedConfig(final Map<?, ?> originals) {
-
-    final Properties props = loadVersionPropertiesFromFile();
-    final String version = props.getProperty("git.build.version", "").trim();
-    final String commitId = props.getProperty("git.commit.id", "").trim();
-
-    if (!version.isEmpty() && !commitId.isEmpty()) {
-      LOG.info("Responsive Client version: {}", version);
```
ableegoldman, line 268: Basically moved the logging into ResponsiveKafkaStreams, and the actual version property file stuff into ClientVersionMetadata

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -58,7 +59,7 @@ public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier
   private final KafkaClientSupplier wrapped;
   private final ResponsiveStoreRegistry storeRegistry;
   private final Factories factories;
-  private final Metrics metrics;
+  private final ResponsiveMetrics metrics;
```
ableegoldman, line 62: no real changes in this file, just going from Metrics -> ResponsiveMetrics

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/MetricPublishingCommitListener.java
```diff
@@ -45,9 +46,8 @@
  */
 public class MetricPublishingCommitListener implements Listener, Closeable {
   private static final Logger LOG = LoggerFactory.getLogger(MetricPublishingCommitListener.class);
-  private final Metrics metrics;
+  private final ResponsiveMetrics metrics;
   private final String threadId;
-  private final String consumerGroup;
```
ableegoldman, line 50: FYI I didn't remove this tag just moved it to ResponsiveMetrics so now all metrics will be tagged with the group

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  public void initialize(
+      final String applicationId,
+      final String streamsClientId,
+      final ClientVersionMetadata versionMetadata,
+      final Map<String, ?> userTags
+  ) {
+    // this is technically redundant with the application id, but sometimes it's useful to have both
+    baseTags.put("consumer-group", applicationId);
+    baseTags.put("application-id", applicationId);
```
rodesai, line 64: can we prefix this with "streams". We have application/applicationId tags that we use in other places to refer to the responsive application ID

ableegoldman: ah yeah good call

ableegoldman:
Copied all the otel file changes over to the soak and redeployed it based on this change to verify that all the Responsive metrics are still healthy. Here's a screenshot of the committed offset lag which depends on both of our custom topic-level metrics, ie the ones in `TopicMetrics`:

Screenshot 2023-10-05 at 8 54 17 PM

Similarly, the state metric from `ApplicationMetrics` is present (and I updated the dash to use some of the new tags): 
Screenshot 2023-10-05 at 9 10 56 PM

rodesai (commented):
Skimmed through the code. Overall the structure makes sense. One bit of annoying feedback inline.

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ApplicationMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+  public static final String STREAMS_STATE = "streams-state";
+  public static final String STREAMS_STATE_DESCRIPTION = "The current KafkaStreams.State expressed as its ordinal value";
+
+  public static final String NUM_RESTORING = "num-restoring-changelogs";
+  public static final String NUM_RESTORING_DESCRIPTION = "The number of changelog topics currently being restored from";
+
+  public static final String NUM_INTERRUPTED = "num-interrupted-changelogs";
+  public static final String NUM_INTERRUPTED_DESCRIPTION = "The total number of changelog partitions that began restoring but did not complete";
+
+  private final Map<String, String> tags = new HashMap<>();
```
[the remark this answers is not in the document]
ableegoldman: Wow I had no idea -- thanks for the heads up. Here we go: https://github.com/responsivedev/responsive-pub/pull/160

PR #158, opened 2023-10-05
refactor the interaction with remote tables

agavra:
Ok touches lots of lines! It's purely a refactor though and actually quite simple and has one intent: make it easier later to have things per-table instead of globally for all tables of that type.

I tried really hard to avoid sneaking in any functionality changes here. There are only two changes:

1. Renamed `RemoteSchema` to `RemoteTable`, now a `RemoteTable` is tied to a specific table so there's only one `Map<String(Table Name), RemoteTable>` instead of one for each prepared statement. This did require some synchronization changes, but I think it should be safe. Would be nice for a reviewer to reason about this. The mapping part of it is put into a `TableFactory` class, which is totally generic to the underlying table type.
2. Abstracted away the shared configurations between tables (e.g. TTL, Compaction Strategy, etc...) into a class called `CassandraTableSpec` which has a bunch of implementations which makes it easy to chain them together depending on how we build up the configuration. This is where we're going to put things like what operations are supported and what are not, and depending on the spec we can add more unsupported operations in a fluent fashion.

The rest is fallout from those changes. After this I'll have some follow-up refactors that touch fewer lines.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/spec/TimeWindowedCompactionTableSpec.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Override
+  public EnumSet<TableOperations> restrictedOperations() {
+    // TODO(agavra): we should restrict PUT, PUT_ALL and DELETE in this spec
+    return super.restrictedOperations();
+  }
+
+  @Override
+  public CreateTableWithOptions applyOptions(final CreateTableWithOptions base) {
+    if (!(delegate() instanceof TtlTableSpec)) {
```
agavra, line 41: this needs DFS

ableegoldman (approved):
LGTM, definitely a huge improvement and a better abstraction.

Just to summarize what we chatted about so I don't forget: the only piece of this that feels a bit off is the way the Specs are organized. A few concrete pain points we discussed:
1. Composing them like this means having to wrap/unwrap all the layers to answer a single question about whether the store has a given property, eg `ttl`. I'm a bit worried about the chance of accidentally introducing bugs in the future because we forget to go through all the layers (might be excessively paranoid, just a concern of mine)
2. It sounds like the spec composition is sensitive to the order in which specs are applied, such as ttl and twcs. This would be quite fragile and difficult to maintain, and could get out of hand if/when new specs are added with new relationships forming between them
3. The TTD seems to be giving us some trouble and breaks the abstraction in at least one place due to the ttl spec. To be fair even with the old scheme, the TTD was a bit awkward to fit into our client. But we should probably take care not to let it get too messy

Think that covers it -- but as I said, go ahead and merge this to unblock the next PR. We can circle back with fresh eyes when the full refactor is complete

agavra:
Thanks for summarizing and reviewing @ableegoldman!

PR #159, opened 2023-10-06
split out operations from state store implementation for better code reuse

agavra:
Unfortunately git is deciding that they're all new/deleted files when I actually just renamed and moved some code around. Here's a recap:

1. Moved most of the logic from `ResponsivePartitionedStore` and `ResponsiveGlobalStore` into implementations of a new interface `KeyValueOperations` (ugh naming is hard, happy for someone to suggest a better name)
2. Renamed `ResponsiveStore` to `ResponsiveKeyValueStore` - this way we have only two State Store implementations: `ResponsiveKeyValueStore` and `ResponsiveWindowStore`. This maps one state store per semantic class. If we change details for improved implementation, we want to introduce new `KeyValueOperations` implementations

I deleted the test that was basically testing that an if branch did what it said it was doing (`ResponsiveStoreTest`), the global integration test covers it.


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
```diff
@@ -0,0 +1,211 @@
...
+import org.apache.kafka.streams.processor.StateStore;
+import org.apache.kafka.streams.processor.StateStoreContext;
+import org.apache.kafka.streams.processor.internals.Task.TaskType;
+import org.apache.kafka.streams.query.Position;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.internals.StoreQueryUtils;
+import org.slf4j.Logger;
+
+public class ResponsiveKeyValueStore implements KeyValueStore<Bytes, byte[]> {
```
ableegoldman, line 42: I'm very happy with this, ie having a single "ResponsiveXXXStore" per actual Streams store type 👍

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
```diff
@@ -0,0 +1,211 @@
...
+  private CassandraTableSpec spec;
+  private StateStoreContext context;
+
+  public ResponsiveKeyValueStore(final ResponsiveKeyValueParams params) {
+    this.params = params;
+    this.name = params.name();
+    this.position = Position.emptyPosition();
+
+    log = new LogContext(
+        String.format("store [%s] ", name.kafkaName())
```
ableegoldman, line 62: nit: probably worth including the spec and/or operations type/name in the log prefix

ableegoldman (approved):
Setting aside the naming here, looks great! One small suggestion but otherwise feel free to merge and get on with the testing

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
agavra: since we don't know this in the constructor I made the logger non-final, and the re-initialized it when `init` is called. I believe there's no logs yet pre-`init`, but this could lead to weird behavior down the road

PR #160, opened 2023-10-06
Add OrderedTagsSupplier to maintain metrics tag ordering 

ableegoldman:
Minor followup to https://github.com/responsivedev/responsive-pub/pull/156

See https://github.com/responsivedev/responsive-pub/pull/156#discussion_r1348442860

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OrderedTagsSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OrderedTagsSupplier.java
@@ -16,6 +16,7 @@
 
 package dev.responsive.kafka.internal.metrics;
 
+import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -55,8 +56,9 @@ public class OrderedTagsSupplier {
   private final String consumerGroup;
   private final String streamsApplicationId;
   private final String streamsClientId;
-  private final List<Entry<String, ?>> orderedUserTags;
+  private final List<Entry<String, String>> orderedUserTags;
 
+  @SuppressWarnings("unchecked")
   public OrderedTagsSupplier(
       final String responsiveClientVersion,
       final String responsiveClientCommitId,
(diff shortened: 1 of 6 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OrderedTagsSupplier.java
```diff
@@ -0,0 +1,127 @@
...
+    this.responsiveClientCommitId = responsiveClientCommitId;
+    this.streamsClientVersion = streamsClientVersion;
+    this.streamsClientCommitId = streamsClientCommitId;
+    this.consumerGroup = consumerGroup;
+    this.streamsApplicationId = streamsApplicationId;
+    this.streamsClientId = streamsClientId;
+    this.orderedUserTags = userTags.entrySet().stream().sorted().collect(Collectors.toList());
+  }
+
+  public LinkedHashMap<String, String> applicationGroupTags() {
```
agavra, line 80: we may want to consider an `EnumMap` here and have all the tags be enums with string values - `EnumMaps` are always maintained in natural order so it would be a global ordering that's all maintained in one place (the enum).

It also makes testing it pretty easy because each enum can be a name + an int for its ordinal (e.g. `FOO("foo", 1)` and you just test that `foo.ordinal() == foo.order()`

agavra (approved):
I had no idea things were order dependent! One suggestion inline to make it easier to enforce

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OrderedTagsSupplier.java
ableegoldman: Oh nice, I hadn't heard of EnumMap before

Great suggestion, but if you don't mind, I think I'll merge this for now so I can focus on the other release tasks -- I'll follow up on this next week

agavra: 👍 sounds good

PR #161, opened 2023-10-06
Use processing guarantee-specific client exceptions when fenced during CommitBuffer#flush

ableegoldman:
Throw the appropriate exception specific to the Kafka client used to commit with a given processing guarantee.

For ALOS, we have to throw a commitFailedException.
For EOS, we should throw a ProducerFencedException

Note: for eos there are technically three different exceptions that Streams will catch from the StreamsProducer's  commit and rethrow as a TaskMigratedException: ProducerFencedException, InvalidProducerEpochException, and CommitFailedException. Until we can clean up the exception handling between Streams and the embedded clients, I believe ProducerFenced is the least brittle and handled across the most of the existing versions of Kafka Streams

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
@@ -177,10 +177,6 @@ public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier 
     );
   }
 
-  public boolean eosEnabled() {
-    return eos;
-  }
-
   /**
    * @param clientId the producer client id
    * @return the extracted StreamThread id, of the form "StreamThread-n"
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
@@ -177,6 +177,10 @@ public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier 
     );
   }
 
+  public boolean eosEnabled() {
+    return eos;
+  }
+
   /**
    * @param clientId the producer client id
    * @return the extracted StreamThread id, of the form "StreamThread-n"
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -494,7 +501,11 @@ private void throwFencedException(final RemoteWriteResult result, final long con
         stored.offset
     );
     log.warn(msg);
-    throw new TaskMigratedException(logPrefix + msg);
+    if (eosEnabled) {
```
rodesai, line 504: I feel like it would be better to push this logic out as far to the periphery as possible and not couple the details of streams exception handling with the implementation of this class. What I mean is rather than storing `eosEnabled` as a field, we could just have something like `Supplier<RuntimeException> fencedExceptionFactory` and just call that here to get the exception.

I'm having trouble articulating why I like that better. It just feels off to have this detail of streams exception handling leaking all the way into this class.

Also, lets definitely leave a comment explaining why we need these exception types.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -177,6 +177,10 @@ public Consumer<byte[], byte[]> getGlobalConsumer(final Map<String, Object> conf
     );
   }
 
+  public boolean eosEnabled() {
```
rodesai, line 180: I don't think this is used anywhere?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
ableegoldman: That's fair -- moved the logic into a new ExceptionSupplier class, lmk if this makes more sense to you now

Tbh this might be useful for other kinds of exceptions as well, or at least somewhere to consolidate some of the error handling. For now it only supplies fenced exceptions but it could theoretically handle some of the other kinds of errors as our handling gets more sophisticated

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/ExceptionSupplier.java
```diff
@@ -0,0 +1,36 @@
+package dev.responsive.kafka.internal.utils;
+
+import dev.responsive.kafka.internal.config.ResponsiveStreamsConfig;
+import java.util.Map;
+import org.apache.kafka.clients.consumer.CommitFailedException;
+import org.apache.kafka.common.errors.ProducerFencedException;
+import org.apache.kafka.streams.StreamsConfig;
+
+public class ExceptionSupplier {
```
rodesai, line 9: nit: rename to `StoreFencedExceptionSupplier` or something more specific

rodesai: ah nvm I saw you've named the function name to be more specific. So the idea is we'd have other types of exceptions supplied here as well? That's fine by me.

PR #164, opened 2023-10-11
add support for different backends, starting with MongoDB

agavra:
This PR is the minimal set of changes I needed to make in order to comfortably support different backends. There are still some places (and I will point them out with inline comments on this PR) where it's awkward and doesn't totally fit our existing. Here's the general review guide:

1. I refactored the `ResponsiveKafkaStreams` chain of constructors into an internal builder to make it easier to modify
2. Added `BackendStorage` enum that signals which backend to use
3. Moved most C* specific things into the corresponding `RemoteTable` classes
4. Made the `ResponsiveExtension` for spinning up test containers smart about whether it should spin up a C* or Mongo container. You can see `MinimalIntegraionTest` as an example of the new usage.

Unfortunately this will mean that we need to package both Mongo/Scylla clients into our JAR. That's not ideal, but hopefully soon we can move all of this into a proxy.

**NOTE:** I haven't spent any time thinking about the MongoDB schema or optimizing it. All I did was the minimal amount of work to get some of the tests to pass (e.g. the restore integration test), I'll spend some time reading up on that later.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
@@ -16,33 +16,41 @@
 
 package dev.responsive.kafka.api;
 
+import static dev.responsive.kafka.api.config.ResponsiveConfig.CLIENT_ID_CONFIG;
+import static dev.responsive.kafka.api.config.ResponsiveConfig.CLIENT_SECRET_CONFIG;
+import static dev.responsive.kafka.api.config.ResponsiveConfig.STORAGE_BACKEND_TYPE_CONFIG;
+import static dev.responsive.kafka.api.config.ResponsiveConfig.STORAGE_HOSTNAME_CONFIG;
 import static dev.responsive.kafka.api.config.ResponsiveConfig.TASK_ASSIGNOR_CLASS_OVERRIDE;
 import static dev.responsive.kafka.internal.metrics.ResponsiveMetrics.RESPONSIVE_METRICS_NAMESPACE;
 
-import com.datastax.oss.driver.api.core.CqlSession;
 import dev.responsive.kafka.api.config.ResponsiveConfig;
+import dev.responsive.kafka.api.config.StorageBackend;
 import dev.responsive.kafka.internal.clients.ResponsiveKafkaClientSupplier;
 import dev.responsive.kafka.internal.config.InternalConfigs;
 import dev.responsive.kafka.internal.config.ResponsiveStreamsConfig;
-import dev.responsive.kafka.internal.db.CassandraClient;
 import dev.responsive.kafka.internal.db.CassandraClientFactory;
 import dev.responsive.kafka.internal.db.DefaultCassandraClientFactory;
+import dev.responsive.kafka.internal.db.mongo.ResponsiveMongoClient;
 import dev.responsive.kafka.internal.metrics.ClientVersionMetadata;
 import dev.responsive.kafka.internal.metrics.ResponsiveMetrics;
 import dev.responsive.kafka.internal.metrics.ResponsiveStateListener;
 import dev.responsive.kafka.internal.stores.ResponsiveRestoreListener;
 import dev.responsive.kafka.internal.stores.ResponsiveStoreRegistry;
+import dev.responsive.kafka.internal.utils.SessionUtil;
 import dev.responsive.kafka.internal.utils.SharedClients;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.config.types.Password;
 import org.apache.kafka.common.metrics.JmxReporter;
 import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
@@ -155,103 +163,23 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
       final Time time
   ) {
     this(
-        topology,
-        ResponsiveConfig.loggedConfig(configs), // this should be the only place this is logged!
-        ResponsiveStreamsConfig.streamsConfig(configs),
-        clientSupplier,
-        time,
-        new DefaultCassandraClientFactory()
+        new Params(topology, configs)
+            .withClientSupplier(clientSupplier)
+            .withTime(time)
+            .build()
     );
   }
 
-  protected ResponsiveKafkaStreams(
-      final Topology topology,
-      final ResponsiveConfig responsiveConfig,
-      final StreamsConfig baseStreamsConfig,
-      final KafkaClientSupplier clientSupplier,
-      final Time time,
-      final CassandraClientFactory clientFactory
-  ) {
-    this(
-        topology,
-        responsiveConfig,
-        baseStreamsConfig,
-        clientSupplier,
-        time,
-        new ResponsiveStoreRegistry(),
-        createMetrics(baseStreamsConfig),
-        clientFactory,
-        clientFactory.createCqlSession(responsiveConfig)
-    );
-  }
-
-  private ResponsiveKafkaStreams(
-      final Topology topology,
-      final ResponsiveConfig responsiveConfig,
-      final StreamsConfig baseStreamsConfig,
-      final KafkaClientSupplier clientSupplier,
-      final Time time,
-      final ResponsiveStoreRegistry storeRegistry,
-      final ResponsiveMetrics metrics,
-      final CassandraClientFactory clientFactory,
-      final CqlSession session
-  ) {
-    this(
-        topology,
-        responsiveConfig,
-        new ResponsiveKafkaClientSupplier(
-            clientSupplier,
-            baseStreamsConfig,
-            storeRegistry,
-            metrics),
-        time,
-        storeRegistry,
-        metrics,
-        clientFactory.createCassandraClient(session, responsiveConfig)
-    );
-  }
-
-  private ResponsiveKafkaStreams(
-      final Topology topology,
-      final ResponsiveConfig responsiveConfig,
-      final ResponsiveKafkaClientSupplier responsiveKafkaClientSupplier,
-      final Time time,
-      final ResponsiveStoreRegistry storeRegistry,
-      final ResponsiveMetrics metrics,
-      final CassandraClient cassandraClient
-  ) {
-    this(
-        topology,
-        responsiveConfig,
-        responsiveKafkaClientSupplier,
-        time,
-        storeRegistry,
-        metrics,
-        new SharedClients(
-            cassandraClient,
-            responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals())
-        )
-    );
-  }
-
-  private ResponsiveKafkaStreams(
-      final Topology topology,
-      final ResponsiveConfig responsiveConfig,
-      final ResponsiveKafkaClientSupplier clientSupplier,
-      final Time time,
-      final ResponsiveStoreRegistry storeRegistry,
-      final ResponsiveMetrics responsiveMetrics,
-      final SharedClients sharedClients
-  ) {
+  protected ResponsiveKafkaStreams(final Params params) {
     super(
-        topology,
+        params.topology,
         propsWithOverrides(
-            responsiveConfig.originals(),
-            sharedClients,
-            storeRegistry,
-            topology.describe()),
-        clientSupplier,
-        time
+            params.responsiveConfig.originals(),
+            params.sharedClients,
+            params.storeRegistry,
+            params.topology.describe()),
+        params.responsiveKafkaClientSupplier,
+        params.time
     );
 
     try {
@@ -260,8 +188,8 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
       throw new StreamsException("Configuration error, please check your properties");
     }
 
-    this.responsiveMetrics = responsiveMetrics;
-    this.sharedClients = sharedClients;
+    this.responsiveMetrics = params.metrics;
+    this.sharedClients = params.sharedClients;
 
     final ClientVersionMetadata versionMetadata = ClientVersionMetadata.loadVersionMetadata();
     // Only log the version metadata for Responsive since Kafka Streams will log its own
@@ -313,11 +241,22 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
       final ResponsiveStoreRegistry storeRegistry,
       final TopologyDescription topologyDescription
   ) {
+    final InternalConfigs.Builder builder = new InternalConfigs.Builder();
+    switch (sharedClients.storageBackend()) {
+      case CASSANDRA:
+        builder.withCassandraClient(sharedClients.cassandraClient());
+        break;
+      case MONGO_DB:
+        builder.withMongoDbClient(sharedClients.mongoClient());
+        break;
+      default:
+        throw new IllegalStateException("Unexpected value: " + sharedClients.storageBackend());
+    }
+
     final Properties propsWithOverrides = new Properties();
     propsWithOverrides.putAll(configs);
-    propsWithOverrides.putAll(new InternalConfigs.Builder()
-            .withCassandraClient(sharedClients.cassandraClient)
-            .withKafkaAdmin(sharedClients.admin)
+    propsWithOverrides.putAll(builder
+            .withKafkaAdmin(sharedClients.admin())
             .withStoreRegistry(storeRegistry)
             .withTopologyDescription(topologyDescription)
             .build());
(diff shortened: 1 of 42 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -0,0 +1,156 @@
...
+      final int partitionKey,
+      final Bytes key,
+      final byte[] value,
+      final long epochMillis
+  ) {
+    collection.updateOne(
+        Filters.eq("key", key.get()),
+        Updates.set("value", value),
+        new UpdateOptions().upsert(true));
+    return null;
```
agavra, line 122: The C* client requires us to be able to batch statements of different types (insert/delete) together and then execute them as a single batch on flush. We should change this API not to return `BoundStatement` which is specific to the Cassandra driver.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriter.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public void delete(final K key) {
+    table.delete(partition, key);
+  }
+
+  @Override
+  public CompletionStage<RemoteWriteResult> flush() {
+    return CompletableFuture.completedFuture(RemoteWriteResult.success(partition));
```
agavra, line 47: We may want to either have two different APIs -- one for sync one for async, or just use the Async mongo driver here. We'll need to do some refactoring of the API to make sure that it doesn't have anything C* specific.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Override
+  public MetadataRow metadata(final int partition) {
+    final MetadataDoc result = metadata.find(
+        Filters.eq("_id", metadataRows.get(partition))
+    ).first();
+    if (result == null) {
+      throw new IllegalStateException("Expected to find metadata row");
+    }
+    return new MetadataRow(result.offset, -1L);
```
agavra, line 139: note that i haven't done any work yet to support EOS or fencing

kafka-client/src/test/java/dev/responsive/kafka/testutils/ResponsiveExtension.java
```diff
@@ -51,12 +55,32 @@ public class ResponsiveExtension implements BeforeAllCallback, AfterAllCallback,
...
 
   public static Admin admin;
 
+  public StorageBackend backend = StorageBackend.CASSANDRA;
+
+  public ResponsiveExtension() {
+  }
+
+  public ResponsiveExtension(final StorageBackend backend) {
+    this.backend = backend;
```
agavra, line 68: we should consider extending this approach to allow for spinning up any combination of containers to speed up runs (e.g. run just with Kafka or just with Cassandra) for certain tests

kafka-client/src/main/java/dev/responsive/kafka/internal/db/TableCache.java
```diff
@@ -28,23 +28,18 @@
  * table are only prepared once during the lifetime of the application.
  */
 @ThreadSafe
-public class TableFactory<T extends RemoteTable<?>> {
+public class TableCache<T extends RemoteTable<?>> {
```
ableegoldman, line 31: Nice, this name definitely makes more sense for the class

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.internal.db.mongo;
+
+import org.bson.types.ObjectId;
+
+public class KVDoc {
```
ableegoldman, line 21: I'm not super familiar with mongo, but I take it they refer to a key-value entry as a "doc" (document?)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
```diff
@@ -0,0 +1,67 @@
...
+
+  ObjectId _id;
```
ableegoldman, line 23: In the MetadataDoc below we use this `@BsonId` annotation, should we do the same here? Don't know anything about it other than it seems like a MongoDB thing

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
```diff
@@ -0,0 +1,67 @@
...
+    this.key = key;
+    this.value = value;
+    this.epoch = epoch;
+  }
+
+  public ObjectId _id() {
+    return _id;
+  }
+
+  public void set_id(final ObjectId _id) {
```
ableegoldman, line 41: What is this, C++? 😛 
```suggestion
  public void setId(final ObjectId id) {
```
ditto everywhere else

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
ableegoldman: Makes sense -- can we leave a TODO or do you plan to follow up on this immediately?

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/SharedClients.java
```diff
@@ -1,32 +1,87 @@
...
 
 /**
  * Basic container class for session clients and other shared resources that should only
  * be closed when the app itself is shutdown
  */
 public class SharedClients {
-  public final CassandraClient cassandraClient;
-  public final Admin admin;
+
+  private final StorageBackend storageBackend;
```
ableegoldman, line 19: I'd prefer not to store this as a separate field, but instead to return the backend type based on whether the mongo or C* client is present -- let's figure it out based on the actual clients so we're guaranteed to reflect reality

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/TableName.java
```diff
@@ -32,12 +32,12 @@ public class TableName {
...
 
   private final String kafkaName;
-  private final String cassandraName;
+  private final String remoteName;
 
   public TableName(final String kafkaName) {
     this.kafkaName = kafkaName;
     final var escaped = kafkaName.replaceAll("_", "__"); // escape existing underscores
-    cassandraName = INVALID_CASSANDRA_CHARS
+    remoteName = INVALID_CASSANDRA_CHARS
```
ableegoldman, line 40: Out of curiosity, have you looked into the valid character set for mongo db names?

Also (and I'm not saying this is necessary the right thing to do) did you consider whether to enforce client-specific naming? I'm not SUPER worried about this, but for example there is a bug in the C* naming scheme since we flatter everything to lowercase whereas Streams will consider case when validating state store names. If MongoDB doesn't have the same constraints as C*, we could at least keep the bug contained to C*

Otherwise I think we can probably see all the state store names from the TopologyDescription, so we could parse it and enforce case-insensitive uniqueness of state store names ourselves

responsive-test-utils/src/main/java/dev/responsive/kafka/internal/clients/TTDCassandraClient.java
```diff
@@ -42,16 +42,16 @@ public class TTDCassandraClient extends CassandraClient {
   private final ResponsiveStoreRegistry storeRegistry = new ResponsiveStoreRegistry();
   private final TTDMockAdmin admin;
 
-  private final TableFactory<RemoteKVTable> kvFactory;
-  private final TableFactory<RemoteWindowedTable> windowedFactory;
+  private final TableCache<RemoteKVTable> kvFactory;
+  private final TableCache<RemoteWindowedTable> windowedFactory;
 
   public TTDCassandraClient(final TTDMockAdmin admin, final Time time) {
```
ableegoldman, line 48: Another thing that would be nice about generalizing the storage backend at the Client/ClientFactory level (ie having the CassandraClient and MongoClient implement a common client & factory interface vs branching immediately based on enums) is that we could probably throw out a lot of the gross hacks we currently put into the TTD version of the C* client, and just have a separate TTDClient and TTDClientFactory that is completely state-store agnostic.

We could even make it fully pluggable so that users could choose whether to run with the mock client or use one of the real storage backends, eg with a container setup. I think that would be pretty slick

Obviously can/should be a followup project, just thinking out loud here

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriter.java
ableegoldman: Just making sure I understand: the C* client is currently/already async, so the choice is between introducing a synchronous API for the mongo client vs plugging in the async mongo client -- is that correct?

If so, two followup questions:
1. why not just plug in the async mongo driver right away?
2. it looks like we're just automatically plugging the results into a completed future/stage in both clients right now, so it's not truly async at the moment anyways. Do we want to make it be truly async at some point in the near future? Like I guess we could do something pretty much exactly the same as the recent changelog truncation optimization -- we don't actually need to know if it failed/was offset fenced during the Kafka commit, since either way we only flush after the txn/offsets are committed to kafka anyways. So we could just kick off a new flush after a commit in the same way that we kick off a new DeleteRecords request, and throw a fenced exception if it turns out the flush did indeed fail

I guess I'm getting ahed of things here, obviously this doesn't need to happen as part of this change. Just trying to envision the future and why we would/would not want the async driver

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/ResponsiveMongoClient.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private final TableCache<RemoteKVTable> cache;
+  private final MongoClient client;
+
+  public ResponsiveMongoClient(final MongoClient client) {
+    this.client = client;
+    cache = new TableCache<>(spec -> new MongoKVTable(client, spec.tableName()));
+  }
+
+  public RemoteKVTable kvTable(final String name) throws InterruptedException, TimeoutException {
```
ableegoldman, line 35: It's not a big deal right now if things are working, but presumably we will eventually want the clients to implement a common interface and integrate into the new Spec/Operations structure, right?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -117,6 +115,29 @@ public static PartitionedOperations create(
     );
   }
 
+  private static RemoteKVTable createCassandra(
```
ableegoldman, line 118: I won't push back on it if you just want to get something merged, but I'm not a fan of this structure or these methods being here/the way they're named.

Seems like we should have a client interface, perhaps with just these methods: `createKVTable`, and `createWindowTable` (and eventually I guess `createSessionTable`. The client interface could also have a `clientType` method that returns `CASSANDRA` or `MONGO_DB` and let us remove the enum entirely (assuming we also replace the config value with plain strings)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
```diff
@@ -94,18 +91,9 @@ public void init(final StateStoreContext context, final StateStore root) {
...
-        operations = PartitionedOperations.create(
-            name,
-            context,
-            params,
-            spec
-        );
-      }
+      operations = (taskType == TaskType.GLOBAL)
+          ? GlobalOperations.create(context, params)
+          : PartitionedOperations.create(name, context, params);
```
ableegoldman, line 96: this is a nice simplification!

ableegoldman (approved):
Alright...lots of comments. Most of them are larger structural feedback about how to fit this into our existing framework and specifically the new table/spec/operations approach (to the extent it makes sense to do so) 

I know this is just the starter PR so you might have already thought of many of these suggestions and just didn't want/have time to fit it into the first iteration. Feel free to merge it if the pluggable client is working, and we can iterate from there

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -417,4 +356,93 @@ public boolean close(final CloseOptions options) {
...
+      final var admin = responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals());
+
+      final var backendType = StorageBackend.valueOf(responsiveConfig
+          .getString(STORAGE_BACKEND_TYPE_CONFIG)
+          .toUpperCase(Locale.ROOT)
+      );
+
+      switch (backendType) {
+        case CASSANDRA:
+          final var cqlSession = cassandraFactory.createCqlSession(responsiveConfig);
```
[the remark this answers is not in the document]
agavra: yes! I was imagining something pretty similar, but didn't want it to spiral at this point. all of this should definitely be encapsulated away somewhere so that we can also more easily test different implementations instead of repeating this same logic throughout

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -68,6 +68,10 @@ public class ResponsiveConfig extends AbstractConfig {
   public static final String CLIENT_SECRET_CONFIG = "responsive.client.secret";
   private static final String CLIENT_SECRET_DOC = "The client secret for authenticated access";
 
+  public static final String STORAGE_BACKEND_TYPE_CONFIG = "responsive.storage.backend.type";
```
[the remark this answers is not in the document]
agavra: enums are nice for `switch` statements in the code b/c IntelliJ can autofill them - otherwise I don't have a strong preference so I can change it. I guess if the config is well enough encapsulated I shouldn't be checking it in too many places anyway ;)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
agavra: MongoDB only has "documents" -- they don't even directly expose the "key", which is an `ObjectId`. I still need to figure out exactly how Mongo works with this regard, but essentially yes everything that's inserted into Mongo is considered a "document"

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
agavra: good catch

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
agavra: ah I have actually updated this and forgot to push the update, but just so you know the original reason is that Mongo has a field that is autogenerated called `_id` and in order to access it I thought the name here has to match. but they actually will magically replace any `id` with `_id` underneath the hood so it isn't necessary.

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/TableName.java
agavra: yeah that was something I was going to do as a follow-up, we probably should be enforcing client specific naming but for now I just did it this way

responsive-test-utils/src/main/java/dev/responsive/kafka/internal/clients/TTDCassandraClient.java
agavra: that would be pretty cool indeed! I'll turn this into a ticket so we don't forget about it. #165

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/ResponsiveMongoClient.java
agavra: yup, that's definitely something I'd like to do

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
agavra: definitely agree that the current state of things isn't in tip top shape from abstractions, this was what I could throw together with the fewest lines of code changed while still improving the situation.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriter.java
agavra: 1. Async is generally more complicated and I wanted to optimize for simplicity for the first go around. Also curious about non-async performance since if it's good enough we can just batch the entire thing in the background
2. I don't think that's true for the `LwtWriter` (it is true for the `FactWriter`), which does return a future that isn't complete. I do think we need to know at least before the _next_ flush whether or not it succeeded, so making that optimization is a bit more difficult. We wouldn't want to flush the next batch if the previous batch failed to flush or we'd corrupt the remote store.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
agavra: I plan to do that relatively soon, it's bugging me quite a bit 😆 and I think it'll become more important when I do EOS support.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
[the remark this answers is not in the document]
agavra: we can refactor this later since it's totally backwards compatible (note that the config itself is actually a String, I'm just maintaining it as an enum)

PR #166, opened 2023-10-13
Maintain the shared session clients as a singular container when passed via internal configs

ableegoldman:
Relatively small in scope refactoring to change the way the `SharedClients` class is handled. Right now we assemble and disassemble the entire `SharedClients` every time we need to access something from it, regardless of whether we in fact needed the entire object or just a single one of its clients. This also ensures that something added to or modified in the SharedClients container is percolated throughout the application (for example if we need to add another client after going through the `KafkaStreams` constructor, which is currently impossible to do).

I'm not leveraging either of these improvements in this PR, but I want to set the stage for some other things and make it easier to pass things around without having to be fully initialized up front.

I also renamed `SharedClients` to `SessionClients`, which is probably why there are so many lines of changes (the actual change in behavior is fairly narrow in scope)

Reviewers can focus on just these two classes: `SessionClients` (formerly SharedClients)  and `InternalSessionConfigs` (formerly InternalConfigs)


agavra (approved):
This change makes a lot of sense to me! I was actually thinking of coupling that with my previous change.

PR #167, opened 2023-10-13
More application health metrics: restoration and thread errors

ableegoldman:
It'd be nice to be able to confirm that our tasks are in fact spending little (or no) time in restoration. This is less about verifying the fundamental hypothesis of remote stores and more about general monitoring of application stability. Unlike with plain Kafka Streams, a high value for any of these metrics in one of our applications would indicate an unhealthy state and should be considered a signal to investigate

To that end, this PR adds three new Responsive metrics using the custom restore listener and one existing Streams metric for the number of "failed threads" (ie thread death & replacement events)

1. num-restoring-changelogs -- simple count of stores currently in restoration
2. num-interrupted-changelogs -- counter for stores being interrupted before finishing restoration
3. restore-latency -- how long a store spent being restored from its changelog
4. failed-stream-threads -- how many times has a thread been killed and replaced on this client


kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.kafka.common.metrics.stats.Max;
+import org.apache.kafka.streams.processor.StateRestoreListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Note: this is a global object which is shared across all StreamThreads and state stores in
+ * this Streams application.
+ */
+public class ResponsiveRestoreListener implements StateRestoreListener, Closeable {
```
ableegoldman, line 48: git's being dumb, I didn't add this class here I just moved it to a different package.

That said, almost all of the code in the listener is in fact new to this PR, so don't skip over it during review 🙏

ableegoldman:
Just going to merge this since I want to open a followup PR that depends on it, any feedback on this one I can address immediately in the followup

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,227 @@
...
+  public ResponsiveRestoreListener(final ResponsiveMetrics metrics) {
+    this.metrics = metrics;
+
+    numRestoringMetricName = metrics.metricName(
+        NUM_RESTORING_CHANGELOGS,
+        NUM_RESTORING_CHANGELOGS_DESCRIPTION,
+        metrics.applicationLevelMetric()
+    );
+    numInterruptedSensor = metrics.addSensor(NUM_INTERRUPTED_CHANGELOGS);
+    restoreLatencySensor = metrics.addSensor(RESTORE_LATENCY);
```
rodesai, line 69: doesn't this sensor need to be store-level? Otherwise the average/max metric tagged with a given store will reflect the average/max for all stores

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
```diff
@@ -0,0 +1,227 @@
...
+  public synchronized void onRestoreStart(
+      final TopicPartition topicPartition,
+      final String storeName,
+      final long startingOffset,
+      final long endingOffset
+  ) {
+    final long startMs = System.currentTimeMillis();
+    restoringChangelogToStartMs.put(topicPartition, startMs);
+
+    restoreLatencySensor.add(
```
rodesai, line 103: it feels a bit weird to track this using windowed stats live average/max. You'd see NaN (or maybe 0?) except for the window where restore finished and then you'd see one data point that has the restore latency, and then back to NaN/0. Maybe it would make more sense as a gauge that reflects the current time the store has been restoring, and that returns -1 or 0 when the store is not restoring? That way we can detect when something has been restoring for a long time?

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
ableegoldman: Ah, yeah it's a bit confusing but the actual metric is indeed store-level. We're just creating the sensor here but don't actually add any metrics until we have a state store begin restoration. If you look at `#onRestoreStart` you'll see where we register the actual metrics to this sensor, and those metrics are in the store-level metrics group/tagged with the specific state store name

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
ableegoldman: Mm yeah I suspected this wasn't quite right but thought it'd be better than nothing...your suggestion makes a lot more sense, will open a PR with this change

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveRestoreListener.java
rodesai: I think you still need store-level sensors in this scenario. For example, if you do some sequence like:

```sensor = metrics.addSensor(RESTORE_LATENCY)
...
sensor.addMetric(metricName(... args for store 1....), new Max())
sensor.addMetric(metricName(... args for store 2...), new Max())
...
// store 1 finishes restoring after 5 seconds)
sensor.record(5000)
...
// store 2 finishes restoring after 5.5 seconds)
sensor.record(5500)
```

Then, both the store 1 metric and store 2 metric will report 5500 as the max latency, where I think what you're going for is for the store-level metrics to report an aggregate over just the specific store. Basically there's no parameter in `sensor.record` that lets you scope the sample to a specific store, so you have to create store-level sensors.

ableegoldman: Ah, you're totally right, will do another quick followup PR

ableegoldman: Oh wait nevermind, this is addressed in the followup PR I already merged

PR #169, opened 2023-10-14
More application health metrics: flushing, fencing, and truncation in the CommitBuffer

ableegoldman:
Next set of metrics pertaining to some of the core functionality in the CommitBuffer. Some of these like the flush latency might be available/tracked natively by the underlying storage engine, but it still makes sense to report it as a Responsive metric so that it's in the same place as our other metrics, has the full set of tags, and we can be sure it reflects the full span of time spent flushing (eg we don't need to consider the subpartitioning scheme)

New metrics:
1. time-since-last-flush (gauge in milliseconds):
2. flushes (sensor: rate & total)
3. flush-latency (sensor: avg & max)
4. commits-fenced (sensor: rate & total)
5. failed-truncations (sensor: rate & total)

Note: I left this out of scope for now, but I think it might also be useful to report a rate & total metric for each of the flush triggers. I'll follow up with these in a separate PR, as well as anything else that we think might be useful so don't hesitate to recommend other related metrics that you want


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -158,10 +201,71 @@ static <K, S extends RemoteTable<K>> CommitBuffer<K, S> from(
...
+        TIME_SINCE_LAST_FLUSH,
+        TIME_SINCE_LAST_FLUSH_DESCRIPTION,
+        metrics.storeLevelMetric(metrics.computeThreadId(), changelog, storeName)
+    );
+    metrics.addMetric(
+        lastFlushMetric,
+        (Gauge<Long>) (config, now) -> now - lastFlush.toEpochMilli()
+    );
+
+    flushesSensor = metrics.addSensor(FLUSH);
```
rodesai, line 219: this name should be unique to this store, otherwise you'll get a single sensor for all stores and mix up samples

ableegoldman: Ah, good catch. I'm still adding unit tests for these metrics, I'll make sure to have coverage for this case

ableegoldman:
Redeployed the soak with these metrics, looking good: 
image

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -75,6 +106,17 @@ class CommitBuffer<K, S extends RemoteTable<K>>
   private final Supplier<Instant> clock;
   private final KeySpec<K> keySpec;
 
+  private final String flushSensorName;
+  private final String flushLatencySensorName;
+  private final String flushFencedSensorName;
+  private final String failedTruncationsSensorName;
+
+  private final MetricName lastFlushMetric;
```
agavra, line 114: we can do this in a follow up, but it'd be nice if we grouped all the metrics together in a single CommitBufferMetrics class just to reduce the verbosity of this (already relatively large) file

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -498,11 +641,17 @@ private void throwFencedException(final RemoteWriteResult result, final long con
     );
     log.warn(msg);
 
+    flushFencedSensor.record();
```
agavra, line 644: I double checked because I wasn't sure this was still right and indeed this code is no longer only called when we're fenced. any time the write fails, this code path will be triggered

we should probably update the entire error handling path here, but for now we can just rename the metric to reflect that it's any error in flush.

agavra (approved):
we should rename the `flush-fenced` metric, but otherwise LGTM

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
agavra: talked offline, I'll push this change to the branch and merge the PR

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
ableegoldman: meant to follow up on this, yeah I would definitely like to consolidate the actual metric/sensor creation into the corresponding metrics group class. Both to reduce the verbosity and to keep all the metrics in one place so it's easy to find them and to tell what metrics are actually being added and how. I left that out of scope for now but I do plan on adding a few more batches of metrics maybe after the window store stuff, at which point I think it makes sense to do this cleanup

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
ableegoldman: nice catch -- if anything it's probably more useful this way as we'd want to know if flushes fail whatever the reason may be.

If we realize that we do need a more fine-grained look then we can always go back and add additional metrics for the specific type of error, although generally I feel it's easy enough to check the logs and figure it out if/when we see a spike in the flush-errors metric

PR #171, opened 2023-10-16
make the Table API generic to backend storage

agavra:
Before this PR the `RemoteTable` API assumed Cassandra and therefore encoded `BoundStatement` in it. This sets the stage for returning something different for different backend implementations.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
@@ -19,6 +19,7 @@ package dev.responsive.kafka.internal.db;
 import com.datastax.oss.driver.api.core.cql.BatchStatementBuilder;
 import com.datastax.oss.driver.api.core.cql.BatchType;
 import com.datastax.oss.driver.api.core.cql.BatchableStatement;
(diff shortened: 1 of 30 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
```diff
@@ -19,6 +19,7 @@
 import com.datastax.oss.driver.api.core.cql.BatchStatementBuilder;
 import com.datastax.oss.driver.api.core.cql.BatchType;
 import com.datastax.oss.driver.api.core.cql.BatchableStatement;
+import com.datastax.oss.driver.api.core.cql.BoundStatement;
```
rodesai, line 22: shall we move this and other sylla implementation types to a subpackage (like `mongo`)

rodesai (approved):
LGTM

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
agavra: 👍 yes that was my intention, but I didn't want this PR to get hard to review and touch a million files.

PR #176, opened 2023-10-19
[mongo] insert in bulk to avoid network roundtrips

agavra:
Note that batches in mongo are _not_ atomic so it's possible that any number of writes within a batch will fail. That's OK because newer writers will succeed in overwriting anything with older epochs (not yet implemented, that's in the next PR that has EOS).

When I have Mongo fully implemented I'll add a suite of unit/integration tests.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
@@ -523,7 +523,7 @@ class CommitBuffer<K, S extends RemoteTable<K, ?>>
     flushSensor.record(1, endMs);
     flushLatencySensor.record(flushLatencyMs, endMs);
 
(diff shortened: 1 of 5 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -523,7 +523,7 @@ private void doFlush(final long consumedOffset, final int batchSize) {
     flushSensor.record(1, endMs);
     flushLatencySensor.record(flushLatencyMs, endMs);
 
-    log.debug("Flushed {} records to remote in {}ms (offset={}, writer={}, numSubPartitions={})",
+    log.info("Flushed {} records to remote in {}ms (offset={}, writer={}, numSubPartitions={})",
```
agavra, line 526: I found this log to be very useful to see how long a specific flush for X records takes so I added it back as info cc @ableegoldman

rodesai (approved):
LGTM

PR #178, opened 2023-10-21
Add WindowOperations and SegmentedOperations

ableegoldman:
Just a minor step in the window store implementation, most of the interesting stuff will be in the followup with the subpartitioner (which for now is just a TODO)

This PRs adds a WindowOperations interface with just the single SegmentedOperations implementation for now. Although there's only the one implementation of this, it's still nice to break up everything that's currently stuffed into the `ResponsiveWindowStore` implementation.

With this, all of the Streams/WindowStore semantics and logic stays in the ResponsiveWindowStore class while the actual storage engine/implementation details will go in the SegmentedOperations

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
@@ -18,7 +18,6 @@ package dev.responsive.kafka.internal.db;
 
 import static org.apache.kafka.streams.state.StateSerdes.TIMESTAMP_SIZE;
 
-import dev.responsive.kafka.internal.stores.ResponsiveWindowStore;
 import dev.responsive.kafka.internal.utils.Stamped;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
@@ -58,6 +57,16 @@ public class StampedKeySpec implements KeySpec<Stamped<Bytes>> {
 
   @Override
   public int compare(final Stamped<Bytes> o1, final Stamped<Bytes> o2) {
-    return ResponsiveWindowStore.compareKeys(o1, o2);
+    return compareKeys(o1, o2);
+  }
+
+  // TODO remove generic from Stamped<> and just implement Comparable
+  public static int compareKeys(final Stamped<Bytes> o1, final Stamped<Bytes> o2) {
+    final int key = o1.key.compareTo(o2.key);
+    if (key != 0) {
+      return key;
+    }
+
+    return Long.compare(o1.stamp, o2.stamp);
   }
 }
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/KeyValueOperations.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/KeyValueOperations.java
@@ -23,10 +23,6 @@ import org.apache.kafka.streams.state.KeyValueIterator;
 
 public interface KeyValueOperations extends Closeable, RecordBatchingStateRestoreCallback {
 
-  void register(ResponsiveStoreRegistry storeRegistry);
-
-  void deregister(ResponsiveStoreRegistry storeRegistry);
-
   void put(final Bytes key, final byte[] value);
 
   byte[] delete(final Bytes key);
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
@@ -19,12 +19,9 @@ package dev.responsive.kafka.internal.stores;
 import static org.apache.kafka.streams.processor.internals.ProcessorContextUtils.asInternalProcessorContext;
 
 import dev.responsive.kafka.api.stores.ResponsiveKeyValueParams;
-import dev.responsive.kafka.internal.config.InternalSessionConfigs;
 import dev.responsive.kafka.internal.utils.TableName;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.TimeoutException;
-import java.util.function.Function;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.streams.KeyValue;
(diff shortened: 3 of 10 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/KeyValueOperations.java
```diff
@@ -23,10 +23,6 @@
 
 public interface KeyValueOperations extends Closeable, RecordBatchingStateRestoreCallback {
 
-  void register(ResponsiveStoreRegistry storeRegistry);
```
ableegoldman, line 26: It seems silly to have the registry still live in the KeyValue/WindowStore while moving the registration to the KeyValue/WindowOperations class and just call through these extra de/register APIs.

Basically I moved all the Responsive logic into the Operations implementations, so that the Window/KeyValueStores themselves are now a fairly clean implementation of only what's relevant to Kafka Streams and the Responsive stuff is all encapsulated inside the various Operations

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
```diff
@@ -118,9 +110,7 @@ public void init(final StateStoreContext storeContext, final StateStore root) {
       operations = opsProvider.provide(params, storeContext, taskType);
       log.info("Completed initializing state store");
 
-      storeRegistry = registryProvider.apply(storeContext.appConfigs());
       open = true;
-      operations.register(storeRegistry);
```
ableegoldman, line 123: No need to explicitly register anything now, we just take care of everything when the Operations are created (or closed)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
```diff
@@ -232,8 +219,4 @@ public KeyValueIterator<Bytes, byte[]> reverseAll() {
     return operations.reverseAll();
   }
 
-  // Visible for testing
-  KeyValueOperations getOperations() {
```
ableegoldman, line 236: Didn't seem to be used (we can of course add it back if any test actually does need it, but I don't think we need this AND the `kvOperationsProvider`)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
```diff
@@ -58,6 +57,16 @@ public boolean retain(final Stamped<Bytes> key) {
 
   @Override
   public int compare(final Stamped<Bytes> o1, final Stamped<Bytes> o2) {
-    return ResponsiveWindowStore.compareKeys(o1, o2);
+    return compareKeys(o1, o2);
+  }
+
+  // TODO remove generic from Stamped<> and just implement Comparable
```
agavra, line 63: by the way, we can make `Stamped<K extends Comparable>` if we did want to keep the generic (just fun java tip, I'm still in favor of removing the generic here)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStore.java
agavra: ah yes, I had written a test to make sure it creates the right operations (global vs local) but then realized the test was kind-of nonsense sine it mocked so much. I deleted it and forgot to update this

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SegmentedOperations.java
```diff
@@ -0,0 +1,306 @@
...
+      final Bytes key,
+      final long timeFrom,
+      final long timeTo
+  ) {
+    final Stamped<Bytes> from = new Stamped<>(key, timeFrom);
+    final Stamped<Bytes> to = new Stamped<>(key, timeTo);
+
+    final int subPartition = partitioner.partition(changelog.partition(), key);
+    return Iterators.windowed(
+        new LocalRemoteKvIterator<>(
```
agavra, line 262: I think `LocalRemoteKVIterator` may have a bug in reverse fetching since it assumes keys are compared in ascending order - we should probably pass in the iteration order

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SegmentedOperations.java
```diff
@@ -0,0 +1,306 @@
...
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.common.utils.LogContext;
+import org.apache.kafka.streams.kstream.Windowed;
+import org.apache.kafka.streams.processor.StateStoreContext;
+import org.apache.kafka.streams.processor.internals.InternalProcessorContext;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.apache.kafka.streams.state.WindowStoreIterator;
+
+public class SegmentedOperations implements WindowOperations {
```
agavra, line 51: Shame on me for not having done this for the other `*Operations` classes (we can do this one as well as a follow up) but we should figure out a testing strategy for it.

agavra (approved):
LGTM, I think there's one bug we should fix or make sure to put on our radar (unrelated to this PR)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
ableegoldman: Yeah this comment is kind of mixing concerns, ultimately I'd like to remove the generic, it just so happens that this also means we can clean up some of the comparing related code

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SegmentedOperations.java
ableegoldman: Good catch -- I'm just going to file a followup ticket for this for now and tackle it after the main window store PR. I added another new kind of iterator in the next PR for segmented range scans but was planning to leave the implementation of it for a followup. I'll tack this onto that patch

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SegmentedOperations.java
ableegoldman: I actually filed a ticket a while ago for building out a unit testing framework for the state store APIs -- I'll think about it once the baseline window store implementation is done

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SegmentedOperations.java
ableegoldman: FYI: https://www.notion.so/responsivedev/Iterators-for-range-scans-6aa2622b2b0b4b01a95eaabdebba347c?pvs=4

PR #179, opened 2023-10-22
Remove batchSize parameter from writers

ableegoldman:
Relatively small refactor to clear up how batches are handled/sized between the CommitBuffer and RemoteWriters. It seems that the `batchSize` we pass in is always either the `MAX_BATCH_SIZE` constant or, in the case of restoration only, the size of the final batch that contains whatever records are remaining up to the `MAX_BATCH_SIZE`

In other words, the batch size is always capped at `MAX_BATCH_SIZE` and furthermore, the only time the `batchSize` parameter is different than the hard-coded constant is when there are fewer updates in the batch and thus the batch will be naturally capped at this smaller size since we will run out of inserts (or deletes) before hitting the `batchSize` condition. Also, only the lwt writer seems to be using this at the moment, so it feels silly for us to pass this in to every other writer as we do right now.

In addition to just being easier to reason about and removing unnecessarily complexity from the code, this is a necessary cleanup/precursor for the window store implementation

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
@@ -23,6 +23,7 @@ import com.datastax.oss.driver.api.core.cql.BatchType;
 import com.datastax.oss.driver.api.core.cql.BatchableStatement;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
 import dev.responsive.kafka.internal.stores.RemoteWriteResult;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
```diff
@@ -34,22 +37,38 @@ public class LwtWriter<K> implements RemoteWriter<K> {
...
+      final Supplier<BatchableStatement<?>> fencingStatementFactory,
+      final RemoteTable<K, BoundStatement> table,
+      final int partition
+  ) {
+    this(
+        client,
+        fencingStatementFactory,
+        table,
+        partition,
+        MAX_BATCH_SIZE
```
ableegoldman, line 55: This is the only real/logical change: we now always just pass in the actual `MAX_BATCH_SIZE` (although we still parameterize it within this class rather than just using hard-coding the constant directly so we can set it to something lower for the unit tests)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriter.java
```diff
@@ -74,7 +93,7 @@ public CompletionStage<RemoteWriteResult> flush() {
       builder.setIdempotence(true);
       builder.addStatement(fencingStatementFactory.get());
 
-      for (int i = 0; i < batchSize && it.hasNext(); i++) {
+      for (int i = 0; i < maxBatchSize && it.hasNext(); i++) {
```
ableegoldman, line 96: Note that when the actual batch size is less than the `MAX_BATCH_SIZE`, the `it.hasNext` condition will catch the end of the batch anyways and so we don't need to pass in the batchSize as we were before

PR #180, opened 2023-10-23
[mongo] eos support with epoch based fencing

agavra:
### Review Guide

There are basically two changes:

1. pipe `epoch` through to where it's necessary
2. Changes in `MongoKVTable` to support EOS

You really only need to review the second to get a sense of whats' going on.

I tested it by running the EOS test against MongoDB, though I'll follow up this PR with a more significant testing suite for MongoDB stores in general.

#### Implementing Epoch-Based Fencing

MongoDB supports atomic updates on individual documents via the [findAndModify](https://www.mongodb.com/docs/manual/reference/method/db.collection.findAndModify/#std-label-findAndModify-upsert-example) upsert operation of the form:

```jsx
db.example.findAndModify({
    query: { key: "0xDEADBEEF", epoch: { $lte: local_epoch } },
    update: { $set: { key: "0xDEADBEEF", value: "0xCAFED00D", epoch: local_epoch } },
    upsert: true
})
```

The `$lte: local_epoch` ensures that the document will not be retrieved at all if the current document has been written by a writer that has a higher epoch. So long as processing deterministically updates the same keys, this should be a safe operation (which is a requirement for EOS in Kafka Streams anyway).

****************************************Managing Tombstones.**************************************** Deletes are more complicated to handle since it is possible that a future writer *deleted* a document. If we were to simply remove those documents from the database, a zombie would be able to revive the document because it would consider it would successfully upsert. 

To handle this, we'll insert a special timestamp field *only for deleted documents* that includes the time in which it was deleted and build a [TTL index](https://www.mongodb.com/docs/manual/core/index-ttl/) on top of it to automatically expire it after some grace period (this is very similar to Cassandra's `gc_grace_seconds` and Kafka's `delete.retention.ms` configurations):

```jsx
db.example.createIndex( { "tombstoneTs": 1 }, {expireAfterSeconds: 86400 } )
db.example.findAndModify({
    query: { key: "0xDEADBEEF", epoch: { $lte: local_epoch } },
    update: { $set: { key: "0xDEADBEEF", value: null, epoch: local_epoch, tombstoneTs: now() } },
})
```

Note that we will have to add `{ $unset: tombstoneTs }` to the normal inserts to ensure we don't accidentally remove a document if it is intentionally revived.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -52,17 +52,20 @@ import org.slf4j.LoggerFactory;
 public class MongoKVTable implements RemoteKVTable<WriteModel<Document>> {
 
   private static final Logger LOG = LoggerFactory.getLogger(MongoKVTable.class);
+  private static final String METADATA_COLLECTION_SUFFIX = "_md";
 
   private final String name;
-  private final MongoCollection<KVDoc> collection;
+  private final MongoCollection<KVDoc> docs;
   private final MongoCollection<MetadataDoc> metadata;
+
+  private final MongoCollection<Document> genericDocs;
+  private final MongoCollection<Document> genericMetadata;
 
   // this map contains the initialized value of the metadata document,
   // for which the epoch and object ID will never change. it is likely
   // that after a few writes to MongoDB the cached MetadataDoc.offset value
   // will be out of date, so it should not be used beyond initialization
   private final ConcurrentMap<Integer, MetadataDoc> metadataRows = new ConcurrentHashMap<>();
-  private final MongoCollection<Document> generic;
 
   public MongoKVTable(final MongoClient client, final String name) {
     this.name = name;
@@ -73,13 +76,15 @@ public class MongoKVTable implements RemoteKVTable<WriteModel<Document>> {
     );
 
     final MongoDatabase database = client.getDatabase(name).withCodecRegistry(pojoCodecRegistry);
-    generic = database.getCollection(name);
-    collection = database.getCollection(name, KVDoc.class);
-    metadata = database.getCollection(name, MetadataDoc.class);
+    genericDocs = database.getCollection(name);
+    docs = database.getCollection(name, KVDoc.class);
+
+    genericMetadata = database.getCollection(name + METADATA_COLLECTION_SUFFIX);
+    metadata = database.getCollection(name + METADATA_COLLECTION_SUFFIX, MetadataDoc.class);
 
     // TODO(agavra): make the tombstone retention configurable
     // this is idempotent
-    collection.createIndex(
+    docs.createIndex(
         Indexes.descending(KVDoc.TOMBSTONE_TS),
         new IndexOptions().expireAfter(12L, TimeUnit.HOURS)
     );
@@ -114,12 +119,12 @@ public class MongoKVTable implements RemoteKVTable<WriteModel<Document>> {
       return metaDoc;
     });
 
-    return new MongoWriterFactory<>(this, generic, kafkaPartition);
+    return new MongoWriterFactory<>(this, genericDocs, genericMetadata, kafkaPartition);
   }
 
   @Override
   public byte[] get(final int kafkaPartition, final Bytes key, final long minValidTs) {
-    final KVDoc v = collection.find(Filters.eq(KVDoc.ID, key.get())).first();
+    final KVDoc v = docs.find(Filters.eq(KVDoc.ID, key.get())).first();
     return v == null ? null : v.getValue();
   }
 
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -24,32 +24,44 @@ import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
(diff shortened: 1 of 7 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriterFactory.java
```diff
@@ -28,21 +32,37 @@ public class MongoWriterFactory<K> implements WriterFactory<K> {
...
+    final MetadataDoc updateResult = metadata.findOneAndUpdate(
+        Filters.eq("partition", kafkaPartition),
+        Updates.inc("epoch", 1),
+        new FindOneAndUpdateOptions()
+            .upsert(true)
+            .returnDocument(ReturnDocument.AFTER)
+    );
+
+    if (updateResult == null) {
+      throw new IllegalStateException("Uninitialized metadata for partition " + kafkaPartition);
```
rodesai, line 54: shouldn't this be a task migrated exception?

agavra: not sure why it would be that? this should never happen because the previous statement creates the document if it doesn't exist (there's no fencing that happens here)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriterFactory.java
```diff
@@ -28,21 +32,37 @@ public class MongoWriterFactory<K> implements WriterFactory<K> {
...
+      final MongoCollection<Document> genericCollection,
+      final MongoCollection<MetadataDoc> metadata,
+      final int kafkaPartition
   ) {
     this.table = table;
     this.genericCollection = genericCollection;
+    final MetadataDoc updateResult = metadata.findOneAndUpdate(
+        Filters.eq("partition", kafkaPartition),
+        Updates.inc("epoch", 1),
+        new FindOneAndUpdateOptions()
```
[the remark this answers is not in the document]
agavra: yup

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -121,19 +138,36 @@ public WriteModel<Document> insert(
...
       final byte[] value,
-      final long epochMillis
+      final long epochMillis,
+      final long epoch
   ) {
     return new UpdateOneModel<>(
-        Filters.eq("key", key.get()),
-        Updates.set("value", value),
+        Filters.and(
+            Filters.eq("key", key.get()),
```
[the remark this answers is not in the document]
agavra: 👍 fair enough

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWriterFactory.java
rodesai: Ah right - not sure what I was thinking. Never mind.

apurvam:
> The $lte: local_epoch ensures that the document will not be retrieved at all if the current document has been written by a writer that has a higher epoch. So long as processing deterministically updates the same keys, this should be a safe operation (which is a requirement for EOS in Kafka Streams anyway).

And the epoch management is the same as we have today? As a refresher to me, how is the epoch bumped when a new instance comes online? 

And assuming we bump the epoch correctly in memory, if the `findAndModify` in this case doesn't return a documeunt, that means the current instance is fenced?

agavra:
> And the epoch management is the same as we have today? 

Somewhat - the major change for Mongo EOS is that the epoch is stored with every document as opposed to in a single metadata row. There is still a single metadata row that stores the most recent offset, but that is only used during initialization (see below).

> As a refresher to me, how is the epoch bumped when a new instance comes online?

Every time a state sore is initialized it reserves an epoch for that state store's partition (essentially happens when a task is initialized, so it's once per partition per producer). In MongoDB, it uses the atomic increment operator to ensure that it will increment the existing one by exactly one.

> And assuming we bump the epoch correctly in memory, if the findAndModify in this case doesn't return a documeunt, that means the current instance is fenced?

Quick clarification first, the epoch isn't bumped in memory -- it's bumped on MongoDB and the new one is returned. `findAndModify` doesn't actually differentiate from fenced vs. non-existing document because we allow upserts. If the document doesn't exist it'll create it, if it does exist and the epoch is <= it will update it, and if the document exists with a larger epoch it will do nothing. 

The fencing itself happens from Kafka, not the mongo read (basically if the current writer is a zombie then none of its writes will succeed and it will get fenced the next time it tries to poll data).

Though now that you mentioned it, I think my query is slightly off - it will actually create a new document with an older epoch instead of not creating one at all. I think it's possible to get the behavior that I want though. I'll tweak the query and update the description.

agavra:
OK, it actually works as a proper fencing mechanism if you have a unique index on key:
```
Atlas join_state_store> db.test_eos.find()
[
  {
    _id: ObjectId("653944f7e6c45197c2f11112"),
    key: '0xDEADBEEF',
    epoch: 3,
    value: '0xCAFED009'
  }
]
Atlas join_state_store> db.test_eos.findAndModify( { query: { key: "0xDEADBEEF", epoch: { $lte: 1 } }, update: { $set: { key: "0xDEADBEEF", value: "0xCAFED00D", epoch: 1 } }, upsert: true })
MongoServerError: Plan executor error during findAndModify :: caused by :: E11000 duplicate key error collection: join_state_store.test_eos index: key_1 dup key: { key: "0xDEADBEEF" }
Atlas join_state_store> db.test_eos.findAndModify( { query: { key: "0xDEADBEEF", epoch: { $lte: 3 } }, update: { $set: { key: "0xDEADBEEF", value: "0xCAFED00D", epoch: 3 } }, upsert: true })
{
  _id: ObjectId("653944f7e6c45197c2f11112"),
  key: '0xDEADBEEF',
  epoch: 3,
  value: '0xCAFED009'
}
```

PR #181, opened 2023-10-23
Finish segmented C* WindowStore implementation

ableegoldman:


kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -53,21 +53,27 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class CassandraKeyValueTable implements RemoteKVTable<BoundStatement> {
+public class CassandraKeyValueTable implements
+    RemoteKVTable<Integer, BoundStatement>,
+    RemoteLwtTable<Bytes, Integer, BoundStatement>{
```
agavra, line 58: this kind of inheritance (when you inherit from two things that extend the same base interface) feels like a code smell to me. I suspect the reason you did this is so that you can reuse the `LwtWriter` - instead is there any reason why `RemoteLwtTable` needs to extend `RemoteTable`? if we make it its own interface then `LwtWriter` can accept both a `RemoteTable` and a `RemoteLwtTable` but we just pass in the same table once (if it implements both).I think then we can also rename the interface to something like `CassandraLwtCoordinator` to indicate that it's really specific to Cassandra's EOS implementation and better explain what it does (it isn't really a "table" after all)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -190,51 +188,51 @@ public WriterFactory<Bytes> init(
             .build()
     );
 
-    return new FactWriterFactory<>(this);
+    return new FactWriterFactory<>(this, client, kafkaPartition);
   }
 
   @Override
-  public MetadataRow metadata(final int partition) {
+  public long fetchOffset(final int kafkaPartition) {
```
agavra, line 195: 👍 I think this change makes sense, probably premature optimization to couple them into a single remote call

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -244,14 +305,20 @@ public byte[] get(
 
   @Override
   public KeyValueIterator<Bytes, byte[]> range(
-      final int partition,
+      final int kafkaPartition,
       final Bytes from,
       final Bytes to,
       final long minValidTs
   ) {
+    // TODO: need to scan a range of sub-partitions to correctly implement range
```
agavra, line 313: should we throw an exception in the meantime?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -280,62 +348,101 @@ public KeyValueIterator<Bytes, byte[]> all(
...
         .setByteBuffer(DATA_KEY.bind(), ByteBuffer.wrap(key.get()));
   }
 
   @Override
-  public MetadataRow metadata(final int partition) {
-    final BoundStatement bound = getMeta
-        .bind()
-        .setInt(PARTITION_KEY.bind(), partition);
-    final List<Row> result = client.execute(bound).all();
+  public long fetchOffset(final int kafkaPartition) {
```
agavra, line 378: thanks for naming the partitions appropriately 🙏 we may even want to (in a follow up) make this type safe by introducing dedicated types (e.g. `TablePartition` and `KafkaPartition`)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -89,24 +110,59 @@ public static CassandraWindowedTable create(
...
+            .value(SEGMENT_ID.column(), bindMarker(SEGMENT_ID.bind()))
+            .value(ROW_TYPE.column(), METADATA_ROW.literal())
+            .value(DATA_KEY.column(), DATA_KEY.literal(ColumnName.METADATA_KEY))
+            .value(WINDOW_START.column(), WINDOW_START.literal(METADATA_TS))
+            .value(EPOCH.column(), bindMarker(EPOCH.bind()))
+            .ifNotExists()
+            .build()
+    );
+
+    // Note: it is safe to ignore the epoch here as we should only ever drop a segment
```
agavra, line 151: I think we have risk of data revival though. Here's the situation:
1. writer A attempts to write to segment 1 kv pair (X,Y)
2. writer A takes a walk (GC, IDK)
3. writer B is assigned as active
4. writer B successfully writes to segment 1 kv pair (X,Y)
5. time passes
6. writer B successfully expires segment 1
7. writer A wakes up and successfully writes kv pair (X, Y) to segment 1, reviving that data

We need to make sure that 7 fails, and if we check the epoch per-segment then it isn't possible. One suggestion is to have expire segment first delete everything, and then atomically write a tombstone metadata row. Not sure how to deal with expiring these tombstones though...

Perhaps we can also just be "OK" with this race condition because that expired segment will never be queried.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -370,59 +561,67 @@ public long approximateNumEntries(final int partition) {
...
    *         also the window start timestamp
    */
   @Override
   @CheckReturnValue
   public BoundStatement delete(
-      final int partitionKey,
+      final int kafkaPartition,
       final Stamped<Bytes> key
   ) {
+    final SegmentPartition segmentPartition = partitioner.tablePartition(kafkaPartition, key);
```
agavra, line 596: do we need to set `pendingFlushStreamTime` here?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/ColumnName.java
```diff
@@ -42,6 +43,7 @@ public enum ColumnName {
 
   static final Bytes METADATA_KEY
       = Bytes.wrap("_metadata".getBytes(StandardCharsets.UTF_8));
+  public static final long METADATA_SEGMENT_ID = -1L;
```
agavra, line 46: I haven't gone through to verify this in the code, but just wanted to make sure that nowhere do we use the partition key `(kafkaPartition, -1L)` in atomic batches with any "data" segment? I suspect this is just used for offset/streamtime tracking?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactWriterFactory.java
```diff
@@ -17,30 +17,62 @@
...
+    return table.name();
+  }
+
+  @Override
+  public Integer tablePartitionForKey(final K key) {
+    return partitioner.tablePartition(kafkaPartition, key);
   }
+
+  @Override
+  public RemoteWriteResult<Integer> setOffset(final long offset) {
```
agavra, line 65: it feels weird to me that writer factory is directly doing "things" outside of creating a writer?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriterFactory.java
```diff
@@ -16,208 +16,154 @@
...
+    return table.fetchOffset(kafkaPartition);
   }
 
-  private static ResultSet reserveEpoch(
-      final RemoteTable<?, BoundStatement> table,
-      final CassandraClient client,
-      final int partition,
-      final long epoch
+  @Override
+  public RemoteWriteResult<P> commitPendingFlush(
```
agavra, line 98: ditto: why is the factory doing this?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriterFactory.java
```diff
@@ -16,13 +16,103 @@
...
+      // be discarded in favor of the next, and real, WriteResult in the loop below
+      CompletionStage<RemoteWriteResult<P>> partialResult =
+          CompletableFuture.completedStage(RemoteWriteResult.success(ignored));
+      for (final RemoteWriter<K, P> writer : activeWriters.values()) {
+        partialResult = partialResult.thenCombine(writer.flush(), (one, two) -> !one.wasApplied() ? one : two);
+      }
+
+      try {
+        final RemoteWriteResult<P> result = partialResult.toCompletableFuture().get();
+        log.debug("Succesfully flushed writes to table partition {}", result.tablePartition());
```
agavra, line 106: do we need to check if the result was applied before logging this?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/ResponsivePartitioner.java
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.internal.db.partitioning;
+
+public interface ResponsivePartitioner<K, P> {
```
agavra, line 19:
```suggestion
public interface RemoteTablePartitioner<K, P> {
```
just a suggestion as adding `Responsive` doesn't actually add much value

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * A segment-based partitioner where kafka partitions are mapped to a subset of remote partitions
+ * divided up into "segments" that correspond to a range of timestamps. By deleting a table
+ * partition we can "drop" a full segment at a time, which can be done if (and only if) the
+ * stream-time advances such that the highest timestamp in the range is now outside the retention
+ * period. This is intended for segmented WindowStores, but could be extended in the future to
+ * other timeseries and/or ttl types of state stores.
+ * <p>
+ * Each segment covers a specific time range in milliseconds. If t_n is the start time of the
+ * nth segment, then that segment covers all timestamps from t_s - t_s + segmentInterval.
```
agavra, line 40:
```suggestion
 * nth segment, then that segment covers all timestamps from t_s -> t_s + segmentInterval.
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -0,0 +1,194 @@
...
+ * <p>
+ * NOTE: because we are already dividing up the partition space into segments, we don't further
+ * split things into true sub-partitions based on key. Each kafka partition still maps to multiple
+ * table partitions which should help with the parallelism, but unlike the subpartitioner, the
+ * segment-partitioning scheme is not static and temporal rather than key-based. If data is skewed
+ * in time, this might result in uneven partitions and a need to further subdivide the partition
+ * space.
+ * <p>
+ * For the time being, we simply recommend that users configure the number of segments
+ * similarly to how they would configure the number of sub-partitions for a key-value store.
```
agavra, line 80: is this feasible? experimentally I needed the number of sub-partitions to be in the thousands for decent performance, but then you may end up scanning many segments to get the range you need (may be OK)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -0,0 +1,194 @@
...
+
+  public SegmentPartitioner(final long retentionPeriodMs, final long segmentIntervalMs) {
+    this.retentionPeriodMs = retentionPeriodMs;
+    this.segmentIntervalMs = segmentIntervalMs;
+  }
+
+  @Override
+  public SegmentPartition tablePartition(final int kafkaPartition, final Stamped<Bytes> key) {
+    // We don't subdivide the kafka partition any further since it's assumed there are
+    // sufficient segments to get good parallelism, but we could make both configurable
```
agavra, line 113: note that it isn't about parallelism and more about bloom filter efficiency, which are only done the the partition key

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SubPartitioner.java
```diff
@@ -16,96 +16,126 @@
...
+
+      final int requestedNumSubPartitions = config.getInt(STORAGE_DESIRED_NUM_PARTITION_CONFIG);
+
+      final int factor = (requestedNumSubPartitions == ResponsiveConfig.NO_SUBPARTITIONS)
+          ? 1 : (int) Math.ceil((double) requestedNumSubPartitions / numKafkaPartitions);
+      final int computedRemoteNum = factor * numKafkaPartitions;
 
-    return new SubPartitioner(factor, hasher);
+      if (actualRemoteCount.isPresent() && actualRemoteCount.getAsInt() != computedRemoteNum) {
+        throw new ConfigException(String.format("%s was configured to %d, which "
```
agavra, line 83: I think checkstyle will catch this, but these lines seem a bit long :) (reviewing side-by-side on GH)

agavra (commented):
mostly looks good to me though good parts of the logic I just trusted you've got them right, figured i'd give the ✅ after you run me through it

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
ableegoldman: Yeah I agree this is probably the weakest part of the abstraction. I was actually wondering if it might make sense to have the Mongo EOS implementation share the `RemoteLwtTable` interface as well, or if we should fully commit to this as C*/Lwt specific

Let's chat about this when we go over it

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
ableegoldman: Mainly I just had noticed that the epoch and offset were never read together/at the same time, so  the optimization  wasn't really doing anything 😞 

That said, I might end up wanting to go back to having a singular `MetadataRow` for the window store where we can persist the stream-time alongside the current offset. But I'll save that for later if/when it's needed

kafka-client/src/main/java/dev/responsive/kafka/internal/db/RemoteKVTable.java
```diff
@@ -19,20 +19,20 @@
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.state.KeyValueIterator;
 
-public interface RemoteKVTable<S> extends RemoteTable<Bytes, S> {
+public interface RemoteKVTable<P, S> extends RemoteTable<Bytes, P, S> {
```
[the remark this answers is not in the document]
ableegoldman: Was able to remove the `P` generic from the `RemoteTable` by just using a `?` type for the WriterFactory returned by `RemoteTable#init`. And because the partitioner was extracted out of the `KeyValue/WindowOperations` class, we actually don't have to care about the WriterFactory type and just pass it directly through to the CommitBuffer. So we don't even have to do any casting after all

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
ableegoldman: Changed the `RemoteLwtTable` to a standalone interface (that doesn't extend `RemoteTable` itself), and also renamed it to `TableMetadata` since it is currently being used to pass along the stream-time in addition to the epoch

I'm not crazy about the new name and think we might be able to clean this up further, but as discussed, I'm going to revisit this part during the followup PR to persist the stream-time

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
ableegoldman: Probably, yeah!

Note: we were already throwing an exception in the layer above, ie `PartitionedOperations`, so it's not like this could accidentally have been invoked. That said, I do think it makes sense to just implement the `PartitionedOperations#range` method and move the exception to the `CassandraKeyValueTable#range`, since that's where the fundamental limitation/missing functionality is. Same with all

(I did leave the UnsupportedOperationExceptions in PartitionedOperations for the reverseRange/reverseAll, however)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
ableegoldman: 100% 🚀

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: Added a TODO and filed a ticket for this. Nice catch

kafka-client/src/main/java/dev/responsive/kafka/internal/db/ColumnName.java
ableegoldman: I think this should have been moved to the `SegmentPartitioner` -- I'll do that now. Hopefully it'll be more clear after that, but basically yes, this is only ever used for metadata rows. The only place it's used in the codebase is for the method `SegmentPartitioner#metadataTablePartition`

kafka-client/src/main/java/dev/responsive/kafka/internal/db/FactWriterFactory.java
ableegoldman: Yes it very much needs to be renamed. I absolutely plan to change it, but wanted to leave that for a followup PR just to cut down on the total number of line changes here (which are already out of control). I'll do that immediately after this is merged

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriterFactory.java
ableegoldman: I think this is also a case of "it just needs to be renamed", but let me know if you had a more structural concern

(I filed a ticket to rename it so I won't forget)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/ResponsivePartitioner.java
ableegoldman: Thanks, yeah, that is a much better name

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
ableegoldman: Guess it depends on what kind of range scans people are actually doing?

Keep in mind, even though I did implement the basic single-key range scan in this PR, only the point lookup is actually used by any DSL operators. The range scan is used for stream-stream joins but we really just need a completely different table implementation for those (maybe not *completely* different, but certainly separate). I've thought a little bit about the stream-stream join implementation and I think in that case we could just do a little math based on the window size, since that determines the length of the range scan.

But I'm getting ahead of myself. If we just focus on the point lookups, we probably want a similarly large number of partitions, whether they are segments or subpartitions, right?

I guess some experimentation is needed here, and we can always recommend users to select a smaller number of segments if they plan to use range scans? This is an internal class so it's unlikely users will see these comments anyways, happy to remove the comment for now or just leave it up until we have something more concrete to say. Either way we definitely need to write a docs page for the window store once it's fully production ready

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SubPartitioner.java
ableegoldman: indeed they were 😛

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
ableegoldman: Fair enough...I'm actually going to remove this comment anyways since this is already noted in the main docs section above. It was more a throwaway note than a suggestion to actually implement sub-partitioning within segmented tables 🤷‍♀️

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriterFactory.java
ableegoldman: good catch

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: yep, thanks

PR #185, opened 2023-11-21
Persist stream-time for window stores

ableegoldman:
- The main change here is to start preserving the stream-time across restarts by persisting it in the metadata segment, along with the offset. 
- Also fixes some arithmetic/edge case errors in the SegmentPartitioner.
- With the above, we can re-enable the ResponsiveWindowStoreIntegrationTest

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
@@ -26,6 +26,7 @@ import static dev.responsive.kafka.internal.db.ColumnName.OFFSET;
 import static dev.responsive.kafka.internal.db.ColumnName.PARTITION_KEY;
 import static dev.responsive.kafka.internal.db.ColumnName.ROW_TYPE;
 import static dev.responsive.kafka.internal.db.ColumnName.SEGMENT_ID;
+import static dev.responsive.kafka.internal.db.ColumnName.STREAM_TIME;
 import static dev.responsive.kafka.internal.db.ColumnName.WINDOW_START;
 import static dev.responsive.kafka.internal.db.RowType.DATA_ROW;
 import static dev.responsive.kafka.internal.db.RowType.METADATA_ROW;
@@ -59,6 +60,7 @@ import java.util.concurrent.TimeoutException;
 import javax.annotation.CheckReturnValue;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.errors.TaskMigratedException;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -90,6 +92,8 @@ public class CassandraWindowedTable implements
   private final PreparedStatement backFetchAll;
   private final PreparedStatement fetchOffset;
   private final PreparedStatement setOffset;
+  private final PreparedStatement fetchStreamTime;
+  private final PreparedStatement setStreamTime;
   private final PreparedStatement fetchEpoch;
   private final PreparedStatement reserveEpoch;
   private final PreparedStatement ensureEpoch;
@@ -135,18 +139,16 @@ public class CassandraWindowedTable implements
     client.execute(createTable.build());
     client.awaitTable(name).await(Duration.ofSeconds(60));
 
-    // TODO: consolidate the #init method into the constructor and create the
-    //  LwtWriterFactory before preparing the statements so we can just plug
-    //  in the epoch directly without having to bind it later/on each request
     final var initSegment = client.prepare(
-        QueryBuilder.insertInto(name)
-            .value(PARTITION_KEY.column(), bindMarker(PARTITION_KEY.bind()))
-            .value(SEGMENT_ID.column(), bindMarker(SEGMENT_ID.bind()))
-            .value(ROW_TYPE.column(), METADATA_ROW.literal())
-            .value(DATA_KEY.column(), DATA_KEY.literal(ColumnName.METADATA_KEY))
-            .value(WINDOW_START.column(), WINDOW_START.literal(METADATA_TS))
-            .value(EPOCH.column(), bindMarker(EPOCH.bind()))
-            .ifNotExists()
+        QueryBuilder
+            .update(name)
+            .setColumn(EPOCH.column(), bindMarker(EPOCH.bind()))
+            .where(PARTITION_KEY.relation().isEqualTo(bindMarker(PARTITION_KEY.bind())))
+            .where(SEGMENT_ID.relation().isEqualTo(bindMarker(SEGMENT_ID.bind())))
+            .where(ROW_TYPE.relation().isEqualTo(METADATA_ROW.literal()))
+            .where(DATA_KEY.relation().isEqualTo(DATA_KEY.literal(METADATA_KEY)))
+            .where(WINDOW_START.relation().isEqualTo(WINDOW_START.literal(METADATA_TS)))
+            .ifColumn(EPOCH.column()).isLessThan(bindMarker(EPOCH.bind()))
             .build()
     );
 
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -98,9 +103,33 @@ public class CassandraWindowedTable implements
   // as these entities have different views of the current time and should not be unified.
   // (Specifically, this table will always lag the view of stream-time that is shared by the
   // ResponsiveWindowStore and CommitBuffer due to buffering/batching of writes)
-  private long lastFlushStreamTime = UNINITIALIZED_STREAM_TIME;
-  private long pendingFlushStreamTime = UNINITIALIZED_STREAM_TIME;
-  private SegmentRoll activeRoll;
+  private final Map<Integer, PendingFlushInfo> kafkaPartitionToPendingFlushInfo = new HashMap<>();
+
+  // TODO: move this into the LWTWriter/Factory to keep this class stateless
```
ableegoldman, line 108: I know this is super ugly, I intend to do this TODO right away (though probably in a separate followup PR)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriterFactory.java
```diff
@@ -85,6 +85,12 @@ public RemoteWriteResult<P> setOffset(final long offset) {
     builder.addStatement(fencingStatement(tablePartition));
     builder.addStatement(table.setOffset(kafkaPartition, offset));
 
+    // TODO(sophie): clean up this hack, perhaps by combining the offset and stream-time into
```
ableegoldman, line 88: I also intend to do this TODO right away as well, though again, probably in a followup PR

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -428,47 +488,112 @@ public String name() {
...
+    // regular data partitions/segments and never expired
+    // therefore we initialize from the metadata partition and then broadcast the epoch to
+    // all the other partitions containing data for active segments
+    for (final SegmentPartition tablePartition : partitioner.activeSegments(kafkaPartition, streamTime)) {
+      final var reserveSegmentEpoch = client.execute(reserveEpoch(tablePartition, epoch));
+
+      if (!reserveSegmentEpoch.wasApplied()) {
+        handleEpochFencing(kafkaPartition, tablePartition, epoch);
+      } else {
+        LOG.info("SOPHIE: reserved epoch {} for kafka partition {} and segment {}",
```
agavra, line 534: I think it could be helpful to have this log - perhaps as a summary at the end of all the epochs that it reserved? (certainly we should remove `SOPHIE:` though 😉 )

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -287,15 +314,41 @@ public static CassandraWindowedTable create(
...
+    final var setStreamTime = client.prepare(
+        QueryBuilder
+            .update(name)
+            .setColumn(STREAM_TIME.column(), bindMarker(STREAM_TIME.bind()))
+            .where(PARTITION_KEY.relation().isEqualTo(bindMarker(PARTITION_KEY.bind())))
+            .where(SEGMENT_ID.relation().isEqualTo(bindMarker(SEGMENT_ID.bind())))
+            .where(ROW_TYPE.relation().isEqualTo(METADATA_ROW.literal()))
+            .where(DATA_KEY.relation().isEqualTo(DATA_KEY.literal(METADATA_KEY)))
+            .where(WINDOW_START.relation().isEqualTo(WINDOW_START.literal(METADATA_TS)))
+            .ifColumn(EPOCH.column()).isEqualTo(bindMarker(EPOCH.bind()))
```
agavra, line 350: does it make sense here to ensure that the existing stream time <= the desired setStreamTime or is that enforced by the application in all edge cases?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -428,47 +488,112 @@ public String name() {
...
-        initSegment(new SegmentPartition(kafkaPartition, segmentId), epoch);
+    pendingFlush.initSegmentRoll(pendingRoll);
+    for (final long segmentId : pendingRoll.segmentsToCreate) {
+      final SegmentPartition segment = new SegmentPartition(kafkaPartition, segmentId);
+      final var createSegment = createSegment(segment, epoch);
+
+      // If the segment creation failed because the table partition already exists, attempt to
+      // update the epoch in case we are fencing an older writer -- if that fails it means we're
+      // the ones being fenced
+      // TODO: what if the segment creation failed for a reason besides already existing?
```
agavra, line 589: then it will throw an error instead of returning `!wasApplied`

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriterFactory.java
```diff
@@ -85,6 +85,12 @@ public RemoteWriteResult<P> setOffset(final long offset) {
...
+    //  a single metadata row update
+    if (table instanceof CassandraWindowedTable) {
+      builder.addStatement(((CassandraWindowedTable) table).setStreamTime(kafkaPartition, epoch));
```
agavra, line 91: since this is being issued as part of the batch, we probably don't need the:
```
ifColumn(EPOCH.column()).isEqualTo(bindMarker(EPOCH.bind()))
```
part of the query. the fencing statement will do that for us (i'm not sure if there's any problem with checking that twice, probably not)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -126,6 +129,8 @@ public SegmentPartitioner(
     this.retentionPeriodMs = retentionPeriodMs;
     this.segmentIntervalMs = segmentIntervalMs;
     this.windowSizeMs = windowSizeMs;
+    LOG.info("Created segment partitioner with retentionPeriod={}ms, segmentInterval={}ms,"
```
agavra, line 132: can we assert that all three of these values are strictly greater than 0? that way we'll avoid all division by zero (we should update `StoreUtil#computeSegmentInterval` to make sure segmentInterval can't be 0)

agavra (approved):
LGTM - some suggestions inline

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: hahahah whoops -- guess I missed removing all the logging from my debug sessions. Fixed

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: I was wondering if adding that condition would have a performance impact since we'd have to read an extra column.

As long as the epoch is checked/we're guaranteed single writer access, I don't see how the stream-time could go backwards, so I don't think it's worth verifying

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: oh, that's good to know 😄 I'll remove the TODO

kafka-client/src/main/java/dev/responsive/kafka/internal/db/LwtWriterFactory.java
ableegoldman: ah, right, good point. I'll just remove it for now

But actually this is something I was thinking about the other day -- my understanding is that functionally, there is no difference between doing the fencing with a separate epoch-checking statement vs adding an epoch-check to each individual insert/delete. But is there a performance difference? 

I'm assuming yes, these conditions have a cost, so it's best to just do the epoch-read once per batch write instead of per-update? I assume there's no way Cassandra or Scylla is smart enough to optimize the extra epoch checks away..

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
ableegoldman: Sure thing. By the way, I lowered the default num_segments by a lot (a few weeks ago, not in this PR) so I _think_ even without this fix, you should be able to run the window QTT tests without hitting the div by zero issue. Of course it doesn't hurt to just wait for the actual fix either

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
ableegoldman: Ok actually I know I said I would do this immediately but I'm actually going to start on the Mongo implementation first, so I can see how it makes sense to fit together with that since we won't be using the LwtWriter for Mongo but will presumably still need the same time-tracking stuff

I'll do it as part of, or else immediately after, the Mongo work

PR #187, opened 2023-11-23
Remove generic <K> from Stamped and clean up Comparator logic

ableegoldman:
Main change here is to remove the generic <K> from the Stamped class and hard-code it to Bytes.

The rest of this PR is all the clean up that's enabled from this one change, mainly around removing generics from window store related iterators and from making the actual key generic <K> extend Comparable<K> instead of using the KeySpec class as a Comparator


agavra (commented):
LGTM, posthumous +1

PR #189, opened 2023-11-27
support metrics-only mode in ResponsiveKafkaStreams

agavra:
the new config (`responsive.compatibility.mode = METRICS_ONLY`) effectively disables the following:
- config validation
- responsive restore consumer
- responsive global consumer
- connecting to responsive storage

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
@@ -37,6 +37,11 @@ public class ResponsiveConfig extends AbstractConfig {
 
   // ------------------ connection configurations -----------------------------
 
+  public static final String COMPATIBILITY_MODE_CONFIG = "responsive.compatibility.mode";
+  private static final String COMPATIBILITY_MODE_DOC = "This configuration enables running Responsive "
+      + "in compatibility mode, disabling certain features.";
+  private static final CompatibilityMode COMPATIBILITY_MODE_DEFAULT = CompatibilityMode.FULL;
+
   public static final String STORAGE_HOSTNAME_CONFIG = "responsive.storage.hostname";
   private static final String STORAGE_HOSTNAME_DOC = "The hostname of the storage server.";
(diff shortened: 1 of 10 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -37,6 +37,11 @@ public class ResponsiveConfig extends AbstractConfig {
 
   // ------------------ connection configurations -----------------------------
 
+  public static final String COMPATIBILITY_MODE_CONFIG = "responsive.compatibility.mode";
+  private static final String COMPATIBILITY_MODE_DOC = "This configuration enables running Responsive "
+      + "in compatibility mode, disabling certain features.";
```
rodesai, line 42: compatibility is a strange word for this. Maybe "instrumentation"?

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/EndOffsetsPoller.java
```diff
@@ -63,8 +64,9 @@ public class EndOffsetsPoller {
 
   public EndOffsetsPoller(
       final Map<String, ?> configs,
-      final ResponsiveMetrics metrics
-  ) {
+      final ResponsiveMetrics metrics,
+      final KafkaClientSupplier clientSupplier
```
rodesai, line 68: nit: can you just have this accept an admin client? I feel like KafkaClientSupplier should only be called by streams.

rodesai (approved):
Mostly LGTM. Couple minor suggestions inline.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
agavra: the idea was whether it should be compatible with existing stream deployments or not - basically I wanted to communicate that if you run it in "compatibility mode" you will not be able to use responsive stores.

PR #190, opened 2023-11-27
fix stamped record deserializer

agavra:
for some bizarre reason I had implemented this originally to read the ts from the key not the value, where it is actually stored. no idea how this ever passed any tests -- we should follow this up with comprehensive tests but I'm making this change to unblock the build.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/StampedKeySpec.java
@@ -16,11 +16,8 @@
 
 package dev.responsive.kafka.internal.db;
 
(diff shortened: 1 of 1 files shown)


ableegoldman:
Super weird that the test passed in this build as well...something must be wrong with the PR build action. You definitely had the right logic the first time, ie the window start timestamp is in the key, not the value. Look at the WindowKeySchema in Streams, we extract the timestamp from the key for Rocksdb window stores as well. This is just how they're encoded in the changelog records, which have keys of the Windowed type, ie, keys with the record_key + timestamp. The timestamp that's in the value is not the window start time, it's from a relatively newer feature that corrected the semantics in some cases where you needed access to the original record timestamps, whereas the timestamp in the windowed key has been around since the first window store because we've always needed to store it

PR #191, opened 2023-11-28
Support for agent-free metrics reporting

agavra:
I tested this with a local controller setup directly reporting to DD: https://us5.datadoghq.com/s/7d2b7ec4-e391-11ed-b891-da7ad0900005/iua-jdy-8wc

In a follow-up PR I'll modify `kafka-client-examples/simple-example` to run with agent-free metrics reporting.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OtelMetricsService.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OtelMetricsService.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.internal.metrics;
+
+import dev.responsive.kafka.api.config.ResponsiveConfig;
+import dev.responsive.kafka.internal.config.ConfigUtils;
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import io.opentelemetry.exporter.otlp.metrics.OtlpGrpcMetricExporter;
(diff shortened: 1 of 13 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveMetrics.java
```diff
@@ -48,10 +48,12 @@ public class ResponsiveMetrics implements Closeable {
   private static final Pattern GLOBAL_THREAD_REGEX = Pattern.compile(".*-(GlobalStreamThread+)");
 
   private OrderedTagsSupplier orderedTagsSupplier;
+  private final OtelMetricsService otelService;
```
rodesai, line 51: can we keep this service decoupled from the metrics collection stuff? Ideally we won't bake in the fact that we're using any specific reporter. One suggestion is to have some way to listen on `ResponsiveMetrics` being closed and register the otel reporter as a listener. And then move the otel stuff to its own package.

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OtelMetricsService.java
```diff
@@ -0,0 +1,123 @@
...
+    onClose.run();
+  }
+
+  private static void buildFromUserRules(MetricConfiguration conf) {
+    RuleParser parserInstance = RuleParser.get();
+    final ClassLoader loader = OtelMetricsService.class.getClassLoader();
+
+    // TODO(agavra): instead of including otel-jmx.config.yaml as a resource we should
+    // fetch it from the Responsive controller on start-up
+    try (final var is = loader.getResourceAsStream("otel-jmx.config.yaml")) {
```
rodesai, line 109: this is a huge improvement 🥳

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/OtelMetricsService.java
```diff
@@ -0,0 +1,123 @@
...
+      parserInstance.addMetricDefsTo(conf, is, "responsive builtin config");
+    } catch (Exception e) {
+      // yaml parsing errors are caught and logged inside addMetricDefsTo
+      // only file access related exceptions are expected here
+      JmxMetricInsight.getLogger().warning(e.toString());
```
rodesai, line 114: I think this should be a fatal error that we should throw the exception up for. Now that we're getting the config from a resource, this indicates something is seriously wrong with the jar we're shipping.

kafka-client/src/main/resources/otel-jmx.config.yaml
```diff
@@ -0,0 +1,334 @@
+---
```
rodesai, line 1: this is the same as https://github.com/responsivedev/responsive-pub/blob/main/etc/otel-jmx.config.yaml right? Can you remove that file now?

rodesai (commented):
Awesome! Left a few thoughts inlline

kafka-client/src/main/resources/otel-jmx.config.yaml
agavra: Yes, I will remove it once I migrate all the examples to use the builtin reporter

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/ResponsiveMetrics.java
agavra: I'll wrap it in a `MetricsExporterService` and move no-op to be it's own implementation of that. That way we don't add yet another series of callbacks, they're hard to keep track of.

PR #192, opened 2023-12-01
Fix window start timestamp extraction

ableegoldman:
The original deserialization logic in the StampedKeySpec had the right idea, but wasn't accounting for the extra 4 bytes at the end for the `seqnum`, which it turns out is always encoded in the windowed key bytes that are sent to the changelog even if the store does not belong to a stream-stream join (ie regardless of `retainDuplicates` for which the seqnum is actually used)

Admittedly it does feel a bit confusing that there are so many different timestamps in the mix: there is the one encoded in the ValueAndTimestamp for timestamped stores, the ProcessorContext's timestamp, the changelog record's timestamp (ie the one in the metadata), the timestamp used by the CommitBuffer, and then of course, the timestamp of the window start.

To try and differentiate the window start times from these other timestamps, I renamed the `Stamped` class to `WindowedKey` (and `StampedKeySpec` to `WindowedKeySpec`). Most of the LOC come from this change -- in fact the only change in logic is to the few lines in StampedKeySpec (which is now WindowedKeySpec) where we extract the timestamp and key


kafka-client/src/main/java/dev/responsive/kafka/internal/db/WindowedKeySpec.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public int sizeInBytes(final WindowedKey key) {
+    return key.key.get().length + Long.BYTES;
+  }
+
+  @Override
+  public WindowedKey keyFromRecord(final ConsumerRecord<byte[], byte[]> record) {
+    final byte[] key = record.key();
+
+    // the WindowKeySchema in Streams always encodes the changelog record keys by
```
agavra, line 44: since this has bitten us before, what do you think about using `WindowKeySchema#fromStoreBytesKey` directly and then just discarding things we don't use (like the window size)? 

It also looks like it's encoded different based on whether its from a changelog topic or a source topic?
```
        // toStoreKeyBinary was used to serialize the data.
        if (this.isChangelogTopic) {
            return WindowKeySchema.fromStoreKey(data, windowSize, inner, topic);
        }

        // toBinary was used to serialize the data
        return WindowKeySchema.from(data, windowSize, inner, topic);
```

`seqNum` is ignored in the latter, which is a bit confusing?

We should make sure to have a test where we source directly from a windowed output topic w/o a changelog

agavra (commented):
Can we add meaningful tests to this PR? Part of what made debugging this really tough was that the tests failures were super hard to track down.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WindowedKeySpec.java
ableegoldman: I'll take a look at the WindowKeySchema to make sure I'm not missing anything, but we definitely do not support windowed source tables or WindowStores with a source-topic changelog. See [KIP-300](https://cwiki.apache.org/confluence/display/KAFKA/KIP-300%3A+Add+Windowed+KTable+API+in+StreamsBuilder) for context (note that this KIP was abandoned so the changes proposed there do not exist)

ableegoldman: Oh wait, unfortunately the WindowKeySchema only converts bytes to a full `Windowed<Bytes>` key, which means it does an extra computation to compute the end time. It wouldn't be the end of the world to just pass in a dummy window size and throw out the end time, but it seems silly to do this extra work when we can just extract only the needed bits. I know it's just a bit of addition but this is very much on the hot path, so may as well cut it out

However, I will add a unit test that verifies the WindowKeySchema format encodes and decodes the way we expect it to, just to be safe

PR #206, opened 2023-12-18
dont bubble up offset polling errors

rodesai:
dont bubble up offset polling errors. If we do, then the poller makes no more attempts to poll. Instead, just log the error and the poller will retry on the next iteration. If its still failing after 3 iterations, it reinitializes the admin client.


agavra (approved):
LGTM, is there any concerns of doing too many admin initializations? I guess this is limited to once every 90s so it can't be too bad

rodesai:
yeah that's what I was thinking. It's very infrequent even in the worst case.

PR #207, opened 2023-12-19
Fix edge case for #range query and add StoreQueryIntegrationTest for range scans

ableegoldman:


kafka-client/src/test/java/dev/responsive/kafka/integration/StoreQueryIntegrationTest.java
```diff
@@ -0,0 +1,296 @@
...
+  private String inputTopic() {
+    return name + "." + INPUT_TOPIC;
+  }
+
+  private String outputTopic() {
+    return name + "." + OUTPUT_TOPIC;
+  }
+
+  @Test
+  public void testKeyValueStoreAllQuery() throws Exception {
```
agavra, line 117: I have strong feelings around unit tests 😆  

I think it makes a big difference when we name tests `shouldX` instead of `testX` - `testX` just indicates that the code for `X` is being run. `shouldX` indicates what the test is testing. I found that naming them properly also tends to have more targeted and meaningful tests.

For example, what are we actually testing in this test? Perhaps `shouldReturnAllStoreDataOnAllQuery` is a better name, and then if the value was a string instead of a number the test could be more easily verified (e.g. the last value is `("E", "ABCDE")`. Then the range query one would be `shouldReturnRangeInclusiveOnStoreRangeQuery` and the last one would be `("E", "BCD")`. 

In the same spirit of only testing what you claim to be testing (writing minimal, targeted, tests is generally good practice), we could further improve this test by only asserting the last item instead of all the intermediate items. 

the tl;dr is that I had to read the entire file to make sense of these tests, but with the two improvements above I could probably make sense of reading just the individual test without the guts of the streams app to understand what it was testing.

agavra (approved):
Thanks @ableegoldman - LGTM. One comment about tests.

PR #209, opened 2023-12-26
Use kafkaPartition for ObjectId in MongoKVTable

ableegoldman:
Minor improvement to clear up a somewhat confusing tactic of saving a map of MetadataDocs per partition when we only need to save the epoch per partition instead.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -35,6 +35,7 @@ import com.mongodb.client.model.Updates;
 import com.mongodb.client.model.WriteModel;
 import dev.responsive.kafka.internal.db.RemoteKVTable;
 import dev.responsive.kafka.internal.db.WriterFactory;
+import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.Date;
 import java.util.concurrent.ConcurrentHashMap;
@@ -46,6 +47,7 @@ import org.bson.Document;
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.codecs.pojo.PojoCodecProvider;
+import org.bson.types.ObjectId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -61,11 +63,7 @@ public class MongoKVTable implements RemoteKVTable<WriteModel<Document>> {
   private final MongoCollection<Document> genericDocs;
   private final MongoCollection<Document> genericMetadata;
 
-  // this map contains the initialized value of the metadata document,
-  // for which the epoch and object ID will never change. it is likely
-  // that after a few writes to MongoDB the cached MetadataDoc.offset value
-  // will be out of date, so it should not be used beyond initialization
-  private final ConcurrentMap<Integer, MetadataDoc> metadataRows = new ConcurrentHashMap<>();
+  private final ConcurrentMap<Integer, Long> kafkaPartitionToEpoch = new ConcurrentHashMap<>();
 
   public MongoKVTable(final MongoClient client, final String name) {
     this.name = name;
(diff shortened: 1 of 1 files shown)


ableegoldman:
@agavra I'm going to go ahead and merge this so I can include it with some other cleanup refactoring I've been playing around with, but please still take a look when you're back in case I'm somehow messing something up here

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -210,4 +207,11 @@ public long approximateNumEntries(final int kafkaPartition) {
     return 0;
   }
 
+  private ObjectId idForKafkaPartition(final int kafkaPartition) {
```
agavra, line 210: I'm pretty sure you can change the `_id` field type if you want, in this case we should just make it an `int` instead of `ObjectId`.

agavra (commented):
LGTM

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -97,28 +95,27 @@ public String name() {
...
+            .returnDocument(ReturnDocument.AFTER)
+    );
+
+    if (metaDoc == null) {
+      throw new IllegalStateException("Uninitialized metadata for partition " + kafkaPartition);
+    }
+
+    LOG.info("Retrieved initial metadata {}", metaDoc);
+
+    kafkaPartitionToEpoch.put(kafkaPartition, metaDoc.epoch);
```
agavra, line 118: I think changing this from computeIfAbsent to put fixed a bug when rebalances that re-assigned the node the same pod would not generate a new epoch, and therefore remain fenced.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -210,4 +207,11 @@ public long approximateNumEntries(final int kafkaPartition) {
     return 0;
   }
 
+  private ObjectId idForKafkaPartition(final int kafkaPartition) {
+    // allocate 16 bytes total since ObjectId needs 12 for internal stuff
+    byte[] bytes = new byte[16];
+    final ByteBuffer buffer = ByteBuffer.wrap(bytes).putInt(kafkaPartition);
```
agavra, line 213: looks like there's some kind of bug here, it always creates the following ID when I'm running it in my test:
```
{"_id":{"$oid":"000000000000000000000000"},"partition":{"$numberInt":"2"},"epoch":{"$numberInt":"1"},"offset":{"$numberLong":"-1"}}
```
I'm going to go ahead and patch this since it's blocking some of my experiments.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
agavra: https://github.com/responsivedev/responsive-pub/pull/210

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -97,28 +95,27 @@ public String name() {
...
-        throw new IllegalStateException("Uninitialized metadata for partition " + kafkaPartition);
-      }
-
-      LOG.info("Retrieved initial metadata {}", metaDoc);
-
-      return metaDoc;
-    });
 
+    final MetadataDoc metaDoc = metadata.findOneAndUpdate(
+        Filters.eq(MetadataDoc.PARTITION, kafkaPartition),
```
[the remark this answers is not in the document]
agavra: https://github.com/responsivedev/responsive-pub/pull/210

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
agavra: https://github.com/responsivedev/responsive-pub/pull/210

PR #210, opened 2024-01-04
fix mongo metadata table _id field

agavra:
for some reason, the previous code was generating `_id` clashes, this fixes it by just setting the `_id` to be the partition.

note (cc @ableegoldman) this is a breaking change that requires manual fix if you have anything that's deployed using Mongo reach out to me and we can migrate it.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MetadataDoc.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MetadataDoc.java
@@ -17,8 +17,6 @@
 package dev.responsive.kafka.internal.db.mongo;
 
 import java.util.Objects;
-import org.bson.codecs.pojo.annotations.BsonId;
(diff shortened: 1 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MetadataDoc.java
```diff
@@ -27,20 +25,19 @@ public class MetadataDoc {
   public static final String OFFSET = "offset";
   public static final String EPOCH = "epoch";
 
-  @BsonId
-  ObjectId id;
+  int id;
   int partition;
```
ableegoldman, line 29: @agavra was just wondering, since we're now using the kafka partition as the `id` should we get rid of the redundant `partition` field?  Can you think of any reason to keep it around?
eg: Even if we change/evolve the `id` at some point in the future, that would be a breaking change, so we can always add back the `partition` field at that time.

PR #212, opened 2024-01-11
Add MongoWindowedTable and refactor WriterFactory approach to flush management

ableegoldman:


kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactFlushManager.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override
+  public RemoteWriteResult<Integer> preFlush() {
+    return RemoteWriteResult.success(null);
+  }
+
+  @Override
+  public RemoteWriteResult<Integer> postFlush(final long consumedOffset) {
+    final int metadataPartition = partitioner.metadataTablePartition(kafkaPartition);
+    final var result = client.execute(table.setOffset(kafkaPartition, consumedOffset));
```
agavra, line 70: just adding some inline comments to make sure this is safe (from first principles): if a batch write succeeds but we don't make it to `postFlush` the new writer will take over, recover from the old offsets, and rewrite them into the store during the restore phase.

let's say that this thread then wakes back up and successfully issues this offset flush - what happens? It'll overwrite the offset then fail b/c it was fenced by Kafka. Let's assume that for whatever reason the current writer (which has progressed past this point) then fails before it can overwrite this now zombie offset.

A new and third writer comes online and recovers from this offset, which may be a problem since it'll overwrite already processed data... 

I think this may be a bug in the current implementation as well (not introduced by this PR). We should just be safe and change the `setOffset` command to only set if the offset it is trying to write is greater than or equal to the one that's there.

Do you have bandwidth to address that as a follow up to this PR? (we should confirm my theory and that I'm not missing something, I thought that I had thought this situation through when designing fact tables but I may have missed it)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKVFlushManager.java
```diff
@@ -0,0 +1,93 @@
...
+  public RemoteWriteResult<Integer> preFlush() {
+    return RemoteWriteResult.success(null);
+  }
+
+  @Override
+  public RemoteWriteResult<Integer> postFlush(final long consumedOffset) {
+    final int tablePartition = partitioner.metadataTablePartition(kafkaPartition);
+
+    final BatchStatementBuilder builder = new BatchStatementBuilder(BatchType.UNLOGGED);
+    builder.addStatement(table.ensureEpoch(tablePartition, epoch));
```
agavra, line 80: (no action, just notes) if the situation described in my previous comment happens and this comes back online, then this will fail - which is an acceptable resolution - because of epoch fencing

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -99,37 +100,8 @@ public class CassandraWindowedTable implements
...
-    void initSegmentRoll(final SegmentRoll pendingSegmentRoll) {
-      this.segmentRoll = pendingSegmentRoll;
-    }
-
-    void finalizeFlush() {
-      segmentRoll = null;
-      lastFlushStreamTime = pendingFlushStreamTime;
-    }
-  }
+  private final ConcurrentMap<Integer, SegmentBatch> kafkaPartitionToSegmentMetadata =
```
agavra, line 103: ideally this class is "stateless" in the sense that it doesn't carry information per-partition. what do you think about moving this as well as all the precommit/postcommit code into the FlushManager (now that we have one per table per partition), which is already per partition IIUC? (basically you don't have a map here)

the benefit is that we don't recycle tables during a rebalance, so it's nice to know that if a rebalance happens twice and we're reassigned a partition we were assigned a long time ago we just create a new thing and never accidentally use corrupted old data from an old assignment

kafka-client/src/main/java/dev/responsive/kafka/internal/db/MongoKVFlushManager.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public RemoteWriteResult<Integer> preFlush() {
+    return RemoteWriteResult.success(kafkaPartition);
+  }
+
+  @Override
+  public RemoteWriteResult<Integer> postFlush(final long consumedOffset) {
+    try {
+      // TODO: should we check result.wasAcknowledged()/use a write concern?
+      metadataDocs.bulkWrite(List.of(table.setOffset(kafkaPartition, consumedOffset)));
```
agavra, line 83: if we have access directly to the table we no longer need to do a bulk write for this (previously we did a bluk write to keep the API the same as for normal writes)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -466,27 +468,19 @@ private void doFlush(final long consumedOffset, final int batchSize) {
         buffer.getReader().size(),
         batchSize,
         consumedOffset,
-        writerFactory
+             writeBatcher
```
agavra, line 471: nit: indentation - not sure if checkstyle catches this or not

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -466,27 +468,19 @@ private void doFlush(final long consumedOffset, final int batchSize) {
...
-    final var flushResult = writerFactory.commitPendingFlush(pendingFlush, consumedOffset);
-    // TODO: we could/probably should move the result checking and handling inside the
-    //  #commitPendingFlush method, since we have to go back to the WriterFactory for the
-    //  specific error message anyways. It might also be nice to extract the exception
-    //  supplier so that different tables and/or writers can throw more specific errors
     if (!flushResult.wasApplied()) {
-      throwFlushException(flushResult, consumedOffset);
+      flushErrorsSensor.record();
+
+      final String errorMsg = writeBatcher.failedFlushMsg(flushResult, consumedOffset);
```
agavra, line 480: (suggestion - personal preference) I personally prefer creating error messages inline so that it's easy to find the place in the code based on the error message with a ctrl+F, I'd rather do that than abstract it away even if there's some code duplication (same with below)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/GlobalOperations.java
```diff
@@ -104,7 +104,8 @@ private void put(
       final long timestamp
   ) {
     client.execute(table.insert(IGNORED_PARTITION, key, value, timestamp));
-    client.execute(table.setOffset(partition, offset));
+    // TODO(sophie) FIXME!!
+    // client.execute(table.setOffset(partition, offset));
```
agavra, line 108: as discussed, let's just hardcode C* fact table for now

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatch.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+import org.apache.kafka.common.utils.LogContext;
+import org.slf4j.Logger;
+
+/**
+ * A class for holding and coordinating the pending writes and other metadata involved
+ * in a flush to the remote table
+ */
+public class WriteBatch<K extends Comparable<K>, P> {
```
agavra, line 38: Offline I wasn't sure what the point of this class is if we have `CommitBuffer` but it makes sense to me now - `CommitBuffer` buffers writes and then constructs a `WriteBatch` (which is effectively immutable after creation). If we really want to make that clear we could rename it `ReadyBatch` or something like that to indicate that it's something ready to go instead of something dynamic (🤮 ok, well that name sucks, but you get the point and hoepfully where my confusion came from)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatcher.java
```diff
@@ -0,0 +1,106 @@
...
+
+package dev.responsive.kafka.internal.db;
+
+import dev.responsive.kafka.internal.stores.RemoteWriteResult;
+import dev.responsive.kafka.internal.utils.Result;
+import java.util.Collection;
+import org.apache.kafka.common.utils.LogContext;
+import org.slf4j.Logger;
+
+public class WriteBatcher<K extends Comparable<K>, P> {
```
agavra, line 27: As discussed offline, I think this makes sense to just be in the `CommitBuffer` - otherwise we can rename it something like `BatchFlusher` since it doesn't actually batch writes, it just flushes writes that have already been batched 😵‍💫 I'm tripping over words haha

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatcher.java
```diff
@@ -0,0 +1,106 @@
...
+    final WriteBatch<K, P> writeBatch = new WriteBatch<>(
+        flushManager::createWriter,
+        keySpec,
+        kafkaPartition,
+        flushManager.partitioner(),
+        bufferedWrites
+    );
+    log.info("SOPHIE: created write batch, about to prepare flush");
+
+    numTablePartitionsInBatch = writeBatch.numTablePartitionsInBatch();
```
agavra, line 59: 😬 this seems like unnecessary mutable state, let's just return an object that contains this in the result if we want

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MetadataDoc.java
```diff
@@ -18,6 +18,7 @@
 
 import java.util.Objects;
 
+// TODO (sophie): rename to KVMetadataDoc and eliminate redundant partition field
```
agavra, line 21: I kind-of liked having `partition` be the name of the partition field, since `_id` is pretty opaque. I think there's a way to configure a table not to use `_id` and use a custom field instead, but it's hardly any extra data overhead I figured the duplication is fine

agavra (commented):
I took a look at the structure of the code and it LGTM, some comments inline. I have not yet looked at the Mongo windowed specific stuff, I'll look at that at some point tonight. If this is blocking you though feel free to merge and continue your testing.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,489 @@
...
+    // the collection name and will indicate whether it's part of the active set or not, perhaps
+    // using a scheduled executor to avoid blocking processing
+    // In fact we may want to move all the segment expiration to a background process since there's
+    // no async way to drop a collection but the stream thread doesn't actually need to wait for it
+    void expireSegment(final SegmentPartition segmentToDelete) {
+      LOG.info("{}[{}] Expiring segment id {}",
+               name, segmentToDelete.tablePartition, segmentToDelete.segmentId);
+
+      final var expiredDocs = segmentWindows.get(segmentToDelete);
+      expiredDocs.drop();
```
agavra, line 122: what happens if this is called on a collection that doesn't exist? (it's a possible race condition if this thread is a zombie)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,489 @@
...
+    if (activeSegments.isEmpty()) {
+      LOG.info("{}[{}] No active segments for initial streamTime {}",
+               name, kafkaPartition, metaDoc.streamTime);
+    } else {
+      for (final SegmentPartition segmentToCreate : activeSegments) {
+        // Most likely if we have active segments upon initialization, the corresponding
+        // collections and windowStart indices already exist, in which case #getCollection and
+        // #createIndex should be quick and won't rebuild either structure if one already exists
+        // TODO: optimize by using listCollectionNames to skip redundant collection/index creation
+        //  since we should almost always have an index for each collection for all active segments
```
agavra, line 187: 🤷 IMO we shouldn't do this unless we determine there's a reason to optimize it - generally my opinion on optimization TODOs is not to add them into the code since we should only do them if experimentally it proves valuable (unless it's also a code or architectural improvement). Not a hill I'll die on though

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,489 @@
...
+
+    final var segmentBatch = partitionSegments.segmentBatch;
+    final SegmentRoll pendingRoll = partitioner.rolledSegments(
+        name, segmentBatch.flushedStreamTime, segmentBatch.batchStreamTime
+    );
+
+    segmentBatch.prepareRoll(pendingRoll);
+
+    for (final long segmentId : pendingRoll.segmentsToCreate) {
+      partitionSegments.createSegment(new SegmentPartition(kafkaPartition, segmentId));
```
agavra, line 237: this is the only side effect in `preCommit`, right? should be safe if there's a race condition so not a concern

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,489 @@
...
+   *         the windowStart timestamp
+   */
+  @Override
+  public WriteModel<WindowDoc> delete(
+      final int kafkaPartition,
+      final WindowedKey windowedKey
+  ) {
+    final var partitionSegments = kafkaPartitionToSegments.get(kafkaPartition);
+    partitionSegments.segmentBatch.updateStreamTime(windowedKey.windowStartMs);
+    throw new UnsupportedOperationException("Deletes not yet supported for MongoDB window stores");
```
agavra, line 350: is this necessary to implement for the DSL operators?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public static final String ID = "_id";
+  public static final String VALUE = "value";
+  public static final String EPOCH = "epoch";
+  public static final String WINDOW_START_TS = "windowStartTs";
+
+  @BsonId
+  byte[] id;
+  byte[] value;
+  long epoch;
```
agavra, line 35: totally unrelated to this PR, but what do you think about changing epoch to be an int instead of a long? this currently has 8 bytes overhead per row, and MAX_INT gives us 2billion rebalances before running out of room. If we have a rebalance every minute, it would take us 4000 years to hit max int (`2000000000/(60*24*365)`)

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveWindowStoreIntegrationTest.java
```diff
@@ -70,22 +72,28 @@
...
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-@ExtendWith(ResponsiveExtension.class)
 public class ResponsiveWindowStoreIntegrationTest {
 
+  // TODO: parametrize so we don't break the Cassandra implementation without knowing
```
agavra, line 93: I intentionally did not parameterize things since it takes a really long time to run if you run both tests. If you want to test the "other" implementation just set the default and run them all manually. For now i don't think we should configure any of them manually unless specifically listed in the test class name

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatch.java
rodesai, line 38: I think this would be cleaner as just a POJO that encapsulates a batch that's ready to be written, with all the code that actually executes the write moved to `WriteBatcher`. I personally would find this easier to make sense of (because the way it's currently chopped up I'm not sure what's "supposed" to be in `WriteBatcher` vs `WriteBatch`, but there's some more objective benefits too:
    - you could also get rid of the `closed` check below.
    - you wind up with code that's less easy to use incorrectly. Right now, something outside this package could create a WriteBatch, which also makes all these `writer.delete` / `writer.insert` calls without going through the proper flush process. If `WriteBatch` was a POJO you could create it sure, but you'd have to pass it to `WriteBatcher` (which does the write using the right protocol) to actually do anything.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatcher.java
```diff
@@ -0,0 +1,99 @@
+/*
+ *
+ *  * Copyright 2023 Responsive Computing, Inc.
```
rodesai, line 3: nit: happy new year!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/FlushManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+  RemoteWriter<K, P> createWriter(final P tablePartition);
+
+  RemoteWriteResult<P> preFlush();
+
+  RemoteWriteResult<P> postFlush(final long consumedOffset);
+
+  String logPrefix();
+
+  default String failedFlushMsg(final long batchOffset) {
```
rodesai, line 36: can we include these in the return values instead of putting methods in the interface?

rodesai (commented):
not done a full review, just took a pass over the structure of the refactor and had a couple thoughts. I'll go through the mongo stuff next.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -103,6 +103,30 @@ public SegmentPartition(final int tablePartition, final long segmentId) {
       this.segmentId = segmentId;
     }
 
+    @Override
```
rodesai, line 106: I know this isn't part of this PR, but how hard would it be to use the segment start timestamp for the segment ID, rather than the index? It would help some when browsing the database. Right now it's hard to make sense of what stream times the segments correspond to.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MetadataDoc.java
ableegoldman: Seems silly to track extra data, although it's only an extra 4 bytes per kafka partition per state store, so the overall size isn't too bad

I'll look into whether we can use a custom field name instead of `_id` but if not, I guess we can just track the partition separately as well

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
ableegoldman: I had the same thought a while back, I'll file a ticket for it and we can address in a followup PR

ableegoldman: https://www.notion.so/responsivedev/Consider-changing-epoch-from-long-to-int-b9ab9e5623864b71b4a193162fe842fb?pvs=4

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveWindowStore.java
```diff
@@ -81,14 +94,26 @@ public void init(final StateStoreContext storeContext, final StateStore root) {
...
+          appConfigs,
+          config,
           window -> window.windowStartMs >= minValidTimestamp()
       );
 
+      enableBloomFilter = config.getBoolean(WINDOW_BLOOM_FILTER_ENABLED_CONFIG);
+      if (enableBloomFilter) {
+        expectedKeysPerWindow = config.getLong(WINDOW_BLOOM_FILTER_EXPECTED_KEYS_CONFIG);
+        fpp = config.getDouble(WINDOW_BLOOM_FILTER_FPP_CONFIG);
+        bloomFilter = BloomFilter.create(Funnels.byteArrayFunnel(), expectedKeysPerWindow, fpp);
```
rodesai, line 114: I think the way this is implemented now we're susceptible to getting false-negative reads from the filter in a situation like:
Otherwise in a scenario like the following, we'll get false negatives from the filter:
1. start a window
2. write some records
3. crash and recover
4. read from filter

We need to make sure that every record that we write into the current window is also added to the bloom filter. The simplest way I can think to do this we would be to only create and use a bloom filter once we've observed the window start time change. We'll get poorer performance for the first window after a restart, but I think that's ok for now since we're optimizing for a use case with a relatively short window.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -157,6 +157,31 @@ public class ResponsiveConfig extends AbstractConfig {
   private static final String SUBPARTITION_HASHER_DOC = "Hasher to use for sub-partitioning.";
   private static final Class<?> SUBPARTITION_HASHER_DEFAULT = Murmur3Hasher.class;
 
+
+  // ------------------ WindowStore configurations ----------------------
+
+  public static final String WINDOW_BLOOM_FILTER_ENABLED_CONFIG = "responsive.window.bloom.filter.enabled";
+  private static final boolean WINDOW_BLOOM_FILTER_ENABLED_DEFAULT = false;
+  private static final String WINDOW_BLOOM_FILTER_ENABLED_DOC = "Whether to enable a bloom filter for "
```
rodesai, line 165: maybe we can make this an int config, and the value can represent the last x windows to maintain bloom filters for, and default to 0 and limit to 1 for now?

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -157,6 +157,31 @@ public class ResponsiveConfig extends AbstractConfig {
...
+  private static final long WINDOW_BLOOM_FILTER_EXPECTED_KEYS_DEFAULT = 1_000L;
+  private static final String WINDOW_BLOOM_FILTER_EXPECTED_KEYS_DOC = "Expected number of elements for "
+      + "the WindowStore bloom filters, or the approximate number of unique keys per partition that "
+      + "will be present in the most recent window of a hopping or tumbling windowed aggregation. "
+      + "For loads that vary over time, use the maximum keys per partition that may appear in a window. "
+      + "An overestimate will use slightly more memory, but an underestimate will significantly degrade "
+      + "the bloom filter's performance";
+
+  public static final String WINDOW_BLOOM_FILTER_FPP_CONFIG = "responsive.window.bloom.filter.fpp";
+  private static final double WINDOW_BLOOM_FILTER_FPP_DEFAULT = 0.03d;
```
rodesai, line 180: we can probably use a much lower fpp by default here (like 1% or .1%)

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -157,6 +157,40 @@ public class ResponsiveConfig extends AbstractConfig {
...
+  private static final String WINDOW_BLOOM_FILTER_FPP_DOC = "A double representing the desired false "
+      + "positive percentage for the WindowStore bloom filters. A smaller value means fewer unnecessary "
+      + "lookups but requires more heap memory";
+
+  public static final String TIMESTAMP_FIRST_WINDOWED_KEY_CONFIG = "responsive.timestamp.first.windowed.key";
```
rodesai, line 185: we should prefix this config name with mongo, since it's mongo specific

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+
+  // whether to put windowStartMs first in the composite windowed key format in WindowDoc
+  private final boolean timestampFirstOrder;
+
+  private final MongoDatabase database;
+  private final MongoCollection<WindowMetadataDoc> metadata;
+  private final ConcurrentMap<Integer, PartitionSegments> kafkaPartitionToSegments =
+      new ConcurrentHashMap<>();
+
+  private class PartitionSegments {
```
rodesai, line 80: I think it would be easier to read if this were a static class and you pass in the reference to `database`

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+
+  private final String name;
+  private final SegmentPartitioner partitioner;
+
+  // whether to put windowStartMs first in the composite windowed key format in WindowDoc
+  private final boolean timestampFirstOrder;
+
+  private final MongoDatabase database;
+  private final MongoCollection<WindowMetadataDoc> metadata;
+  private final ConcurrentMap<Integer, PartitionSegments> kafkaPartitionToSegments =
```
rodesai, line 77: instances of this class are 1:1 with partitions right? why do we need a map here?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  public String collectionNameForSegment(final SegmentPartition segmentPartition) {
+    return String.format(
+        "%d-%d",
+        segmentPartition.tablePartition, segmentPartition.segmentId
+    );
+  }
+
+  public MongoCollection<WindowDoc> windowsForSegmentPartition(
```
rodesai, line 219: this can be private

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
+/*
+ *
+ *  * Copyright 2023 Responsive Computing, Inc.
```
rodesai, line 3: nit: license formatting looks a little weird (two *s on each line)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  public MongoCollection<WindowDoc> windowsForSegmentPartition(
+      final int kafkaPartition,
+      final SegmentPartition segment
+  ) {
+    return kafkaPartitionToSegments.get(kafkaPartition).segmentWindows.get(segment);
+  }
+
+  public RemoteWriteResult<SegmentPartition> preCommit(
```
rodesai, line 226: it would be good to test the segment expiry/create logic as the stream time moves past the segment boundaries

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveWindowStore.java
ableegoldman: Yes, this is definitely the case. One of the tests covers it and is failing, I just haven't pushed a fix yet. My plan is to just ignore the filter unless we are certain it contains the entire window so far. 

I think that's probably fine for cases with smaller window sizes, we can probably go 15 minutes without a restart/rebalance for example. But we'll definitely need something a bit more sophisticated for longer window sizes (or account for that in the assignor to avoid moving tasks with windowed state, if possible)

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
rodesai: nvm this is supposed to be a number between 0 and 1, so this is 3% fppp

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: No, there's only one instance of a RemoteTable per logical state store -- ie it's shared by all kafka partitions.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: Will do (also open to naming suggestions for any of these classes, I think better names would help but I can only come up with so many names lol)

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
ableegoldman: I guess? Would be nice to do some tests to find out how much memory is used. I just went with the actual default this class uses.
Anyways just FYI I'm going to split the bloom filter out into a separate PR so we can punt on this for a minute

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
ableegoldman: Ack, will do in the followup PR

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+   */
+  @Override
+  public WriteModel<WindowDoc> insert(
+      final int kafkaPartition,
+      final WindowedKey windowedKey,
+      final byte[] value,
+      final long epochMillis
+  ) {
+    final var partitionSegments = kafkaPartitionToSegments.get(kafkaPartition);
+    partitionSegments.segmentBatch.updateStreamTime(windowedKey.windowStartMs);
```
rodesai, line 318: I think it would be better for the stream time to be tracked by the `WriteBatcher` (which is common for all stores) and passed in to `preFlush`. That way we don't have to repeat this for every store implementation, and it's not tracked as a side effect (which makes it a bit harder to follow how it's tracked)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatcher.java
```diff
@@ -0,0 +1,99 @@
...
+ *  *
+ *  * Unless required by applicable law or agreed to in writing, software
+ *  * distributed under the License is distributed on an "AS IS" BASIS,
+ *  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  * See the License for the specific language governing permissions and
+ *  * limitations under the License.
+ *
+ */
+
+package dev.responsive.kafka.internal.db;
```
rodesai, line 19: we should unit test that the write batcher executes batches correctly

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+      final int kafkaPartition,
+      final SegmentPartition segment
+  ) {
+    return kafkaPartitionToSegments.get(kafkaPartition).segmentWindows.get(segment);
+  }
+
+  public RemoteWriteResult<SegmentPartition> preCommit(
+      final int kafkaPartition
+  ) {
+    final var partitionSegments = kafkaPartitionToSegments.get(kafkaPartition);
```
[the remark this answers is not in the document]
rodesai: We should add tests for the batch rolling logic

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: Technically it's possible to invoke this since our store implementation delegates to `delete` when it receives an insert with a null value. It's probably rare but still very much possible (and definitely allowed) for an aggregation to return `null` so I definitely intend to implement this at some point. However based on the code snippet of the aggregator app, they definitely don't/can't return `null`, so I decided to put this off for the first iteration of the window store.

tl;dr yes we should implement it, no we don't need to ASAP

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
ableegoldman: FYI I changed it to use the segment start timestamp in this PR, for the collection name only. Honestly, if we do this then imo we should just go ahead and convert all instances of the segment id to "segment start timestamp" instead, and get rid of the "segment id" concept entirely

That feels reasonable but I'll save that for a followup PR. Filed a ticket: [here](https://www.notion.so/responsivedev/Replace-segment-id-with-start-timestamp-3582ca33324148169cca8d1d04b870de?pvs=4)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+    for (final long segmentId : segments.segmentBatch.segmentRoll.segmentsToExpire) {
+      segments.expireSegment(new SegmentPartition(kafkaPartition, segmentId));
+    }
+    segments.segmentBatch.finalizeRoll();
+    return RemoteWriteResult.success(null);
+  }
+
+  @Override
+  public long fetchOffset(final int kafkaPartition) {
+    final WindowMetadataDoc result = metadata.find(
```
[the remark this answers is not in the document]
ableegoldman: The `fetchOffset` (maybe better named "fetchStartOffset" or "fetchRestoreOffset") is always invoked just a few lines after the `init` call, so I think it would be rare to get fenced between the `init` and `fetchOffset` calls. Being fenced by offset would be next to impossible since it would require another client to start up, overwrite the epoch, process a batch of records and then flush them to bump the offset, all in the time between when the table is initialized and when we read out the starting offset to restore from. 

An epoch mismatch is probably rare as well, but that case is at least possible. I suppose it doesn't hurt to check the epoch here. I checked the Streams callstack and it looks like we can safely throw TaskMigratedException from here and have it interpreted correctly so that's good

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: Great question -- if only there were remotely useful javadocs for Mongo's java API 😞

That said, based on my own testing, it seems to be a no-op -- so this is safe

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: yep!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/MongoKVFlushManager.java
ableegoldman: Good catch.

Somewhat unrelated to this PR, but I noticed we're updating both the offset AND epoch in the KV table's `#setOffset` -- but we should be ok to only update the offset, right? Not sure how much overhead really could be coming from this additional write but if we really want to minimize any unnecessary writes to maximize Mongo's performance...we should get rid of the epoch write here, right?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactFlushManager.java
ableegoldman: > it'll overwrite already processed data...

Isn't this the basic premise of a fact table -- that this can happen but it's acceptable? eg for the deduplicator case, we may overwrite old values but we'll never "erase" data that already exists in the table. The assumption is that fact tables are only used in cases like that where the value can not necessarily be trusted to be up to date, but the "is this key present in the state store" answer is always consistent -- which it still would be if we accidentally restored some extra data?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WriteBatch.java
ableegoldman: Cleaned it up so that this class only keeps track of the actual writers for the batch, rather than handling the writes themselves. The main goal of this class was to encapsulate the writers for a single given batch, but since we have to iterate over all the buffered writes to build up the set of writers, I figured we may as well stage those writes as well. 

We now have to iterate over the batch twice now (first to build up the set of writers based on the keys, then later to issue the actual `writer.insert/delete` operations) but that's not too big a price if the outcome feels cleaner (does it?)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WindowFlushManager.java
```diff
@@ -0,0 +1,106 @@
...
+
+    pendingFlushSegmentMetadata.finalizeRoll();
+    return RemoteWriteResult.success(null);
+  }
+
+  /**
+   * Persist the latest consumed offset and stream-time corresponding to the batch that was just
+   * flushed to the remote table
+   */
+  public abstract RemoteWriteResult<SegmentPartition> updateOffsetAndStreamTime(
```
rodesai, line 85: this can be protected

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -0,0 +1,472 @@
...
+        final long streamTime,
+        final long epoch
+    ) {
+      this.database = database;
+      this.partitioner = partitioner;
+      this.epoch = epoch;
+      this.segmentWindows = new ConcurrentHashMap<>();
+
+      final var activeSegments = partitioner.activeSegments(kafkaPartition, streamTime);
+      if (activeSegments.isEmpty()) {
```
rodesai, line 102: would it make sense to do this from the base flush manager instead so all the segment creation/deletion logic is in one place?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
rodesai: sounds good. thanks!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/SegmentPartitioner.java
```diff
@@ -268,14 +296,22 @@ private long minValidTs(final long streamTime) {
...
-    public final long[] segmentsToCreate;
+    private final long[] segmentsToExpire;
+    private final long[] segmentsToCreate;
 
     public SegmentRoll(final long[] segmentsToExpire, final long[] segmentsToCreate) {
       this.segmentsToExpire = segmentsToExpire;
       this.segmentsToCreate = segmentsToCreate;
     }
 
+    public long[] segmentsToExpire() {
```
rodesai, line 307: nit: make this an immutable list so you know it can't be mutated by a caller

rodesai (commented):
Thanks for the updates, I found it much easier to follow! Left a few more comments/questions but I think this is almost ready

PR #213, opened 2024-01-16
Use constant collection names and validate TableName length

ableegoldman:
Mongo limits database names to 65 characters (at the most -- may be even smaller in some cases) and also enforces the combined `<database_name>.<collection_name>` string to not exceed 128 characters. 

We should validate the table name length up front for fast-failure (I actually ran into this in a test with a long name). We should also remove the table name from the collection naming scheme for the MongoKVTable. We don't need to include the table name again since the collections are by definition scoped to that database, which is named after the state store. This way we are able to use up to the full allotted character count for table names (since otherwise the metadata collections could exceed 128 characters with a 64-character table name).

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/utils/TableName.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/utils/TableName.java
@@ -18,38 +18,48 @@ package dev.responsive.kafka.internal.utils;
 
 import java.util.Objects;
 import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * {@code TableName} represents the name for a table and all variations
  * of it - some characters that are valid for Kafka names are not valid
- * as Cassandra table names.
+ * as table names.
  */
 public class TableName {
 
-  // cassandra names are case-insensitive and can only contain
+  private static final Logger LOG = LoggerFactory.getLogger(TableName.class);
+
+  private static final int MAXIMUM_LENGTH = 65;
+
+  // table names are case-insensitive and should only contain
   // alphanumeric and underscore characters
-  private static final Pattern INVALID_CASSANDRA_CHARS =
+  private static final Pattern INVALID_CHARS =
       Pattern.compile("[^a-zA-Z0-9_]");
 
   private final String kafkaName;
-  private final String remoteName;
+  private final String tableName;
 
   public TableName(final String kafkaName) {
     this.kafkaName = kafkaName;
     final var escaped = kafkaName.replaceAll("_", "__"); // escape existing underscores
-    remoteName = INVALID_CASSANDRA_CHARS
+    tableName = INVALID_CHARS
         .matcher(escaped)
         .replaceAll("_")
         .toLowerCase();
+
+    if (tableName.length() > MAXIMUM_LENGTH) {
+      LOG.error("Invalid table name for state store {}, must be no more than {} characters long"
+                    + " but length was {}", tableName, MAXIMUM_LENGTH, tableName.length());
+    }
   }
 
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -72,11 +73,11 @@ public MongoKVTable(final MongoClient client, final String name) {
...
     final MongoDatabase database = client.getDatabase(name).withCodecRegistry(pojoCodecRegistry);
-    genericDocs = database.getCollection(name);
-    docs = database.getCollection(name, KVDoc.class);
+    genericDocs = database.getCollection(KV_COLLECTION_NAME);
+    docs = database.getCollection(KV_COLLECTION_NAME, KVDoc.class);
 
-    genericMetadata = database.getCollection(name + METADATA_COLLECTION_SUFFIX);
-    metadata = database.getCollection(name + METADATA_COLLECTION_SUFFIX, MetadataDoc.class);
+    genericMetadata = database.getCollection(METADATA_COLLECTION_NAME);
+    metadata = database.getCollection(METADATA_COLLECTION_NAME, MetadataDoc.class);
```
agavra, line 80: these lines are the "change" right? Just want to make sure it doesn't affect existing C* deployments

agavra (approved):

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
ableegoldman: yes, sorry, to clarify this only affects mongo

ableegoldman: The only thing that would potentially affect C* as well is the new length check -- but no existing deployments are past that limit. And anyways, I think C* actually does have a length limit as well, though it was more ambiguous

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/TableName.java
```diff
@@ -18,38 +18,48 @@
...
     this.kafkaName = kafkaName;
     final var escaped = kafkaName.replaceAll("_", "__"); // escape existing underscores
-    remoteName = INVALID_CASSANDRA_CHARS
+    tableName = INVALID_CHARS
         .matcher(escaped)
         .replaceAll("_")
         .toLowerCase();
+
+    if (tableName.length() > MAXIMUM_LENGTH) {
+      LOG.error("Invalid table name for state store {}, must be no more than {} characters long"
```
ableegoldman, line 52: FYI: I do intend to eventually throw an error here, but some of our tests are exceeding this limit. I'm a bit confused because from the docs it sounds like C* has, if anything, a lower character count limit on table names...but it might be that the limits I'm looking at are only for Scylla, and the embedded C* we use in the test harness doesn't have such a limit. Hence things passing now

Frankly it would be kind of a bummer to have to limit all our test names to 65 characters or less, but we'll have to do so for anything using a Mongo backend. And since we should probably parametrize the tests to use both C* and Mongo in the end, the only thing we'd be restricting "unnecessarily" are the Cassandra-specific unit tests.

Personally I'm down to just rename them, but it's certainly a bit annoying. Any strong feelings on this before I go ahead?

ableegoldman: Oh actually, the tests I thought were too long were just timing out for other reasons. 65 characters is definitely still a bit of a constraint, but it leaves a fair amount of room. I'm going to go ahead with throwing the error here

PR #220, opened 2024-02-07
add flush jitter config

rodesai:
adds a property that configures jitter around the flush interval to spread out load from flushes.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
@@ -139,6 +139,12 @@ public class ResponsiveConfig extends AbstractConfig {
   public static final long STORE_FLUSH_INTERVAL_TRIGGER_DEFAULT
       = Duration.ofSeconds(30).toMillis();
 
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER_JITTER_CONFIG = "responsive.store.flush.trigger.local.jitter.ms";
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER_JITTER_DOC = "The jitter to apply to the flush interval." +
+      "For flush interval i and jitter j, the flush interval for a given flush will a randomly selected duration" +
+      "between i-j and i+j.";
+  public static final int STORE_FLUSH_INTERVAL_TRIGGER_JITTER_DEFAULT = 0;
+
   // TODO: consider if we want this as a local, global or per-store configuration
   public static final String MAX_CONCURRENT_REQUESTS_CONFIG = "responsive.max.concurrent.requests";
   private static final String MAX_CONCURRENT_REQUESTS_DOC = "The maximum number of requests"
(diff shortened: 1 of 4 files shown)


kafka-client/src/test/java/dev/responsive/kafka/internal/stores/CommitBufferTest.java
```diff
@@ -577,6 +581,46 @@ public void shouldOnlyFlushWhenIntervalTriggerElapsed() {
...
+        return Optional.of(clock.get());
+      }
+      clock.set(clock.get().plus(Duration.ofSeconds(1)));
+    }
+    return Optional.empty();
+  }
+
+  @Test
+  public void shouldApplyJitterToFlushInterval() {
+    final AtomicReference<Instant> clock = new AtomicReference<>(Instant.now());
```
agavra, line 603: nit: can we put this in given/when/then and write a small comment about exactly what/how it's testing? I need to critically think in order to figure out what this test is doing

agavra (approved):
LGTM

kafka-client/src/test/java/dev/responsive/kafka/internal/stores/CommitBufferTest.java
rodesai: I'll definitely add a comment. It's a hard test to format as a `given-when-then` since it's testing randomness over many iterations.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -139,6 +139,12 @@ public class ResponsiveConfig extends AbstractConfig {
   public static final long STORE_FLUSH_INTERVAL_TRIGGER_DEFAULT
       = Duration.ofSeconds(30).toMillis();
 
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER_JITTER_CONFIG = "responsive.store.flush.trigger.local.jitter.ms";
+  public static final String STORE_FLUSH_INTERVAL_TRIGGER_JITTER_DOC = "The jitter to apply to the flush interval."
+      + "For flush interval i and jitter j, the flush interval for a given flush will a randomly selected duration"
```
ableegoldman, line 144:
```suggestion
      + "For flush interval i and jitter j, the actual flush interval for a given flush will be a randomly selected duration"
```

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -439,6 +439,7 @@ private boolean triggerFlush() {
           now
       );
       timeTrigger = true;
+      flushTriggers.reset();
```
ableegoldman, line 442: Maybe this is a bit pedantic, but why not just have it "reset" inside the `getInterval` call?

Or better yet, pull all the logic in line 436 inside an evaluation method on the FlushTriggers class (eg #shouldTriggerOnInterval) and have it test the interval + reset it with the next jitter-applied interval?

rodesai: I thought about doing it this way originally, but chose the current structure to keep the logic for what happens when a flush actually happens in a single place. So FlushTriggers stays dumb and just tells you what the triggers are, and can be reset to apply a jitter. And CommitBuffer actually decides whether or not a flush actually happens and does a reset if it actually does flush. That said, it probably makes sense to move `reset` to the branch where the flush actually happens rather than just for the time interval trigger.

PR #222, opened 2024-02-13
IQ integration test

ableegoldman:
Adds a basic integration test for IQv1 

As we finish implementing other query types, I'll add them to this test as well


agavra (approved):
quickly skimmed it, test makes sense to me

PR #223, opened 2024-02-13
Mongo KV range queries (option 1)

ableegoldman:
Implementation of `range` and `all` queries for the MongoKVTable.

In order to follow Streams semantics, we need to be able to differentiate data that falls within a kafka partition. There are two options:

1. Add a `kafkaPartition` field to KVDoc and filter range queries with this
2. Separate data into different collections for each kafka partition

This PR implements option 1, although we may want to go with option 2 for two reasons: (a) increased read/write/storage due to added field (though it's just an int) and (b) in the current implementation for Mongo window stores, we need separate collections for each kafka partition anyway (to track stream-time on a partition basis), so it probably makes sene to be consistent.

I'll follow up with a PR for option 2 after finishing the window store range queries

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
@@ -24,47 +24,59 @@ import org.bson.codecs.pojo.annotations.BsonId;
 public class KVDoc {
 
   // TODO(agavra): figure out if we can use @BsonProperty to set the names explicitly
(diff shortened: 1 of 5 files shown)


agavra:
@ableegoldman regarding the options we should make it a priority that we can increase/decrease number of partitions without having to make a massive storage change. we may even want to consider less-efficient range scans and using the partitioner to filter client-side based on key.

ableegoldman:
closing in favor of the alternative approach via client-side filtering: https://github.com/responsivedev/responsive-pub/pull/251

PR #224, opened 2024-02-13
Mongo window range queries

ableegoldman:
Finish implementation for the remaining range queries of MongoWindowedTable. 

The two queries take a slightly different approach, as documented in code comments for each. Basically for the all-keys time range scan we can identify exact borders and issue a precise range query for only valid data. However for the key and time range scan, we have to scan over all keys and post-filter them when timestamp-first order is used (and vice versa for key-first order). This is due to the byte layout of the composite primary keys, and the fact that key bytes are variable length

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
@@ -66,6 +66,9 @@ public class MongoWindowedTable implements RemoteWindowedTable<WriteModel<Window
 
   private static final UpdateOptions UPSERT_OPTIONS = new UpdateOptions().upsert(true);
 
+  // The "minimum" possible key when comparing bytewise, used to define range query bounds
+  private static final Bytes MIN_KEY = Bytes.wrap(new byte[0]);
+
   private final String name;
   private final SegmentPartitioner partitioner;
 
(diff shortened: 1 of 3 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -457,13 +459,37 @@ public KeyValueIterator<WindowedKey, byte[]> backFetch(
...
   ) {
-    throw new UnsupportedOperationException("fetchRange not yet supported for Mongo backends");
+    final List<KeyValueIterator<WindowedKey, byte[]>> segmentIterators = new LinkedList<>();
+    final var partitionSegments = kafkaPartitionToSegments.get(kafkaPartition);
+
+    for (final var segment : partitioner.range(kafkaPartition, timeFrom, timeTo)) {
+      final var segmentWindows = partitionSegments.segmentWindows.get(segment);
+
+      // Since we use a flat keyspace by concatenating the timestamp with the data key and have
+      // variable length data keys, it's impossible to request only valid data that's within
```
rodesai, line 474: question to clarify my understanding - why does the variability of the length of the data key matter? even if we knew all the keys were the same size, this filter would still pick up too much data right? if the key is first, you'd still pick up a bunch of time windows you don't want. And if the time is first you'd still pick up a bunch of keys you don't want.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -483,7 +509,35 @@ public KeyValueIterator<WindowedKey, byte[]> fetchAll(
       final long timeFrom,
       final long timeTo
   ) {
-    throw new UnsupportedOperationException("fetchAll not yet supported for Mongo backends");
+    if (!timestampFirstOrder) {
+      throw new UnsupportedOperationException("Range queries such as fetchAll require stores to be "
```
rodesai, line 513: this is because mongo doesn't let you scan the whole table?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -457,13 +459,37 @@ public KeyValueIterator<WindowedKey, byte[]> backFetch(
...
+      final var segmentWindows = partitionSegments.segmentWindows.get(segment);
+
+      // Since we use a flat keyspace by concatenating the timestamp with the data key and have
+      // variable length data keys, it's impossible to request only valid data that's within
+      // the given bounds. Instead issue a broad request from the valid bounds and then post-filter
+      final var lowerBound = compositeKey(keyFrom, timeFrom);
+      final var upperBound = compositeKey(keyTo, timeTo);
 
+      final FindIterable<WindowDoc> fetchResults = segmentWindows.find(
+          Filters.and(
```
rodesai, line 480: I'm wondering if we should include these fields outside of the key and filter on them server-side. This has the potential to jack up our data transfer costs.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -495,6 +549,34 @@ public KeyValueIterator<WindowedKey, byte[]> backFetchAll(
     throw new UnsupportedOperationException("backFetchAll not yet supported for MongoDB backends");
   }
 
+  @Override
+  public KeyValueIterator<WindowedKey, byte[]> all(
+      final int kafkaPartition,
+      final long streamTime
```
rodesai, line 555: don't we need to filter on this time?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -457,13 +459,37 @@ public KeyValueIterator<WindowedKey, byte[]> backFetch(
   @Override
   public KeyValueIterator<WindowedKey, byte[]> fetchRange(
       final int kafkaPartition,
-      final Bytes fromKey,
-      final Bytes toKey,
+      final Bytes keyFrom,
```
rodesai, line 462: question about the state store api in general - what are these cross-key APIs actually used for? All the kafka streams processors are scoped to one key afaik.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: I guess this was a "prevent users from shooting themselves in the foot" moment -- it's possible, just really inefficient since we have to scan everything in between (see comment about variable length keys).

Of course if we're going to go the route of scanning everything in order to enable partition scaling, then this constraint doesn't make any sense. I'll remove it

PR #228, opened 2024-02-16
set default consistency level to quourm with ability to override

agavra:


kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -246,6 +259,26 @@ public class ResponsiveConfig extends AbstractConfig {
           REMOTE_TABLE_CHECK_INTERVAL_MS_DEFAULT,
           Importance.LOW,
           REMOTE_TABLE_CHECK_INTERVAL_MS_DOC
+      ).define(
```
rodesai, line 262: I think it would be better to compute the default from the write consistency level. Don't want to block the PR on it though.

rodesai (approved):
LGTM

agavra:
CI/CD is struggling so I'm going to merge after local run
```
BUILD SUCCESSFUL in 6m 50s
76 actionable tasks: 51 executed, 25 up-to-date
```

PR #231, opened 2024-02-16
Segment Partitioner: Parametrize the key type

apourchet:
This will eventually allow us to reuse this segment partitioner for
session keys, in addition to the window keys that are currently
supported.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/TimePartitionableKey.java
```diff
@@ -0,0 +1,11 @@
+package dev.responsive.kafka.internal.db;
+
+import java.sql.Time;
+
+public interface TimePartitionableKey<T> extends Comparable<T> {
+  /**
+   * @return the timestamp that partitioners can use to partition
+   *         keys of this type into different segments.
+   */
+  long getPartitionTimestamp();
```
apourchet, line 10: This is probably a bad name for this method.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraWindowedTable.java
```diff
@@ -49,6 +49,7 @@
 import dev.responsive.kafka.internal.stores.RemoteWriteResult;
 import dev.responsive.kafka.internal.utils.Iterators;
 import dev.responsive.kafka.internal.utils.WindowedKey;
+import java.awt.Window;
```
ableegoldman, line 52: accidental import? 😉

kafka-client/src/main/java/dev/responsive/kafka/internal/db/TimePartitionableKey.java
ableegoldman: two notes on the naming:
1. style-wise, we use "get"-less getters -- eg `timestamp()` rather than `getTimestamp`
2. The term "partition" is massively overloaded right now so it's probably best to be specific or avoid it altogether. In this case the "partitions" are really just the "segments" -- it made more sense in the original C* implementation, where there was a concept of a partition in the C* table itself. Mongo doesn't exactly have partitions but the name stuck for things like the "SegmentPartitioner"
In this case, nothing feels grammatically perfect, but I'd suggest calling this method just `timestamp()` and then naming the class itself `SegmentKey` or `SegmentableKey` (is "segmentable" a word? definitely an awkward one if so...)
I think the important thing is really the javadocs (which are spot on here)

ableegoldman (commented):
I think this makes sense overall, but let me throw out one suggestion (which I haven't thought all the way through so I'll let you figure out if it actually makes sense or not):

Would it be possible to avoid parameterizing the SegmentPartitioner and instead just hardcode the key type as `TimePartitionableKey` (or whatever we end up calling it)? I'm just thinking that from the SegmentPartitioner's point of view, it doesn't really care whether the key is a WindowedKey or a SessionKey -- the only thing that matters is it's a key that's partitioned according to a single timestamp field.

apourchet:
> I think this makes sense overall, but let me throw out one suggestion (which I haven't thought all the way through so I'll let you figure out if it actually makes sense or not):
> 
> Would it be possible to avoid parameterizing the SegmentPartitioner and instead just hardcode the key type as `TimePartitionableKey` (or whatever we end up calling it)? I'm just thinking that from the SegmentPartitioner's point of view, it doesn't really care whether the key is a WindowedKey or a SessionKey -- the only thing that matters is it's a key that's partitioned according to a single timestamp field.

This makes a ton of sense and I hadn't thought about that. However because `TimePartitionableKey<K>` (now `SegmentableKey<K>` per your suggestion, it's a word turns out) implements `Comparable<K>`, it needs to be a parametrized type. Which means that, in turn, `SegmentPartitioner` needs to be parametrized (since we cannot use the raw `SegmentableKey` type without `warning: [rawtypes]`).

I haven't found a way to make this `SegmentableKey` a raw type, but I guess it's for the same reason Comparable is also parametrized.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/WindowedTableCache.java
```diff
@@ -0,0 +1,68 @@
...
+    tables.put(spec.tableName(), table);
+    return table;
+  }
+
+  /**
+   * @param name the name of the table
+   * @return the table, if it was already created by
+   *         {@link #create(CassandraTableSpec, SegmentPartitioner)}
+   */
+  public T getTable(final String name) {
```
agavra, line 64: I think this should be `synchronized` (same with the original `TableCache` implementation -- shame on me!)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentableKey.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package dev.responsive.kafka.internal.db;
+
+public interface SegmentableKey<T> extends Comparable<T> {
+  /**
+   * @return the timestamp that partitioners can use to partition
+   *         keys of this type into different segments.
+   */
+  long segmentTimestamp();
```
agavra, line 24: nit: this is only used when passed into the `segmentId()` function. I think this class would make more sense if this just returned the `segmentId` directly instead of exposing a segment timestamp.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentableKey.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.internal.db;
+
+public interface SegmentableKey<T> extends Comparable<T> {
```
agavra, line 19: re: the discussion you were having with @ableegoldman around the awkwardness of the type system - I feel like this interface shouldn't extend `Comparable<T>` and instead we should have:
```java
public class WindowedKey implements SegmentableKey, Comparable<WindowedKey> {
```
This interface doesn't actually require that the `SegmentableKey` is compareable to another `SegmentableKey` (in fact, it doesn't guarantee that at all, so it's a little weird to pass the type parameter through).

If you do this, then you don't need to parameterize the `SegmentParitioner`.

agavra (commented):
LGTM! Thanks @apourchet! 🙌 (Not hitting the green check since Sophie has more context)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentableKey.java
apourchet: The segment id returned is based on other things such as segment interval, which is out of scope / responsibility for these keys. I considered having the `segmentIntervalMs` parameter passed into this function to return the actual `segmentId` but thought it felt worse than just returning a timestamp according to which the key-space is segmentable.

Additionally the timestamp => segment ID calculation is also done in the partitioner (to figure out which segments it should search, for instance, given a raw timestamp), which would mean we do this math in two separate places:
```    
return Long.max(0, windowTimestamp / segmentIntervalMs);
```
I'm open to making that change, though if you feel strongly about it.

ableegoldman: FWIW I agree with Antoine here -- best to keep this a "pure" key class that only contains data about itself, and let the partitioner determine how that key data maps to a specific segment/partition/etc

ableegoldman:
@apourchet  This is not strictly speaking related, and by no means necessary, but given the nature of what you're refactoring here I'm just wondering if it would make sense to include another small refactoring as part of this cleanup: [replacing the segmentId with a segmentStartTimestamp](https://www.notion.so/responsivedev/Replace-segment-id-with-start-timestamp-3582ca33324148169cca8d1d04b870de)

We definitely don't need to complicate things any further -- I just wanted to point this out in case this would be a good opportunity to make this change. The linked ticket should have all the context but the basic idea is  to just get rid of the "segment id" concept altogether, and use a timestamp (specifically, the starting timestamp for that segment range) to identify segments. 

Right now we're using the "segmentStartTimestamp" when naming the Mongo collections, but internally use the segment id for everything else. I think it would be best to consolidate these and do only one or the other, and as Rohan pointed out, the segmentStartTimestamp is better for naming segments since it tells us directly what's in the segment. A segmentId feels nice in theory, but given the values we're dealing with for typical timestamps, the segmentId is less a nice clean "1, 2, 3" and more like "135927876245"

Which is just nonsense at best, and at worst is misleading in that it appears to be a timestamp, but is not

Anyways, in sum: don't feel like you need to pick this up as part of the current PR, but if you find yourself messing with the segmentId stuff a bunch, maybe it makes sense to just knock it out now. Up to you

kafka-client/src/main/java/dev/responsive/kafka/internal/db/ColumnName.java
```diff
@@ -30,7 +30,7 @@ public enum ColumnName {
 
   // shared partition key column
   PARTITION_KEY("partitionKey", "partitionkey"),
-  SEGMENT_ID("segmentId", "segmentid"),
+  SEGMENT_START_TIMESTAMP("segmentStartTimestamp", "segmentstarttimestamp"),
```
agavra, line 33: @ableegoldman we're no longer running any Scylla-based soaks right? this would break compatibility

apourchet: I checked with @ableegoldman and rodesai offline, we should be OK to make this change.

apourchet:
Closing in favor of #243

PR #232, opened 2024-02-16
add an option for migration mode and avoid unnecessary ttl writes

rodesai:
This patch starts the work to support running responsive in "migrate" mode,
which will be used to migrate apps to Responsive, or to a different Responsive
store. For now, all its used to do is enable a write optimization for TTL stores,
where we don't write records that are older than the TTL when the migration
started.


kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
```diff
@@ -19,6 +19,7 @@
 import dev.responsive.kafka.internal.stores.SchemaTypes.KVSchema;
 import dev.responsive.kafka.internal.utils.TableName;
 import java.time.Duration;
+import java.time.Instant;
```
agavra, line 22:
```suggestion
```

agavra (approved):
LGTM! Thanks @rodesai - we may want a quick test for it if it isn't too hard to add since it's in the main code path.

agavra:
@rodesai I'm going to go ahead and merge this so I can build on top of it

PR #233, opened 2024-02-17
add changelog migration tooling

agavra:
Testing Done:
1. The integration test here allows us to manually verify that the migration populated data into what we expected and that there were no additional topics created
2. Deployed to an EKS cluster with a VPC Peered Cassandra cluster and migrated from a Mongo-backed application to Cassandra while the application was still running.

We still need to figure out the best way to deliver this.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dev.responsive.kafka.bootstrap;
+
+import static dev.responsive.kafka.bootstrap.ChangelogMigrationConfig.CHANGELOG_TOPIC_CONFIG;
+import static dev.responsive.kafka.bootstrap.ChangelogMigrationConfig.TABLE_NAME_CONFIG;
+
+import dev.responsive.kafka.api.ResponsiveKafkaStreams;
+import dev.responsive.kafka.api.config.ResponsiveConfig;
+import dev.responsive.kafka.api.config.ResponsiveMode;
+import dev.responsive.kafka.api.stores.ResponsiveStores;
+import java.time.Instant;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.kstream.KTable;
+import org.apache.kafka.streams.kstream.Materialized;
+import org.apache.kafka.streams.processor.api.Processor;
+import org.apache.kafka.streams.processor.api.ProcessorContext;
+import org.apache.kafka.streams.processor.api.Record;
+
+public class ChangelogMigrationTool {
+
+  private static final org.slf4j.Logger LOG =
+      org.slf4j.LoggerFactory.getLogger(ChangelogMigrationTool.class);
+
+  private final ResponsiveKafkaStreams streams;
+
+  public ChangelogMigrationTool(final Properties properties) {
+    properties.put(StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG, StreamsConfig.OPTIMIZE);
+    properties.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    properties.put(ResponsiveConfig.RESPONSIVE_MODE, ResponsiveMode.MIGRATE.name());
+
+    // it is possible to push this number higher, but this is a relatively safe
+    // number and still gets decent performance -- use putIfAbsent in case the
+    // customer wants to override this to push more performance at the risk of
+    // stability
+    properties.putIfAbsent(ResponsiveConfig.STORE_FLUSH_RECORDS_TRIGGER_CONFIG, 10_000);
+
+    final ChangelogMigrationConfig config = new ChangelogMigrationConfig(properties);
+    this.streams = buildStreams(config, properties);
+  }
+
+  public ResponsiveKafkaStreams getStreams() {
+    return streams;
+  }
+
+  private ResponsiveKafkaStreams buildStreams(
+      final ChangelogMigrationConfig config,
+      final Properties properties
+  ) {
+    final StreamsBuilder builder = new StreamsBuilder();
+    final String source = config.getString(CHANGELOG_TOPIC_CONFIG);
+    final String tableName = config.getString(TABLE_NAME_CONFIG);
+    final AtomicLong processed = new AtomicLong();
+
+    final KTable<byte[], byte[]> table =
+        builder.table(
+            source,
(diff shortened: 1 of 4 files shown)


kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
```diff
@@ -0,0 +1,102 @@
...
+
+public class ChangelogMigrationTool {
+
+  private static final org.slf4j.Logger LOG =
+      org.slf4j.LoggerFactory.getLogger(ChangelogMigrationTool.class);
+
+  private final ResponsiveKafkaStreams streams;
+
+  public ChangelogMigrationTool(final Properties properties) {
+    properties.put(StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG, StreamsConfig.OPTIMIZE);
```
rodesai, line 46: minor thing, but can we just explicitly turn on the ktable optimization, and leave the other ones off?

rodesai, line 46: the caller is expected to set the application id?

kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
```diff
@@ -0,0 +1,102 @@
...
+          private ProcessorContext<byte[], byte[]> context;
+
+          @Override
+          public void init(final ProcessorContext<byte[], byte[]> context) {
+            this.context = context;
+          }
+
+          @Override
+          public void process(final Record<byte[], byte[]> record) {
+            if (processed.incrementAndGet() % 10_000 == 0) {
```
rodesai, line 91: this seems pretty frequent. consider changing to a time based threshold (e.g. log every 10 seconds)

kafka-client/src/test/java/dev/responsive/kafka/bootstrap/ChangelogMigrationToolTest.java
```diff
@@ -0,0 +1,185 @@
...
+    final Properties bootstrapProps = new Properties();
+    bootstrapProps.putAll(getProperties());
+    bootstrapProps.put(ChangelogMigrationConfig.CHANGELOG_TOPIC_CONFIG, name + "-count-changelog");
+    bootstrapProps.put(ChangelogMigrationConfig.TABLE_NAME_CONFIG, name + "-count");
+    final ChangelogMigrationTool app =
+        new ChangelogMigrationTool(bootstrapProps);
+    startAppAndAwaitRunning(Duration.ofSeconds(120), app.getStreams());
+
+    // Manually Check Cassandra
+    Thread.sleep(100000);
```
rodesai, line 141: we could automate the check here right?

kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
```diff
@@ -0,0 +1,102 @@
...
+      final ChangelogMigrationConfig config,
+      final Properties properties
+  ) {
+    final StreamsBuilder builder = new StreamsBuilder();
+    final String source = config.getString(CHANGELOG_TOPIC_CONFIG);
+    final String tableName = config.getString(TABLE_NAME_CONFIG);
+    final AtomicLong processed = new AtomicLong();
+
+    final KTable<byte[], byte[]> table =
+        builder.table(
```
rodesai, line 74: We should explicitly consume the table with byte[] serdes. Otherwise it will just use whatever the default serdes are, which might muck up the data. It also will put the type associated with the default serdes in the records - so, for example in your test if you try to do something like `record.value.length` I think you'd get an exception (since it's actually a Long).

kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
agavra: yes, that way if they need to run this multiple times for whatever reason they can set whatever ID they need.

kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
```diff
@@ -0,0 +1,124 @@
...
+  public ResponsiveKafkaStreams buildStreams() {
+    final StreamsBuilder builder = new StreamsBuilder();
+    final String source = config.getString(CHANGELOG_TOPIC_CONFIG);
+
+    final KTable<byte[], byte[]> table =
+        builder.table(
+            source,
+            Materialized
+                .<byte[], byte[]>as(ResponsiveStores.keyValueStore(params))
+                .withValueSerde(Serdes.ByteArray())
```
rodesai, line 93: These will set the serdes to use for reading/writing the store, but does it also get used for deserializing the source topic? Consider passing `Consumed` to `table` to be extra sure.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -230,7 +230,7 @@ public BoundStatement setOffset(
 
   @Override
   public long approximateNumEntries(final int kafkaPartition) {
-    return client.count(name(), kafkaPartition);
+    throw new UnsupportedOperationException("all is not supported on fact tables");
```
rodesai, line 233: was this intentional? this is also not `all`

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
```diff
@@ -370,7 +371,9 @@ public KeyValueIterator<K, Result<K>> all(
   ) {
     return Iterators.kv(
         Iterators.filter(
-            buffer.getReader().entrySet().iterator(),
+            // make a full copy to avoid concurrent mod exceptions
+            // we can worry about performance for all() some other time
+            new TreeMap<>(buffer.getReader()).entrySet().iterator(),
```
rodesai, line 376: probably better to use `ImmutableMap.copyOf`

kafka-client/src/test/java/dev/responsive/kafka/bootstrap/ChangelogMigrationToolIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+          .store(StoreQueryParameters.fromNameAndType(
+              tableName, QueryableStoreTypes.keyValueStore()));
+
+      final Map<Long, Long> all = new HashMap<>();
+      for (long k = 0; k < numKeys; k++) {
+        all.put(k, table.get(k));
+      }
+
+      for (long k = 0; k < numKeys; k++) {
+        assertThat(all, Matchers.hasEntry(k, 1L));
```
rodesai, line 281: can we add some variation to the values?

rodesai (commented):
Nice tests! Left a couple more comments inline.

kafka-client/src/main/java/dev/responsive/kafka/bootstrap/ChangelogMigrationTool.java
```diff
@@ -0,0 +1,124 @@
...
+      final Properties properties,
+      final ResponsiveKeyValueParams params,
+      final Consumer<Record<byte[], byte[]>> processor
+  ) {
+    this.processor = processor;
+
+    properties.put(TOPOLOGY_OPTIMIZATION_CONFIG, REUSE_KTABLE_SOURCE_TOPICS);
+    properties.put(AUTO_OFFSET_RESET_CONFIG, "earliest");
+    properties.put(RESPONSIVE_MODE, ResponsiveMode.MIGRATE.name());
+
```
rodesai, line 72: lets set the write consistency to ALL

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/CommitBuffer.java
agavra: I reverted this from this patch since I no longer need it and I'm not 100% sure it was necessary

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
agavra: copy-pasta 😨

rodesai (approved):
LGTM!

PR #240, opened 2024-02-20
Segment Partitioner: Parametrize the key type

apourchet:
This will eventually allow us to reuse this segment partitioner for
session keys, in addition to the window keys that are currently
supported.


apourchet:
Closing in favor of #243

PR #248, opened 2024-02-22
Session Store: Main implementation

apourchet:


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/SessionMetadataDoc.java
```diff
@@ -0,0 +1,104 @@
...
+package dev.responsive.kafka.internal.db.mongo;
+
+import java.util.Objects;
+
+/**
+ * The 'id' of a session metadata doc corresponds to the kafka partition,
+ * which is why we don't have to include the partition as a separate
+ * field for the metadata docs the way we do for the data docs
+ */
+public class SessionMetadataDoc {
```
agavra, line 26: I think it might make sense to share this and `WindowMetadataDoc` for now - including all the code in `MongoSessionTable` related to metadata management

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
```diff
@@ -0,0 +1,112 @@
...
+  private static final long MINIMUM_ALLOWED_NUM_SEGMENTS = 2L;
+  private static final long MAXIMUM_ALLOWED_NUM_SEGMENTS = Long.MAX_VALUE;
+  private static final long MAXIMUM_DEFAULT_NUM_SEGMENTS = 100L;
+  private static final long DEFAULT_NUM_SEGMENTS = 10L; // space amplification = 1.1x
+
+  private final TableName name;
+  private final SessionSchema schemaType;
+  private final long inactivityGapMs;
+  private final long gracePeriodMs;
+  private final boolean retainDuplicates;
```
ableegoldman, line 42: Might have forgotten to mention this before --there is no "retain duplicates" for session stores. It's a special parameter of the WindowStore specifically. One could argue it's a bit of a hack, but basically the WindowStore gets reused by Kafka Streams for stream-stream joins, where it effectively treats the WindowStore as a regular KV-plus-timestamp store. But it has to be able to hold multple values for the same key-timestamp since you could have multiple events with the same key and timestamp in your join, thus Streams will set `retainDuplicates` to true for WindowStores that are to be used for stream-stream joins (as opposed to aggregations, which should have strict update semantics)

tl;dr we should remove all the retainDuplicates stuff for session stores

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
```diff
@@ -0,0 +1,112 @@
...
+  public boolean truncateChangelog() {
+    return this.truncateChangelog;
+  }
+
+  private static long maximumAllowedNumSegments(final long retentionPeriodMs) {
+    return Math.min(MAXIMUM_ALLOWED_NUM_SEGMENTS, retentionPeriodMs);
+  }
+
+  private static long computeDefaultNumSegments(final long retentionPeriodMs) {
+    // TODO: Smart implementation.
```
ableegoldman, line 109: Just an aside: I think it'll be a lot harder for us to come up with a "smart" choice for session stores. We can leave the TODO of course, but this is probably a case where we want to set a reasonable default and then in practice work with our users to figure out a reasonable value

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -62,7 +64,7 @@ public static KeyValueBytesStoreSupplier keyValueStore(final ResponsiveKeyValueP
    *
    * @param name the store name
    * @return a supplier for a key-value store with the given options
-   *         that uses Responsive's storage for its backend
+   * that uses Responsive's storage for its backend
```
ableegoldman, line 67: Is your IDE set to autoformat? I usually turn that off just because it can make an already-long PR even longer -- but it's not a big deal. To me, at least -- if you submit a PR to Kafka/Kafka Streams, some of the other committers _will_ complain about extraneous formatting changes. Even if they are an improvement lol. Just a heads up

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -295,5 +299,47 @@ public static <K, V> Materialized<K, V, WindowStore<Bytes, byte[]>> windowMateri
...
+   */
+  public static <K, V> StoreBuilder<SessionStore<K, V>> sessionStoreBuilder(
+      final SessionBytesStoreSupplier storeSupplier,
+      final Serde<K> keySerde,
+      final Serde<V> valueSerde
+  ) {
+    return new ResponsiveStoreBuilder<>(
+        Stores.sessionStoreBuilder(storeSupplier, keySerde, valueSerde),
+        false
```
ableegoldman, line 321: This should be set to the value of `truncateChangelog` that's in the session params

I'm pretty sure this was an existing bug though, we should probably fix that in a separate PR. We might want to be careful to ensure that Metronome sets this correctly -- @agavra do they still want to truncate the changelog? I'm a little hesitant after it saved us hahaha. Should we wait to fix this bug?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/MongoSessionFlushManager.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.function.Function;
+import org.apache.kafka.common.utils.LogContext;
+import org.slf4j.Logger;
+
+public class MongoSessionFlushManager extends SessionFlushManager {
+
+  private final String logPrefix;
+  private final Logger log;
+
+  private final MongoSessionTable table;
```
ableegoldman, line 40: Is this class basically the same as the MongoWindowFlushManager except for the type of this table (and the partitioner)? Just wondering, did you look into whether it's possible to consolidate them by parametrizing somehow?

My main concern with all this is: how easy would it be to make a change in one class but not the other, and accidentally screw up one of the implementations. It's not the end of the world to have repeated code, but what's the risk of code which should be identical getting out of sync?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/SessionDoc.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public void setValue(final byte[] value) {
+    this.value = value;
+  }
+
+  public void setEpoch(final long epoch) {
+    this.epoch = epoch;
+  }
+
+  public byte[] getValue() {
```
ableegoldman, line 76: QQ: are the getter names enforced by the bson format? If not, let's drop the `get` from this to be consistent with naming elsewhere in the code

(I know we're inconsistent in some of the other Mongo Doc classes right now, there was some confusion around what was required for bson objects. And clearly we still got some things wrong)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/SessionDoc.java
```diff
@@ -0,0 +1,137 @@
...
+public class SessionDoc {
+
+  public static final String ID = "_id";
+  public static final String VALUE = "value";
+  public static final String EPOCH = "epoch";
+  public static final String TOMBSTONE_TS = "tombstoneTs";
+
+  // Subfields of the composite key _id
+  public static final String ID_RECORD_KEY = "key";
+  public static final String ID_SESSION_END_TS = "sessionEndTs";
```
ableegoldman, line 36: nit: can we change the suffix from TS to MS? I know I myself used TS in the WindowDoc (and we should change that too) but it's pretty clearly a timestamp whereas MS at least tells us the units. And we tend to use `Ms` suffixes for similar variable names in Kafka Streams

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,300 @@
...
+    final SessionWindows window =
+        SessionWindows.ofInactivityGapAndGrace(inactivityGap, gracePeriod);
+
+    final Materialized<String, String, SessionStore<Bytes, byte[]>> store =
+        ResponsiveStores.sessionMaterialized(
+            ResponsiveSessionParams.session(name, inactivityGap, gracePeriod)
+        );
+
+    // Start from timestamp of 0L to get predictable results
+    final List<KeyValueTimestamp<String, String>> inputEvents = asList(
```
agavra, line 132: can we test things outside the inactivity gap and also inside the inactivity gap but requiring the grace period (i.e. out of order events)?

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,300 @@
...
+    properties.put(APPLICATION_SERVER_CONFIG, "host1:1024");
+    final KafkaProducer<String, String> producer = new KafkaProducer<>(properties);
+    try (
+        final ResponsiveKafkaStreams kafkaStreams =
+            new ResponsiveKafkaStreams(builder.build(), properties);
+    ) {
+      startAppAndAwaitRunning(Duration.ofSeconds(15), kafkaStreams);
+      pipeRecords(producer, inputTopic(), inputEvents);
+
+      assertThat(outputLatch.await(3_000, TimeUnit.MILLISECONDS), Matchers.equalTo(true));
```
agavra, line 172: 3s might be too fast for the janky CI machines -- consider bumping this up to something like 20s (since it's expected to pass fast, it's OK if the timeout is slow)

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,300 @@
...
+        var actual = actualPeeks.get(i);
+        var expected = expectedPeeks.get(i);
+        assertThat(actual.key, Matchers.equalTo(expected.key));
+        assertThat(actual.value, Matchers.equalTo(expected.value));
+      }
+    }
+  }
+
+  @Test
+  public void shouldComputeSessionAggregateWithMultipleKeys() throws Exception {
```
agavra, line 185: any reason not to just merge this with the first test? (e.g. does it use different functionality)

kafka-client/src/test/java/dev/responsive/kafka/internal/db/SessionKeySpecTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.kafka.streams.state.internals.SessionKeySchema;
+import org.junit.jupiter.api.Test;
+
+public class SessionKeySpecTest {
+
+  private static final Random RANDOM = new Random();
+  private static final SessionKeySpec keySpec = new SessionKeySpec(w -> w.sessionEndMs > 0L);
+
+  @Test
+  public void verifySessionKeySpecEncoding() {
```
agavra, line 35: please use given,when,then and name `shouldXYZ` 😉 the one nit I'm passionate about haha

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTableTest.java
```diff
@@ -0,0 +1,158 @@
...
+      @ResponsiveConfigParam final Map<String, Object> props
+  ) {
+    name = info.getDisplayName().replace("()", "");
+
+    final String mongoConnection = (String) props.get(STORAGE_HOSTNAME_CONFIG);
+    client = SessionUtil.connect(mongoConnection, null, null);
+  }
+
+  @Test
+  public void shouldSucceedSimpleSetGet() {
```
agavra, line 65: are there any meaningful negative tests we should test? (i.e. throwing exceptions)

also suggestion - we should consider splitting out: `shouldFindInsertedKeyWithValidStartTimestamp`, `shouldNotFindInsertedKeyWithInvalidStartTimestamp` and `shouldNotFindKeyThatWasNotInserted`

That way it's very clear what each of the tests are doing and when they fail the test name says it all

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTableTest.java
```diff
@@ -0,0 +1,158 @@
...
+        DEFAULT_VALUE,
+        table.localEpoch(0)
+    );
+    writer.flush();
+
+    // Then:
+    var it = table.fetchAll(0, byteKey, 100, 100);
+    var kvs = new ArrayList<KeyValue<SessionKey, byte[]>>();
+    it.forEachRemaining(kvs::add);
+    assertThat(kvs, Matchers.hasSize(1));
```
agavra, line 154: feel like we should test this with more than one result!

kafka-client/src/test/java/dev/responsive/kafka/testutils/IntegrationTestUtils.java
```diff
@@ -336,6 +338,21 @@ public static void startAppAndAwaitRunning(
     }
   }
 
+  public static Merger<String, String> sessionMerger() {
```
agavra, line 341: personal preference, but I prefer to have these in the test file directly so I don't need to read two files (especially when reviewing) to see what a test is doing especially since it's important to the output

agavra (commented):
I haven't actually thought at all about the most important class yet (`MongoSessionTable`) but figured I'd get all the minor comments out of the way first. Most of them are about code duplication and seeing if we can reduce it a bit.

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
agavra: I'm definitely supportive of not fixing it in this PR. I do think we should consider removing changelog truncation altogether.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/MongoSessionFlushManager.java
apourchet: You're right, it is almost entirely a copy paste of the MongoWindowFlushManager class, which itself is a relatively close replica of the MongoKVFlushManager. My plan was to refactor these 3 into 1 separately in another PR so as to keep this one entirely devoid of any changes impacting other stores (KV / Window).

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTableTest.java
apourchet: No negative test that I can think of right now, I wrote those tests as mirrors of the ones for the KV and Window stores but we might want to expand all of them in similar ways (to include negative tests as you mentioned).

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/SessionMetadataDoc.java
apourchet: I agree that this needs to be consolidated. I wrote this down and will address in a follow-up PR.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,494 @@
...
+        );
+      }
+      segmentSessions.put(segmentToCreate, sessionDocs);
+
+      // TODO(agavra): make the tombstone retention configurable
+      // this is idempotent
+      sessionDocs.createIndex(
+          Indexes.descending(SessionDoc.TOMBSTONE_TS),
+          new IndexOptions().expireAfter(12L, TimeUnit.HOURS)
+      );
```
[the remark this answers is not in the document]
ableegoldman: Actually yes, they are deleted as part of the regular session window aggregation. While session stores are similar to window stores but with an additional timestamp, how they are used (in the DSL operators) is quite different. 

Window stores are essentially simple read-update-write -- session stores are more like rangeScan-maybeMerge-maybeDelete

Deletes don't necessarily happen on every record but they are a foundation of the session store

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
ableegoldman: Just to build off of this -- given the startup time for our integration tests, it often makes sense to throw a bit of "good testing practice" out the window and try to do as much as possible within a single test. If it's possible to combine tests without sacrificing the coverage, we should do so.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,494 @@
...
+          // Most likely if we have active segments upon initialization, the corresponding
+          // collections and windowStart indices already exist, in which case #getCollection and
+          // #createIndex should be quick and won't rebuild either structure if one already exists
+          createSegment(segmentToCreate);
+        }
+
+        final long firstSegmentId = activeSegments.get(0).segmentStartTimestamp;
+        LOG.info("{}[{}] Initialized active segments in range {} - {}",
+                 database.getName(), kafkaPartition, firstSegmentId,
+                 firstSegmentId + activeSegments.size());
```
ableegoldman, line 127: This comment doesn't quite make sense since we switched from segment ids to start timestamps. I think you can just replace this line with `activeSegments.get(activeSegments.length() - 1)).segmentStartTimestamp

(Probably also makes sense to just inline `activeSegments.get(0).segmentStartTimestamp` and remove the `firstSegmentId` variable too)

ableegoldman: Can you do a quick check to see if this needs to be updated in the MongoWindowTable as well? This feels a bit familiar so I'm wondering if maybe this was just based on the older version of the window table code before your other PR changed it

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,494 @@
...
+      final long segmentStartTimeMs =
+          segmentPartition.segmentStartTimestamp * partitioner.segmenter().segmentIntervalMs();
+      return String.format(
+          "%d-%d",
+          segmentPartition.tablePartition, segmentStartTimeMs
+      );
+    }
+
+    private void createSegment(final Segmenter.SegmentPartition segmentToCreate) {
+      LOG.info("{}[{}] Creating segment id {}",
```
ableegoldman, line 141: nit: same here, should be segmentStartTimestamp -- I'm guessing this is just a holdover from before your last PR, so maybe just give that a quick pass to see what changed in MongoWindowTable during the segmentId --> segmentStartTimestamp update and now needs to be reflected here too?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,494 @@
...
+        );
+      } else {
+        sessionDocs = database.getCollection(
+            collectionNameForSegment(segmentToCreate),
+            SessionDoc.class
+        );
+      }
+      segmentSessions.put(segmentToCreate, sessionDocs);
+
+      // TODO(agavra): make the tombstone retention configurable
```
ableegoldman, line 163: I'm still a bit hesitant about whether we want to support it at all for window stores, but as noted already, we clearly need to for KV stores and session stores.

However I don't think there's any reasonable need to configure it per store. Maybe we can just throw in a global config for tombstone retention and call it a day? (Not in this PR of course)  cc @agavra

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,263 @@
...
+        .toStream()
+        .peek((k, v) -> actualPeeks.add(new KeyValue<>(k, v)))
+        .peek((k, v) -> outputLatch.countDown())
+        .selectKey((k, v) -> k.key())
+        .to(outputTopic());
+
+    // When:
+    final Map<String, Object> properties = getMutablePropertiesWithStringSerdes();
+    properties.put(STATESTORE_CACHE_MAX_BYTES_CONFIG, 0);
+    properties.put(APPLICATION_SERVER_CONFIG, "host1:1024");
```
ableegoldman, line 171: I know you just copied this over so it's probably @agavra who I should aim this comment at, but -- just FYI, the `APPLICATION_SERVER_CONFIG` is really only for IQ and giving instances a name that the user can recognize/control. In other words, we definitely don't need to set it in any of our functional tests (IQ == Interactive Queries, a Streams feature for reading the state store contents from outside the app)

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,263 @@
...
+      startAppAndAwaitRunning(Duration.ofSeconds(15), kafkaStreams);
+      pipeRecords(producer, inputTopic(), inputEvents);
+
+      assertThat(
+          outputLatch.await(20_000, TimeUnit.MILLISECONDS),
+          Matchers.equalTo(true)
+      );
+
+      // Thread.sleep(5_000);
+      // actualPeeks.forEach(kv -> System.out.println("actual peek: " + kv));
```
ableegoldman, line 186: Guessing we can remove this 😄

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,263 @@
...
+        new KeyValueTimestamp<>("key", "c", 7_000L),
+        new KeyValueTimestamp<>("key1", "d", 8_000L),
+        new KeyValueTimestamp<>("key1", "e", 16_000L),
+        new KeyValueTimestamp<>("key1", "f", 12_000L),
+        new KeyValueTimestamp<>("key1", "g", 7_500L)
+    );
+    final List<KeyValue<Windowed<String>, String>> expectedPeeks = List.of(
+        new KeyValue<>(windowedKey("key", 0, 0), "a"),
+        new KeyValue<>(windowedKey("key", 0, 0), null),
+        new KeyValue<>(windowedKey("key", 0, 6000), "ab"),
```
ableegoldman, line 144: I'm confused...the inactivity gap is 5000, right? So if a is at 0 and b is at 6000, then there's more than 5000ms between these two events, and therefore they should not be merged into the same session...right?

It's been a long time since I thought about session aggregations so it's completely possible I'm not looking at this the right way. But my understanding is that  events separated by more than the inactivity gap should stay separate, at least unless/until another event comes in between them and is within the inactivity gap of both. In that case you get could a and b merged with each other -- if, say, c had timestamp 3000 for example, we would see "a" and "b" merged into "abc". But I think (not 100% sure) that we'd never see just an "ab" output

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SessionOperationsImpl.java
```diff
@@ -0,0 +1,285 @@
...
+        key.sessionEndMs
+    );
+    return remoteResult;
+  }
+
+  @Override
+  public KeyValueIterator<Windowed<Bytes>, byte[]> fetchAll(
+      final Bytes key,
+      final long earliestSessionEnd,
+      final long latestSessionEnd
```
ableegoldman, line 244: parameter name should be latestSessionStart

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SessionOperationsImpl.java
```diff
@@ -0,0 +1,285 @@
...
+  ) {
+    final SessionKey from = new SessionKey(key, 0, earliestSessionEnd);
+    final SessionKey to = new SessionKey(key, 0, latestSessionEnd);
+
+    final var localResults = this.buffer.all(kv -> {
+      return kv.key.key.equals(key) && kv.key.sessionEndMs >= earliestSessionEnd
+          && kv.key.sessionEndMs <= latestSessionEnd;
```
ableegoldman, line 251:
```suggestion
          && kv.key.sessionStartMs <= latestSessionEnd;
```

I'm wondering if we should have a separate integration test where we _don't_ set the CommitBuffer to flush on every record -- sadly the soak won't catch things like this since we don't do any correctness checking at the moment (of course eventually we hope to use your new store-comparison tool for that). It's still probably worth doing an integration test.

Maybe to save time we could knock out all three store types in the same integration test with less-frequent buffer flushing (rather than having an extra test for this in each store type's integration test class)? This would be followup work of course, just thinking out loud. For now, maybe just try running the test once with the buffer flush config commented out, to make sure the results line up (but merging it with the buffer flush config re-enabled)

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -0,0 +1,261 @@
...
+    properties.put(DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.LongSerde.class.getName());
+    properties.put(NUM_STREAM_THREADS_CONFIG, 1);
+    properties.put(COMMIT_INTERVAL_MS_CONFIG, 1); // commit as often as possible
+
+    properties.put(consumerPrefix(REQUEST_TIMEOUT_MS_CONFIG), 5_000);
+    properties.put(consumerPrefix(SESSION_TIMEOUT_MS_CONFIG), 5_000 - 1);
+
+    properties.put(consumerPrefix(MAX_POLL_RECORDS_CONFIG), 1);
+
+    properties.put(ResponsiveConfig.STORE_FLUSH_RECORDS_TRIGGER_CONFIG, 1);
```
ableegoldman, line 230: This is the config I was referring to above, just fyi -- worth running just once with this disabled

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTableTest.java
ableegoldman: Expanding the basic test coverage (ie unit tests) of all the stores has been on my TODO list for a while. It's hard to make time to prioritize writing tests for things that seem to be working when there is so much else to do -- but we don't want to wait until it's too late to catch bugs. 

Perhaps we could dedicate something like one day each month towards "bug bashing and bug prevention" -- including filling in holes in our test coverage. I think it would be pretty valuable. Even a half day a month could pay off if all of us do it together

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,503 @@
...
+          segmentPartition.segmentStartTimestamp * segmenter.segmentIntervalMs();
+      return String.format(
+          "%d-%d",
+          segmentPartition.tablePartition, segmentStartTimeMs
+      );
+    }
+
+    private void createSegment(final Segmenter.SegmentPartition segmentToCreate) {
+      LOG.info(
+          "{}[{}] Creating segment start timestamp {}",
```
ableegoldman, line 148:
```suggestion
          "{}[{}] Creating segment with start timestamp {}",
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,503 @@
...
+    // stream-time such that the segment is already expired
+    // However this does leave us vulnerable to zombie segments. We should consider doing a sweep
+    // for old segments every so often with #listCollectionNames, since the segmentId is embedded in
+    // the collection name and will indicate whether it's part of the active set or not, perhaps
+    // using a scheduled executor to avoid blocking processing
+    // In fact we may want to move all the segment expiration to a background process since there's
+    // no async way to drop a collection but the stream thread doesn't actually need to wait for it
+    private void deleteSegment(final Segmenter.SegmentPartition segmentToExpire) {
+      LOG.info(
+          "{}[{}] Expiring segment start timestamp {}",
```
ableegoldman, line 190:
```suggestion
          "{}[{}] Expiring segment with start timestamp {}",
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -0,0 +1,503 @@
...
+        .first();
+
+    return sessionDoc == null ? null : sessionDoc.value();
+  }
+
+  public KeyValueIterator<SessionKey, byte[]> fetchAll(
+      final int kafkaPartition,
+      final Bytes key,
+      final long earliestSessionEnd,
+      final long latestSessionEnd
```
ableegoldman, line 372:
```suggestion
      final long latestSessionStart
```

ditto elsewhere this appears

ableegoldman (approved):
A bunch of minor things (eg parameter name updates) but all in all this LGTM.

I don't think there was anything that would need another pass so feel free to merge when you've addressed all the comments (obviously excluding the ones about followup work outside of this PR). I'm also happy to take another look at anything you want me to do another pass over before merging, as usual just give me a ping if so

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
apourchet: I am actually using earliest / latest session end to filter the remote table and find the sessions.

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SessionOperationsImpl.java
apourchet: Same as for the RemoteSessionTable interface, I only filter based on session end, so I pass in [earliest, latest] as a way to lower + upper bound the session end.
The latestSessionStart timestamp is only used as a last-minute filter on the results from filtering the session end.

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveSessionStore.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @Override
+  public void close() {
+    this.sessionOperations.close();
+  }
+
+  @Override
+  public void put(final Windowed<Bytes> session, final byte[] aggregator) {
+    // Write the aggregated value for the provided key to the store.
```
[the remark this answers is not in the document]
apourchet: I think this is left-over from when I pasted the comments from the javadocs. They're not good comments to have in the first place so I will remove them.

PR #250, opened 2024-02-26
update offsets on flush even when buffer is empty

agavra:
run `preFlush` and `postFlush` even if the actual data is empty - this only makes a difference in the bootstrapping tool where we filter TTL'd data as we process it


ableegoldman (approved):
LGTM

ableegoldman:
Checkstyle issue in test though

agavra:
> Checkstyle issue in test though

grumble grumble grumble... 😆

PR #251, opened 2024-02-26
Mongo range queries: setup and KVTable implementations

ableegoldman:
This PR includes the basic setup for our approach to range queries via client-side filtering based on a partitioner, as well as the implementation for KV stores. The window range queries are in a separate PR since they're a bit more involved.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -18,6 +18,7 @@ package dev.responsive.kafka.internal.db.mongo;
 
 import static com.mongodb.MongoClientSettings.getDefaultCodecRegistry;
 import static dev.responsive.kafka.internal.stores.ResponsiveStoreRegistration.NO_COMMITTED_OFFSET;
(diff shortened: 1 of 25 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -57,17 +63,39 @@ public class MongoKVTable implements RemoteKVTable<WriteModel<KVDoc>> {
...
   private final ConcurrentMap<Integer, Long> kafkaPartitionToEpoch = new ConcurrentHashMap<>();
 
+  public static MongoKVTable create(
+      final MongoClient client,
+      final String name,
+      final String changelogTopicName,
+      final Admin admin,
+      final CollectionCreationOptions collectionCreationOptions
+  ) {
+    final int numPartitions = numPartitionsForKafkaTopic(admin, changelogTopicName);
```
agavra, line 81: piping the changelog topic name and the admin all the way down here just to get the number of partitions (which is then used to build a partitioner) seems to be leaking a lot of unnecessary information

instead I suggest that we put the partitioner itself in the `RemoteTableSpec` interface and pass the partitioner directly into this.

ableegoldman:
closing this since it was implemented in https://github.com/responsivedev/responsive-pub/pull/347

PR #252, opened 2024-02-27
Windowed Table: Unit testing

apourchet:
Fixed the BSON parsing bug that was hidden by the KS state store caching mechanism. Without the proper @BsonCreator constructor, our document retrieval and parsing from MongoDB were never getting their ID fields populated correctly.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
@@ -23,6 +23,8 @@ import dev.responsive.kafka.internal.utils.WindowedKey;
 import java.util.Arrays;
 import java.util.Objects;
 import org.apache.kafka.common.utils.Bytes;
+import org.bson.codecs.pojo.annotations.BsonCreator;
+import org.bson.codecs.pojo.annotations.BsonProperty;
 
 public class WindowDoc {
 
@@ -39,6 +41,17 @@ public class WindowDoc {
   long epoch;
 
   public WindowDoc() {
+  }
+
+  @BsonCreator
+  public WindowDoc(
+      @BsonProperty(ID) BasicDBObject id,
+      @BsonProperty(VALUE) byte[] value,
+      @BsonProperty(EPOCH) long epoch
+  ) {
+    this.id = id;
+    this.value = value;
+    this.epoch = epoch;
   }
 
   public BasicDBObject getKey() {
--- a/kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoWindowTableTest.java
+++ b/kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoWindowTableTest.java
@@ -1,0 +1,74 @@
+package dev.responsive.kafka.internal.db.mongo;
+
+import static dev.responsive.kafka.api.config.ResponsiveConfig.STORAGE_HOSTNAME_CONFIG;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import com.mongodb.client.MongoClient;
+import dev.responsive.kafka.api.config.StorageBackend;
+import dev.responsive.kafka.internal.db.partitioning.WindowSegmentPartitioner;
+import dev.responsive.kafka.internal.utils.SessionUtil;
+import dev.responsive.kafka.internal.utils.WindowedKey;
+import dev.responsive.kafka.testutils.ResponsiveConfigParam;
+import dev.responsive.kafka.testutils.ResponsiveExtension;
+import java.util.Map;
+import org.apache.kafka.common.utils.Bytes;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+class MongoWindowedTableTest {
+
+  @RegisterExtension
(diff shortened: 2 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
```diff
@@ -41,6 +43,17 @@ public class WindowDoc {
   public WindowDoc() {
   }
 
+  @BsonCreator
```
agavra, line 46: do we need to do the same for `KVDoc`? I can't remember if you said that's going to be a separate PR or not. Also for `KVDoc` I noticed there's `@BsonId` annotation on the `id` field, is that necessary/needed?

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoWindowTableTest.java
```diff
@@ -0,0 +1,129 @@
...
+        table.localEpoch(0)
+    );
+    writer.flush();
+
+    // Then:
+    var it = table.fetch(0, byteKey, 100, 600);
+    var kvs = new ArrayList<KeyValue<WindowedKey, byte[]>>();
+    it.forEachRemaining(kvs::add);
+
+    assertThat(kvs, Matchers.hasSize(1));
```
agavra, line 124: nit: since we're trying to fetch a range, can we verify that multiple values can be returned instead of just one?

I know that's unrelated to this PR, but since this is somewhat of an integration test I think it's worth covering that part as well

agavra (approved):
Nice fix! Thanks @apourchet

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoWindowTableTest.java
```diff
@@ -0,0 +1,129 @@
...
+    assertThat(kvs.get(0).key.key, Matchers.equalTo(windowedKey.key));
```
agavra, line 125: was this the thing that was failing on `main`?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
apourchet: I did some testing on KVDoc independently and it looks like `@BsonId` isn't doing what we want it to. I wanted to see what the feedback was on this Window stuff before applying the same logic to the `KVDoc` (since it's already in use in production).

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoWindowTableTest.java
apourchet: Yep, it's the range test that's failing on main. The reason behind the failure is that the WindowKey was being populated using the `.id` field, which throws NullPointer as it's never populated by the Bson decoder.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
ableegoldman: note: none of the Mongo stuff is in production yet, the KV store that we have in prod is based on Scylla. So now is the time to make any changes we want/need

ableegoldman (approved):
LGTM -- still curious why this bug surfaced for session stores but not the others. I did check and our integration tests do a `flush` on every record, so (un?)fortunately I don't think the CommitBuffer serving all the records in the tests is the answer.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KVDoc.java
```diff
@@ -38,7 +38,13 @@ public class KVDoc {
...
   }
 
-  public KVDoc(final byte[] key, final byte[] value, final long epoch) {
+  @BsonCreator
+  public KVDoc(
+      @BsonProperty(ID) byte[] id,
+      @BsonProperty(VALUE) byte[] value,
+      @BsonProperty(EPOCH) long epoch
+  ) {
+    this.id = id;
```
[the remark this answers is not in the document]
apourchet: I just tested this locally, which goes back to failing the new test:
```
public class WindowDoc {

@@ -43,11 +41,10 @@ public class WindowDoc {
   public WindowDoc() {
   }

-  @BsonCreator
   public WindowDoc(
-      @BsonProperty(ID) BasicDBObject id,
-      @BsonProperty(VALUE) byte[] value,
-      @BsonProperty(EPOCH) long epoch
+      BasicDBObject id,
+      byte[] value,
+      long epoch
   ) {
     this.id = id;
     this.value = value;
```
So it looks like the Bson decoder needs the actual annotations and not just the right constructor signature.

apourchet:
> LGTM -- still curious why this bug surfaced for session stores but not the others. I did check and our integration tests do a `flush` on every record, so (un?)fortunately I don't think the CommitBuffer serving all the records in the tests is the answer.

I believe the reason we didn't catch this earlier is that none of the integration tests call out to the `MongoWindowedTable` range queries, which means we are never creating `WindowedKey` from the raw `WindowDoc`. More specifically, this line is never hit:
https://github.com/responsivedev/responsive-pub/blob/main/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java#L451
```
      segmentIterators.add(
          Iterators.kv(fetchResults.iterator(), MongoWindowedTable::windowFromDoc)
      );
```

PR #253, opened 2024-02-28
add support to repair offsets on restore

agavra:
`responsive.restore.offset.repair.enabled` allows users to restore from remote committed offsets that have fallen off retention for whatever reason, the behavior will instead seek to beginning

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
@@ -231,6 +231,15 @@ public class ResponsiveConfig extends AbstractConfig {
       + "runs in. When set to RUN, runs the Kafka Streams app. When set to MIGRATE, runs app"
       + " migration.";
 
+  // ------------------ Misc functional overrides ----------------------
+  public static final String RESTORE_OFFSET_REPAIR_ENABLED_CONFIG = "responsive.restore.offset.repair.enabled";
+  public static final boolean RESTORE_OFFSET_REPAIR_ENABLED_DEFAULT = false;
+  public static final String RESTORE_OFFSET_REPAIR_ENABLED_DOC = "When set to 'true', " + RESTORE_OFFSET_REPAIR_ENABLED_CONFIG
+      + " will ignore OffsetOutOfRangeException and instead seek to the earliest available offset. This exception "
+      + "should only happen in situations where there is truncation/retention on the changelog topic and restoring from the latest "
+      + "committed offset in the remote store is no longer possible. Note that in some situations this may cause data "
+      + "loss, use this configuration with caution";
+
   // ------------------ StreamsConfig overrides ----------------------
 
   // These configuration values are required by Responsive, and a ConfigException will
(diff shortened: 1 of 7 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -231,6 +231,15 @@ public class ResponsiveConfig extends AbstractConfig {
       + "runs in. When set to RUN, runs the Kafka Streams app. When set to MIGRATE, runs app"
       + " migration.";
 
+  // ------------------ Misc functional overrides ----------------------
+  public static final String RESTORE_OFFSET_REPAIR_ENABLED_CONFIG = "responsive.restore.offset.repair.enabled";
+  public static final boolean RESTORE_OFFSET_REPAIR_ENABLED_DEFAULT = false;
+  public static final String RESTORE_OFFSET_REPAIR_ENABLED_DOC = "When set to 'true', " + RESTORE_OFFSET_REPAIR_ENABLED_CONFIG
+      + " will ignore OffsetOutOfRangeException and instead seek to the earliest available offset. This exception "
+      + "should only happen in situations where there is truncation/retention on the changelog topic and restoring from the latest "
+      + "committed offset in the remote store is no longer possible. Note that in some situations this may cause data "
```
rodesai, line 240: I'm still not really understanding where it can actually cause data loss.

It depends on what the source of truth is I guess. If the changelog is the source of truth, it should not be truncated by responsive, and any truncation in the changelog is like deleting that data from the store so its safe to skip.

If we are treating the remote store as the source of truth, then responsive could truncate the changelog and we should be more careful.

Today, we usually operate in the first mode. I think over time we want to move to the second mode. I wonder if it makes sense to make that distinction explicit as a config. Not really a comment on this PR in particular, but more spitballing - maybe we should have the config be between these two modes of operation and then decide to skip or not skip accordingly.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveRestoreConsumer.java
```diff
@@ -101,7 +106,23 @@ public ConsumerRecords<K, V> poll(final Duration timeout) {
       log.error("Found uninitialized changelog partitions during poll: {}", uninitializedOffsets);
       throw new IllegalStateException("Restore consumer invoked poll without initializing offsets");
     }
-    return super.poll(timeout);
+
+    try {
+      return super.poll(timeout);
+    } catch (final OffsetOutOfRangeException e) {
+      // e.partitions() should never be empty, but we check it because accidentally
+      // passing in empty partitions is catastrophic (it will seek all partitions to
```
rodesai, line 114: yikes. nice catch

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveRestoreConsumer.java
```diff
@@ -101,7 +106,23 @@ public ConsumerRecords<K, V> poll(final Duration timeout) {
...
+      // beginning)
+      if (repairRestoreOffsetOutOfRange && !e.partitions().isEmpty()) {
+        log.warn("The restore consumer attempted to seek to offsets that are no longer in range. "
```
rodesai, line 117: if we log `offsetOutOfRangePartitions()` we'll get the fetched offsets in the log too

rodesai (approved):
LGTM, couple thoughts/questions inline

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
agavra: Let's chat about it more offline. I agree with your logic, but something still makes me anxious.

PR #255, opened 2024-03-02
Store Comparator: Created a StoreComparator class

apourchet:
This new class allows us to check for behavior parity accross multiple stores. Essentially, it exposes a single StoreSupplier interface that proxies the function calls to both stores and checks that the result of those function calls are identical in both cases.


kafka-client/src/test/java/dev/responsive/kafka/testutils/StoreComparator.java
```diff
@@ -0,0 +1,225 @@
...
+import org.apache.kafka.streams.processor.CommitCallback;
+import org.apache.kafka.streams.processor.StateRestoreCallback;
+import org.apache.kafka.streams.processor.StateStore;
+import org.apache.kafka.streams.processor.StateStoreContext;
+import org.apache.kafka.streams.state.KeyValueBytesStoreSupplier;
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.SessionStore;
+import org.apache.kafka.streams.state.StoreSupplier;
+
+public class StoreComparator {
```
rodesai, line 34: nit: maybe rename to `StoreComparators` or `StoreComparatorSuppliers` or something to indicate its not actually a comparator itself

kafka-client/src/test/java/dev/responsive/kafka/testutils/StoreComparator.java
```diff
@@ -0,0 +1,225 @@
...
+      }
+      return new SessionStoreComparator<>(
+          this.sourceOfTruth.get(),
+          this.candidate.get(),
+          this.compare
+      );
+    }
+  }
+
+  public static class MultiStateStoreContext implements InvocationHandler, StateRestoreCallback,
```
rodesai, line 135: nit: this feels like it should be its own class - why's it nested here alongside the factory methods?

rodesai (approved):
Took a quick pass over this. LGTM! This will be super useful

PR #256, opened 2024-03-04
add host.name to otel metrics

agavra:
Without this the `host.name` resource tag was not populated by the OTEL metrics exporter. This means that anything that aggregates by host would incorrectly aggregate metrics for every host by aggregating metrics for all hosts.

ex: if we had two hosts that were reporting 10 records/s, the metrics would consider both of them to be processing 20 records/s


apourchet (approved):

rodesai (approved):
LGTM. We should also do a controller change to, when evaluating policies, throw a validation error if this is not set

PR #260, opened 2024-03-05
Integration Tests: Fix session store integration test

apourchet:
Also locally tested with infinite commit buffer.

### DIFF
--- a/kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
+++ b/kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
@@ -179,7 +179,7 @@ public class ResponsiveSessionStoreIntegrationTest {
       startAppAndAwaitRunning(Duration.ofSeconds(15), kafkaStreams);
       pipeRecords(producer, inputTopic(), inputEvents);
 
-      final boolean awaited = outputLatch.await(20_000, TimeUnit.MILLISECONDS);
+      final boolean awaited = outputLatch.await(25_000, TimeUnit.MILLISECONDS);
       assertThat(
           String.format(
               "The application did not receive the expected number of peeks: %d / %d\n%s\nVS\n\n%s",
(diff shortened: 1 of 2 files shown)


kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -180,31 +179,13 @@ public void shouldComputeSessionAggregate() throws Exception {
...
-        final ResponsiveKafkaStreams kafkaStreams =
-            new ResponsiveKafkaStreams(builder.build(), properties);
-    ) {
-      startAppAndAwaitRunning(Duration.ofSeconds(15), kafkaStreams);
-      pipeRecords(producer, inputTopic(), inputEvents);
-
-      assertThat(
-          outputLatch.await(20_000, TimeUnit.MILLISECONDS),
+          String.format(
+              "The application did not receive the expected number of peeks: %d / %d",
```
agavra, line 185: let's just log all the peeks instead of just the number - it's not that many events :) that will tell us whether we're missing something in the middle vs something in the end

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
```diff
@@ -172,6 +170,7 @@ public void shouldComputeSessionAggregate() throws Exception {
     // When:
     final Map<String, Object> properties = getMutablePropertiesWithStringSerdes();
     properties.put(STATESTORE_CACHE_MAX_BYTES_CONFIG, 0);
+    properties.put(ResponsiveConfig.STORE_FLUSH_RECORDS_TRIGGER_CONFIG, 1);
```
agavra, line 173: nit: any reason not to just put these in `getMutableProperites`?

agavra (approved):

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveSessionStoreIntegrationTest.java
apourchet: Along with the statestore cache config, the stretch goal is that we should run these tests with multiple configurations for each of these two settings. So I wanted to have those two properties split from the rest.

PR #261, opened 2024-03-06
lookup stores by thread id in addition to topic-partition

rodesai:


kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -226,6 +227,20 @@ private String threadIdFromProducerConfig(final String clientId) {
     return match.group(1);
   }
 
+  /**
+   * @param clientId the restore consumer client id
+   * @return the extracted StreamThread id, of the form "StreamThread-n"
+   */
+  private String threadIdFromRestoreConsumerConfig(final String clientId) {
```
agavra, line 234: why not just use the entire `clientId`?

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/StoreCommitListener.java
```diff
@@ -27,14 +28,14 @@ private void onCommit(
     for (final var e : committedOffsets.entrySet()) {
       final TopicPartition p = e.getKey().getPartition();
       for (final ResponsiveStoreRegistration storeRegistration
-          : storeRegistry.getRegisteredStoresForChangelog(p)) {
+          : storeRegistry.getRegisteredStoresForChangelog(p, threadId)) {
```
agavra, line 31: unrelated, but do you mind adding a comment about why we do this twice (once for committed offsets and once for written offsets)? it's not obviously clear to me what's the difference

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StoreUtil.java
```diff
@@ -92,6 +93,17 @@ public static long durationToMillis(final Duration duration, final String parame
     }
   }
 
+  public static String streamThreadId() {
+    final String threadId = Thread.currentThread().getName();
+    final var regex = Pattern.compile(".*-(StreamThread-\\d+)");
```
agavra, line 98: ditto as above, why not just use the entirety of the thread name?

agavra (approved):
Awesome sleuthing @rodesai! Thanks :)

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StoreUtil.java
```diff
@@ -92,6 +93,17 @@ public static long durationToMillis(final Duration duration, final String parame
     }
   }
 
+  public static String streamThreadId() {
```
agavra, line 96: Looks like we already have `ResponsiveMetrics#computeThreadId` - may want to consider reusing that (it also addresses Global store threads)

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
rodesai: We need to be able to look up the store from all the different clients, so thread id is the finest granularity that's common to them all.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/StoreCommitListener.java
rodesai: one path gets triggered for changelogs, the other for source topics (when the source table optimization is on)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveStoreRegistry.java
```diff
@@ -41,10 +43,12 @@ public synchronized OptionalLong getCommittedOffset(final TopicPartition topicPa
   }
 
   public synchronized List<ResponsiveStoreRegistration> getRegisteredStoresForChangelog(
-      final TopicPartition topicPartition
+      final TopicPartition topicPartition,
+      final String threadId
   ) {
     return stores.stream()
-        .filter(s -> s.changelogTopicPartition().equals(topicPartition))
+        .filter(s -> s.changelogTopicPartition().equals(topicPartition)
```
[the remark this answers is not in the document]
rodesai: no - I'm going to add a couple checks like this to this patch to tighten things up a bit. The other thing I'll check is that the offset passed to flush is greater than the last offset the store instance was aware of

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StoreUtil.java
rodesai: yeah I guess we could. It's a bit more of a change and it was late and I was tired :)

PR #264, opened 2024-03-07
Integration Test: Fix session store integration test and implementation

apourchet:


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -380,18 +381,24 @@ public KeyValueIterator<SessionKey, byte[]> fetchAll(
     );
 
     final var minKey = new SessionKey(key, 0, earliestSessionEnd);
-    final var maxKey = new SessionKey(key, 0, latestSessionEnd);
+    final var maxKey = new SessionKey(key, Long.MAX_VALUE, latestSessionEnd);
```
ableegoldman, line 384: This is the bug, I assume...I figured it would be something silly like this. Sorry for not catching it during review 

But totally awesome that you caught this in the test. You've already picked up on so many testing gaps and started to improve the testing framework 🙏

ableegoldman (approved):
Do I understand correctly that the reason this test was flaky (vs failing all the time) was due to the CommitBuffer hiding the issue at times by serving the results correctly?

apourchet:
> Do I understand correctly that the reason this test was flaky (vs failing all the time) was due to the CommitBuffer hiding the issue at times by serving the results correctly?

Yep that's exactly it, the bugs all along were in the RemoteSessionTable implementation for Mongo.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoSessionTable.java
```diff
@@ -380,18 +381,24 @@ public KeyValueIterator<SessionKey, byte[]> fetchAll(
...
         continue;
       }
 
-      final FindIterable<SessionDoc> fetchResults = sessionsSegment.find(
+      final FindIterable<SessionDoc> fetchResults = collection.find(
           Filters.and(
               Filters.gte(SessionDoc.ID, compositeKey(minKey)),
-              Filters.lte(SessionDoc.ID, compositeKey(maxKey))
+              Filters.lte(SessionDoc.ID, compositeKey(maxKey)),
+              Filters.exists(SessionDoc.TOMBSTONE_TS, false)
```
apourchet, line 401: This was the second bug.

PR #265, opened 2024-03-11
WIP: Async Processing POC

ableegoldman:
Opening a WIP PR so folks can follow along, check things out on their own time. This is not fully complete

::::: Reviewing Help :::::

- See "Current limitations" in the javadocs of AsyncProcessorSupplier
- See "Threading Notes" in class/method javadocs for clarity on usage and thread safety measures
- Start with AsyncProcessor for implementation details
- Each category of record has a queue lifecycle, record class, and specific generic type naming:
  - Note: in practice they tend to have the same type, but in theory all three can be different types -- eg KIn vs KOut vs KS -- and there might even be multiple "KS" of different types
  - "Input" records:
	- lifecycle is SchedulingQueue --> ProcessingQueue
	- type is KIn/VIn
	- use the terms "scheduling" and "processing" to describe their execution:
		- "to schedule" means to pass a processable record from the StreamThread (SchedulingQueue) to the 
		    AsyncThreadPool (ProcessingQueue)
		- "to process" means to pick up a record from the ProcessingQueue and execute the user's #process on it 
		    from the AsyncThread
		- the adjective "processable" means an input record waiting to be scheduled (see below) that has satisfied all 
		   the conditions such as having completed all previous records with the same key
  - "Output" records:
	- output records are placed into ForwardingQueue
	- type is KOut/VOut
	- use the term "forwarding" to describe their execution, ie forwarding to the original StreamThread context
  - "Stored" records:
	- state store insertion records are placed into the WritingQueue
	- type is KS/VS
	- use the term "write" to describe their execution, ie issuing the #put call on the actual store itself


ableegoldman:
Closed in favor of https://github.com/responsivedev/responsive-pub/pull/268

PR #270, opened 2024-03-30
couple tweaks to support safe concurrent access to stores

rodesai:
This patch adds support for concurrent access to stores. After this change, we should be able to concurrently issue read calls to stores while other reads are happening, or while the main stream thread is doing put/delete or flushing.

First, we add a mechanism for injecting values into the store that can only be computed after the store has been initialized. We need to do this because the stores depend on some info that is made available only from the call to , which is kind of a quirk in streams (ideally, these things are just parameters to put/get/etc). In particular in today's implementation we depend on the processor context supplied in init to get the record's timestamp which we need for expiry calculation. This patch supports reading this from a reference to  in the store registration. When an async getter is reading the store, it can inject the appropriate context.

Second, we use a concurrent map for the commit buffer's store of uncommitted records

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SizeTrackingBuffer.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SizeTrackingBuffer.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.NavigableMap;
 import java.util.Objects;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 public class SizeTrackingBuffer<K extends Comparable<K>> {
   private final NavigableMap<K, Result<K>> buffer;
@@ -32,7 +33,7 @@ public class SizeTrackingBuffer<K extends Comparable<K>> {
 
   public SizeTrackingBuffer(final KeySpec<K> extractor) {
     this.extractor = Objects.requireNonNull(extractor);
-    buffer = new TreeMap<>();
+    buffer = new ConcurrentSkipListMap<>();
     reader = Collections.unmodifiableNavigableMap(buffer);
   }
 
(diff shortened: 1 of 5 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SizeTrackingBuffer.java
```diff
@@ -32,7 +33,7 @@ public class SizeTrackingBuffer<K extends Comparable<K>> {
 
   public SizeTrackingBuffer(final KeySpec<K> extractor) {
     this.extractor = Objects.requireNonNull(extractor);
-    buffer = new TreeMap<>();
+    buffer = new ConcurrentSkipListMap<>();
```
ableegoldman, line 36: While I think it should be fine given the frequency of flushing, this data structure can start to break down at scale (in terms of number of unique keys) and using it for the in-memory KV store was the reason behind a major performance regression in C3/Streams a long time ago.

Of course, that would only grow in size whereas the CommitBuffer would only accumulate as many unique keys as are seen within the commit triggers. Still, we should consider running a benchmark and/or doing some quick math on what this might look like for a deduper-like app that contains mostly new keys

ableegoldman: Ah, one more thing I remembered is that this doesn't have an O(1) `size()` implementation (for some wild reason) so if we ever call `#size` on this map, we should use an AtomicInt to keep track of the size ourselves instead

IIRC `isEmpty` is fine, it's just `size` that needs to be avoided

ableegoldman (commented):
LGTM, although there are some NPEs in the tests. Feel free to merge after fixing that

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/SizeTrackingBuffer.java
rodesai: Noted - I was not aware of the perf issues with it. I'm hopeful that we'll be ok because (1) like you said, the commit buffer size is relatively small and is bounded, and (2) our workloads tend to not be cpu-bound, but instead bound on i/o to the stores. Given that, I'm leaning toward trying to keep the implementation here simple and keep the current data structure. We'll see if it's problematic when we benchmark the async stuff.

ableegoldman (approved):
LGTM, thanks

PR #274, opened 2024-04-10
add actions api to controller and implement operator side

rodesai:
This patch makes two related changes:

First, it adds APIs to the controller for serving and reporting the status of actions. The controller will use actions rather than target state to sync down commands. Actions are more flexible:
1. a given action only needs to specify the target state that pertains to the given action, rather than all target state.
2. actions don't need to be target-state based, but can be more traditional "commands", e.g. restart pod X

Second, it updates the operator to use the new actions APIs.

### DIFF
--- a/operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
+++ b/operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
@@ -105,8 +105,9 @@ public class KafkaStreamsPolicyPlugin implements PolicyPlugin {
     LOG.info("target state for app {} {}", appName, targetState);
 
     if (targetState.getTargetState().isPresent()) {
-      LOG.info(
-          "we were not able to get a target state from controller, so don't try to reconcile one");
+      LOG.debug(
+          "we were not able to get a target state from controller, either due to error or"
+              + "there not being a current target. so don't try to reconcile one");
       maybeScaleApplication(
           targetState.getTargetState().get().getKafkaStreamsState().getReplicas(),
           managedApp,
--- a/operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
+++ b/operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
@@ -35,6 +35,8 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import responsive.controller.v1.controller.proto.ControllerOuterClass;
+import responsive.controller.v1.controller.proto.ControllerOuterClass.Action;
+import responsive.controller.v1.controller.proto.ControllerOuterClass.UpdateActionStatusRequest;
 
 public class KafkaStreamsPolicyPlugin implements PolicyPlugin {
   private static final Logger LOG = LoggerFactory.getLogger(KafkaStreamsPolicyPlugin.class);
@@ -84,7 +86,8 @@ public class KafkaStreamsPolicyPlugin implements PolicyPlugin {
 
     LOG.info("Found type {} for app {}/{}", managedApp.appType(), appNamespace, appName);
 
-    responsiveCtx.getControllerClient().currentState(
+    final var controllerClient = responsiveCtx.getControllerClient();
+    controllerClient.currentState(
         ControllerProtoFactories.currentStateRequest(
             environment,
             policy,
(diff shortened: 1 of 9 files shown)


operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
```diff
@@ -101,13 +104,80 @@ public void reconcile(
...
+          targetState.getTargetState().get().getKafkaStreamsState().getReplicas(),
+          managedApp,
+          appNamespace,
+          appName,
+          ctx
+      );
+    }
+
+    for (final Action action : targetState.getActions()) {
+      switch (action.getActionCase()) {
```
agavra, line 120: does it make sense for the action to implement the action itself instead of having a big switch statement? e.g.
```
targetState.getActions().forEach(action -> action.execute(...))
```
That would make it easier to review each action state

operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
```diff
@@ -101,13 +104,80 @@ public void reconcile(
     final var targetState = maybeTargetState.get();
     LOG.info("target state for app {} {}", appName, targetState);
 
-    if (targetState.getTargetState().isEmpty()) {
+    if (targetState.getTargetState().isPresent()) {
       LOG.info(
           "we were not able to get a target state from controller, so don't try to reconcile one");
-      return;
+      maybeScaleApplication(
```
agavra, line 110: it seems like the log message above is no longer accurate?

operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
```diff
@@ -101,13 +104,80 @@ public void reconcile(
     final var targetState = maybeTargetState.get();
     LOG.info("target state for app {} {}", appName, targetState);
 
-    if (targetState.getTargetState().isEmpty()) {
+    if (targetState.getTargetState().isPresent()) {
```
agavra, line 107: just to make sure I understand the compat story here - either (1) it gets a target state (old controller) and it performs the applications scaling or (2) it gets the new "action" and performs that.

if it gets no target state and no action (old controller, no action necessary) it just does nothing (instead of previously where it would log)?

makes sense to me

operator/src/main/java/dev/responsive/k8s/operator/reconciler/ResponsivePolicyReconciler.java
```diff
@@ -74,27 +80,54 @@ PolicyType.KAFKA_STREAMS, new KafkaStreamsPolicyPlugin(environment)
     this.plugins = Objects.requireNonNull(plugins);
   }
 
+  private Optional<ApplicationState> pollTargetState(final ResponsivePolicy policy) {
+    try {
+      return Optional.of(responsiveCtx.getControllerClient().getTargetState(
+          ControllerProtoFactories.emptyRequest(environment, policy)));
+    } catch (final StatusRuntimeException e) {
+      if (e.getStatus().getCode().equals(Status.NOT_FOUND.getCode())) {
+        LOG.info("no target state found");
```
agavra, line 89: these are going to be really spammy right? something we should fix with the operator is that today when there's no target state we have spammy exceptions logs which make it feel like something is wrong. I think it's probably OK not to log this here and instead log when it does find something.

agavra (approved):
LGTM

operator/src/main/java/dev/responsive/k8s/operator/reconciler/ResponsivePolicyReconciler.java
rodesai: yeah that makes sense. I'll drop this

operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
rodesai: yep, that's right

operator/src/main/java/dev/responsive/k8s/operator/reconciler/KafkaStreamsPolicyPlugin.java
rodesai: This is the type generated by protobuf, so we can't customize it like that. We'd need to build up a mirror set of types for use in the main code, and then do conversions back and forth when serializing/deserializing. I do like this approach in general, since these proto-generated types are annoying to deal with. But that's a bigger change than I want to take on here.

PR #275, opened 2024-04-10
operator side changes for restarting a pod

rodesai:
This patch implements the operator changes required to execute an action that restarts a given pod.

Builds on #274


rodesai:
@agavra I'm taking your +1 on #276 as a +1 here (since it's stacked)

PR #277, opened 2024-04-17
switch to using the native thread pool for async processor

rodesai:
this patch switches the async processor over to using the native thread pool instead of a custom thread pool implementation.
- Change AsyncThreadPool to use an ExecutorService. All per-record context is passed in the task that is submtited to the service rather than being tracked in a custom thread implementation.
- Changes AsyncThreadProcessorContext lifetime to be per-record rather than per-thread. We create one of these for each event and set it in the thread-local in AsyncUserProcessorContext before processing a given record.
- Cleans up most of the custom thread pool implementation.


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/stores/AsyncKeyValueStore.java
```diff
@@ -82,10 +84,12 @@ public void executeDelayedWrite(final DelayedWrite<KS, VS> delayedWrite) {
...
     final Thread currentThread = Thread.currentThread();
-
-    if (currentThread instanceof AsyncThread) {
-      final AsyncEvent currentAsyncEvent =
-          ((AsyncThread) currentThread).context(partition).currentAsyncEvent();
+    final MergedProcessorContext<?, ?> rawThreadCtx = userProcessorContext.delegate();
+    if (rawThreadCtx instanceof AsyncThreadProcessorContext) {
+      // todo: yuck
+      final AsyncThreadProcessorContext<?, ?> threadCtx = (AsyncThreadProcessorContext<?, ?>)
+          userProcessorContext.delegate();
```
ableegoldman, line 91: I really agree with the "yuck" here -- I do worry that I'm a bit biased in favor of the way I wrote it originally just because that's now what feels familiar to me, but I really don't like the idea of exposing/using the user context router (ie the AsyncUserProcessorContext) outside of the user's processor. I think it's confusing enough that the context we pass in to the user is going to return a different actual underlying context depending on which thread is currently executing, and I'd prefer to keep that isolated to only what we pass in to the user and not use the "thread-local routing" mechanism within the async processing framework

I'll acknowledge that what I was doing before was exactly the same in practice: grabbing the context for the specific thread that's executing. I just think it's a lot more clear which specific context we're accessing here when you do something like ((AsyncThread) currentThread).context() vs asking for the "delegate context" from some other context which is really obtuse until you already understand exactly what everything is doing and how all the different contexts interact.

I'm also maybe a bit biased because from lurking on various AK tickets I know the use of ThreadLocal has caused some major bugs and general confusion in core Kafka. I'm not against the use of ThreadLocal itself, but I do think it's just somewhat of a non-obvious behavior and the whole which-context-is-which thing is already confusing enough given the many different contexts and types of contexts we're dealing with in the async framework

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/contexts/AsyncThreadProcessorContext.java
```diff
@@ -76,20 +76,14 @@ public class AsyncThreadProcessorContext<KOut, VOut> implements MergedProcessorC
   // time a new event is picked up from the processing queue but before beginning
   // to process it (ie invoking #process on the input record for this event), as
   // part of the preparation for each async process
-  private AsyncEvent currentAsyncEvent;
+  private final AsyncEvent currentAsyncEvent;
```
ableegoldman, line 79: This change makes a lot of sense. I was just blindly following the way the processor contexts work in Streams, which is to have a single context and just mutate its per-record fields as we go through the subtopology. But that mutation has been the direct cause of much grief and added complexity so there's no reason to continue with it here

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -1,173 +1,146 @@
...
-          asyncProcessorName,
-          partition,
-          threadLocalContext,
-          finalizingQueue
-      );
-
-      thread.addProcessor(processorContainer, threadLocalContext, userContext);
+    // todo: can also check in-flight for failed tasks
+    for (final var e : events) {
+      final Future<?> future = executorService.submit(new AsyncEventTask<>(
```
ableegoldman, line 69: I guess I'm also a little uncomfortable with the idea of losing control over the ProcessingQueue. Do we really want to hand over all scheduling to the ExecutorService? Is there any way to customize the scheduling algorithm? I did actually start the POC with an executor service pool but replaced it with the custom threads for some reason, and I think this had something to do with it.

For example: going forward we might want to be smarter about prioritizing certain keys and/or partitions based on how relative weight in the set of blocked tasks. We may also be in a situation where Streams is attempting to close one of its tasks and commit only that one, in which case we should flush all the events for that partition first/with priority. What worries me most is that we just don't even know what a good scheduling algorithm will be and/or how we might want to apply heuristics to improve it under specific conditions or in extreme cases. So I guess I'd prefer to leave that door open.

idk. As always, maybe I'm just being paranoid here. Is there a way to use the ExecutorService without giving up control over the scheduling altogether? (Or is giving up control over scheduling more or less the point of using the ExecutorService?)

Doesn't mean we can't do it now, we should optimize for stability and production-readiness over theoreticals. Just trying to understand the stakes a bit bettter

rodesai: Yep, that all makes sense and was something I thought about a bit as well. In general I'm in favor of removing complexity and adding it if/when we actually need to solve those problems. That said, the native pool is pretty flexible and if we need to we can provide our own queue implementation to the thread pool, which I think lets us do all the stuff that you're describing: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#ThreadPoolExecutor-int-int-long-java.util.concurrent.TimeUnit-java.util.concurrent.BlockingQueue-

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/stores/AsyncKeyValueStore.java
rodesai: I agree with your thoughts here - especially the bit about breaking the context layering encapsulation. I tried to address this in the latest patch by adding a new interface called `DelayedAsyncStoreWriter`, which a caller can use to do a delayed write to the state store. This is what gets used by the async store. It's implemented by the user context, which does the delayed write to the current event if its called on the async thread.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
ableegoldman: Fair enough. The only other thing I was worried about is the error handling aspect and how to deal with failures and user/processors exceptions, but I think you've pointed out in this PR how we can go about that. So I'm starting to get on board with the ExecutorService approach. Will give the PR another pass but I can agree with the general approach for now

rodesai: We have a couple of options for how to do that:
1. We can just have our task implementation catch exceptions thrown by `process` and stick them in the event when we append it to the finalizing queue. And then the finalizer can deal with them.
2. We can have the caller of `AsyncThreadPool` check the open futures for any failed ones and then deal with the errors then.

(1) seems the most straightforward to me.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -1,173 +1,146 @@
...
+  }
 
-    for (final AsyncThread thread : threadPool.values()) {
-      thread.start();
+  public void removeProcessor(final String processorName, final int partition) {
+    log.info("clean up records for {}:{}", processorName, partition);
+    final var key = InFlightWorkKey.of(processorName, partition);
+    final Map<AsyncEvent, Future<?>> inFlightForTask = inFlight.remove(key);
+    if (inFlightForTask != null) {
+      inFlightForTask.values().forEach(f -> f.cancel(true));
```
ableegoldman, line 55: I guess this is another reason to keep the total number of in-flight events pretty low. Closing processors should be fast. But I guess there's no problem since we'll add a config to make sure this isn't a huge list

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
ableegoldman: Agreed -- I think the more context we can keep scoped to a specific event, the better. Will help us provide useful diagnostic info as well

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -1,173 +1,146 @@
...
-    shutdownThread.start();
-  }
+    private InFlightWorkKey(final String processorName, final int partition) {
+      this.processorName = processorName;
+      this.partition = partition;
+    }
 
-  private Thread shutdownThread() {
-    return new Thread(() -> {
+    private static InFlightWorkKey of(final String task, final int partition) {
```
ableegoldman, line 124: nit: task --> processorName

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/contexts/DelayedAsyncStoreWriter.java
```diff
@@ -0,0 +1,7 @@
+package dev.responsive.kafka.api.async.internals.contexts;
+
+import dev.responsive.kafka.api.async.internals.events.DelayedWrite;
+
+public interface DelayedAsyncStoreWriter {
```
ableegoldman, line 5: nice, this approach feels much better to me

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/contexts/StreamThreadProcessorContext.java
```diff
@@ -54,15 +55,18 @@ public class StreamThreadProcessorContext<KOut, VOut>
   private final Map<String, AsyncKeyValueStore<?, ?>> storeNameToAsyncStore = new HashMap<>();
   private final ProcessorNode<?, ?, ?, ?> asyncProcessorNode;
   private final InternalProcessorContext<KOut, VOut> originalContext;
+  private final AsyncUserProcessorContext<KOut, VOut> userProcessorContext;
```
ableegoldman, line 58: Can we change this to a DelayedAsyncStoreWriter as well?

ableegoldman (commented):
Nice, going to make another pass in a bit to look at the tests but I'm fully on board with this approach now

ableegoldman (approved):
Alrighty I think this LGTM (modulo the few comments in my last review)

Feel free to merge once you address those

PR #280, opened 2024-04-18
Use TimestampedKeyValueStore in AsyncProcessorIntegrationTest

ableegoldman:
Forgot to merge this with the async POC branch before merging that PR, so here's a quick followup that makes sure we actually use the TimestampedKeyValueStore in the async integration test.

I needed to make a few small changes to the way we build the normal (non-async) state stores for this to work, mot notably adding a ResponsiveTimestampedKeyValueStore that implements the TimestampedBytesStore marker interface in Streams. We actually had this a while back but I took it out to simplify things because we didn't strictly need it for plain Streams. Well, now we need it for the async processing stuff.

(If you're curious why we need it now, it's because we override the actual builder that adds the various store layers so that we can stick in our special async flushing layer to the async stores. And the Streams builder we override does some fancy magic to wrap stores that don't implement the TimestampedBytesStore in a timestamping-layer. So we could either re-implement this fancy/hacky magic in our builder, which is a non-trivial amount of extra code, or we could just have an actual timestamped version of our own KV store, which is actually very little extra code since it just has to add the marker interface but can otherwise simply extend the non-timestamped store. Also it's probably the right thing to do anyway)

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
@@ -17,6 +17,7 @@
 package dev.responsive.kafka.api.stores;
 
 import dev.responsive.kafka.internal.stores.ResponsiveKeyValueStore;
+import dev.responsive.kafka.internal.stores.ResponsiveTimestampedKeyValueStore;
 import java.util.Locale;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.state.KeyValueBytesStoreSupplier;
@@ -26,8 +27,14 @@ public class ResponsiveKeyValueBytesStoreSupplier implements KeyValueBytesStoreS
 
   private final ResponsiveKeyValueParams params;
 
+  private boolean isTimestamped;
+
   public ResponsiveKeyValueBytesStoreSupplier(final ResponsiveKeyValueParams params) {
     this.params = params;
+  }
+
+  public void asTimestamped() {
+    this.isTimestamped = true;
   }
 
   @Override
@@ -37,7 +44,11 @@ public class ResponsiveKeyValueBytesStoreSupplier implements KeyValueBytesStoreS
 
   @Override
   public KeyValueStore<Bytes, byte[]> get() {
-    return new ResponsiveKeyValueStore(params);
+    if (isTimestamped) {
+      return new ResponsiveTimestampedKeyValueStore(params);
+    } else {
+      return new ResponsiveKeyValueStore(params);
+    }
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
```diff
@@ -26,18 +27,28 @@ public class ResponsiveKeyValueBytesStoreSupplier implements KeyValueBytesStoreS
 
   private final ResponsiveKeyValueParams params;
 
+  private boolean isTimestamped;
```
ableegoldman, line 30: @agavra I know this is super awkward but I wanted to be as non-intrusive as possible, and most importantly, not break things for Metronome.

Of course they'll have to make a few (very minor) code changes to convert to an async processor anyways, so a few more wouldn't hurt. That said, I'd still rather keep the current API where users only need to specify whether it's timestamped or not via the builder, and don't need to worry about having the store supplier match the builder in terms of using the correct API in ResponsiveStores. I don't feel super strongly one way or another though so I'm happy to delegate if you want to design the API differently

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -153,6 +153,14 @@ public static <K, V> StoreBuilder<TimestampedKeyValueStore<K, V>> timestampedKey
       final Serde<K> keySerde,
       final Serde<V> valueSerde
   ) {
+    if (storeSupplier instanceof ResponsiveKeyValueBytesStoreSupplier) {
```
ableegoldman, line 156: This is another slightly awkward thing -- ideally we would just have the storeSupplier parameter type be a `ResponsiveKeyValueBytesStoreSupplier` right off the bat. But then we'd have to change the return type of our store supplier APIs to that, and I worry it might confuse people if they return anything other than the plain KeyValueStoreSupplier type that matches what is needed to plug it into Streams.

Obviously if people are following our docs it's fine, but we know not everyone will. Any thoughts or preferences on this choice? cc @agavra

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueBytesStoreSupplier.java
agavra: I've always been a fan of copying Kafka Streams, even if it isn't ideal so they can just `s/Stores/ResponsiveStores/g` - i.e. it even took me a second to migrate "word count" from the Kafka Streams example because I had to figure out the ResponsiveStores API. It's not a big deal, but I think the more we lean on "Responsive == Kafka Streams" the smoother the transition feels and the easier it is to write/maintain migration docs.

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
agavra: c.f. my comment above - I wouldn't block this PR but I think just copying what Stores does is generally the safest bet.

agavra (approved):
My preference in comments, but won't block this approach :)

PR #281, opened 2024-04-19
Add per-key backpressure mechanism and unit tests for SchedulingQueue

ableegoldman:
Introduced a new config for the max size of the per-key queues that make up the SchedulingQueue. 

We can follow up with the last config which controls the overall number of pending events (see #2 of [this comment](https://github.com/responsivedev/responsive-pub/pull/281#discussion_r1573018556))

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
@@ -16,7 +16,6 @@
 
 package dev.responsive.kafka.api.async.internals;
 
-import static dev.responsive.kafka.api.config.ResponsiveConfig.ASYNC_SCHEDULING_QUEUE_SIZE_CONFIG;
 import static dev.responsive.kafka.internal.config.InternalSessionConfigs.loadAsyncThreadPoolRegistry;
 
 import dev.responsive.kafka.api.async.AsyncProcessorSupplier;
@@ -195,9 +194,9 @@ public class AsyncProcessor<KIn, VIn, KOut, VOut>
     userContext.setDelegateForStreamThread(streamThreadContext);
 
     final ResponsiveConfig configs = ResponsiveConfig.responsiveConfig(userContext.appConfigs());
-    final long schedulingQueueSize = configs.getLong(ASYNC_SCHEDULING_QUEUE_SIZE_CONFIG);
+    final int maxEventsPerKey = configs.getInt(ResponsiveConfig.ASYNC_MAX_EVENTS_PER_KEY_CONFIG);
 
-    this.schedulingQueue = new SchedulingQueue<>(logPrefix, schedulingQueueSize);
+    this.schedulingQueue = new SchedulingQueue<>(logPrefix, maxEventsPerKey);
     this.finalizingQueue = new FinalizingQueue(logPrefix);
 
     this.threadPool = getAsyncThreadPool(internalContext, streamThreadName);
@@ -256,8 +255,9 @@ public class AsyncProcessor<KIn, VIn, KOut, VOut>
   private void processNewAsyncEvent(final AsyncEvent event) {
     pendingEvents.add(event);
 
-    if (schedulingQueue.isFull()) {
-      backOffSchedulingQueue();
+    final KIn eventKey = event.inputRecordKey();
+    if (schedulingQueue.keyQueueIsFull(eventKey)) {
+      backOffSchedulingQueueForKey(eventKey);
     }
 
     schedulingQueue.offer(event);
@@ -358,16 +358,18 @@ public class AsyncProcessor<KIn, VIn, KOut, VOut>
    */
   private int finalizeAtLeastOneEvent() {
     final int numFinalized = drainFinalizingQueue();
-    if (numFinalized == 0) {
-      try {
-        final AsyncEvent finalizableEvent = finalizingQueue.waitForFinalizableEvent();
-        completePendingEvent(finalizableEvent);
-      } catch (final Exception e) {
-        log.error("Exception caught while waiting for an event to finalize", e);
-        throw new StreamsException("Failed to flush async processor", e, taskId);
-      }
+    if (numFinalized > 0) {
+      return numFinalized;
     }
-    return numFinalized == 0 ? 1 : numFinalized;
+
+    try {
+      final AsyncEvent finalizableEvent = finalizingQueue.waitForNextFinalizableEvent();
+      completePendingEvent(finalizableEvent);
+      return 1;
+    } catch (final Exception e) {
+      log.error("Exception caught while waiting for an event to finalize", e);
+      throw new StreamsException("Failed to flush async processor", e, taskId);
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/queues/FinalizingQueue.java
```diff
@@ -76,6 +78,8 @@ public AsyncEvent nextFinalizableEvent() {
 
   /**
    * See {@link ReadOnlyFinalizingQueue#waitForFinalizableEvent()}
+   * <p>
+   * Note: blocking API
    */
   @Override
   public AsyncEvent waitForFinalizableEvent() throws InterruptedException {
```
rodesai, line 85: nit: name this `waitForNextFinalizableEvent` to keep it consistent with `nextFinalizableEvent`

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -204,13 +204,23 @@ public class ResponsiveConfig extends AbstractConfig {
...
+
   public static final String ASYNC_THREAD_POOL_SIZE_CONFIG = "responsive.async.thread.pool.size";
   private static final int ASYNC_THREAD_POOL_SIZE_DEFAULT = 0;
   private static final String ASYNC_THREAD_POOL_SIZE_DOC = "The number of async processing threads to "
       + "start up for each StreamThread in this app. Setting this to 0 (the default) means async processing "
       + "will not be enabled. Setting this to a positive integer will enable async processing, but only if "
       + "there is at least one AsyncProcessor in the topology. See javadocs for AsyncProcessorSupplier for details.";
 
+  public static final String ASYNC_SCHEDULING_QUEUE_SIZE_CONFIG = "responsive.async.scheduling.queue.size";
+  private static final long ASYNC_SCHEDULING_QUEUE_SIZE_DEFAULT = 1L;
```
rodesai, line 218: why a default of 1 here? seems like it should be higher

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -252,8 +255,12 @@ public void process(final FixedKeyRecord<KIn, VIn> record) {
 
   private void processNewAsyncEvent(final AsyncEvent event) {
     pendingEvents.add(event);
-    schedulingQueue.offer(event);
 
+    if (schedulingQueue.isFull()) {
```
rodesai, line 259: instead of a cap on `schedulingQueue`, why not just track a cap on `pendingEvents`? Then you don't need separate backpressure mechanisms for the scheduling queue and thread pool. The changes here won't help much if there is no overlap between record keys.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
ableegoldman: whoops, didn't mean to commit this. I agree

That said, I'm not really sure what a good default would be? I had it set to 100_000_000L initially as an arbitrary large-but-not-infinite value. I figured we could play around with it a bit when we have the perf testing set up. But idk -- thoughts?

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
ableegoldman: > The changes here won't help much if there is no overlap between record keys.

Agreed, I was going to do a separate PR with backpressure on the ProcessingQueue (or the ExecutorService task queue if we get rid of the ProcessingQueue). The FinalizingQueue doesn't really need a cap since the StreamThread can/should/will just drain that before proceeding anyways.

I was going to define one config for the SchedulingQueue and one for the "ProcessingQueue" (again, or SE task queue). And then tell users that one of them is used to apply backpressure when you have a large keyspace/many unique keys, while the other config helps backpressure when you have many same-key uypdates. I thought this would be more straightforward for users to act on (assuming they actually read the config doc string) and it could be useful to apply backpressure individually to different stages of async processing.

I guess a cap on `pendingEvents` is maybe easier for users to actually understand the meaning of, even if the actual affect/scenario it applies to is less clear (ie much key overlap vs large keyspace).

I don't feel too strongly either way, just explaining the thought process behind doing it this way. WDYT?

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
rodesai: if we set it too high we'd risk buffering up so much we miss a poll right? I think something pretty small is actually just fine - like some multiple of the number of async threads.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
rodesai: I think I'm mostly coming around. The way I'm thinking about it now is that a user would want to be able to control:
1. The desired parallelism per stream thread - this is just the size of the async thread pool
2. The max number of records to buffer up waiting to run, to try to limit the time you'd ever be blocked in a `flush` or `process`. We should default this to something pretty small - lets say 2x the size of the async thread pool. And this limit should be at the thread level.
3. Limit the number of records queued up for one key to limit the size of any queue that you have to process serially. But in that case, wouldn't it make sense to have the config be to limit the size of any one key's queue? E.g. a limit of 3 means you can only ever have at most 2 records blocked waiting on an earlier record with the same key to be processed. And again, this limit should be pretty small to avoid long periods of blocking.

ableegoldman: I think this makes sense to me. To put it in terms of concrete configs, we'd have:

1. num async threads per StreamThread -- done (already merged)
2. cap on the overall number of in-flight records, ie the size of the `pendingEvents` across all processors for the StreamThread -- this can be a followup PR (I'll convert the ticket for backpressure on the ProcessingQueue into this)
3. cap on the size of the per-key queues within the SchedulingQueue -- I'll do that in this PR (will ping you when ready)

rodesai: For (2) I think that's reasonable to implement as a limit on the number of records queued at the processing queue/pool level for a given thread (as opposed to as a limit on `pendingEvents`). I can't convince myself one is better than the other. WDYT?

ableegoldman: 2x seems kinda low to me for a Thread-wide limit. It's also not going to be possible to even implement exactly -- because if you have 5 async threads and a limit of 10 in flight events, but 40 partitions on that StreamThread, then if we're at that limit and processing a new record for some task X, there's only a 1/4 chance we even have an in-flight record for that partition. And since we can't have the StreamThread finalizing records for other partitions from the task it's currently processing, we'd be stuck forever -- or more realistically, we'd just need to ignore the cap and move on. But then what's the point of a cap this low if we have to ignore it so often?

(40 partitions per StreamThread maybe sounds silly but one of the main benefits of async processing could be/should be to use fewer resources to get the same performance -- which means stacking more partitions per node/StreamThread. So we should make sure the framework works especially well in this case)

ableegoldman: Sorry I only just saw your earlier question about (2). I'm personally a bit worried about some processors/partitions potentially crowding out others, but I do think it makes sense to have a thread-wide limit. If you look at the way I implement this in the followup PR ([here](https://github.com/responsivedev/responsive-pub/pull/282/files)), there's no per-partition/processor limit enforced, but it would be very easy to add one. Since we basically need to track the number of queued events per partition anyways (in order to drop partitions).

I'll file a ticket for per-partition limits but put it under "Developer Experience" for now as I think the thread-wide limit is sufficient for production

rodesai (commented):
LGTM. nice test!

PR #282, opened 2024-04-23
Add StreamThread-level backpressure mechanism for queued up events

ableegoldman:
Followup to the[ first backpressure PR ](https://github.com/responsivedev/responsive-pub/pull/281)and implements item 2 from [this comment.](https://github.com/responsivedev/responsive-pub/pull/281#discussion_r1573018556)

NOTE: currently blocked on [the async threadpool refactor](https://github.com/responsivedev/responsive-pub/pull/277)
Once that's merged, we'll need to rebase this so we can call `#dequeueEvent` from the new async event runner


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -40,6 +41,7 @@ public class AsyncThreadPool {
   private final Logger log;
 
   private final String streamThreadName;
+  private final QueuedEvents queuedEvents;
```
ableegoldman, line 44: TODO: once we merge https://github.com/responsivedev/responsive-pub/pull/277 we need to pass this into the runnable so it can notify when a new event is dequeued

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/events/AsyncEvent.java
```diff
@@ -111,7 +111,10 @@ public enum State {
 
   private final Object inputRecordKey;
   private final Object inputRecordValue;
+
+  private final String asyncProcessorName;
```
ableegoldman, line 115: I didn't end up using/needing this in the current PR, but I'm leaving it in since I think it might be useful/necessary later (ie once we expand to allow more than one async processor per topology)

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -214,14 +214,26 @@ public class ResponsiveConfig extends AbstractConfig {
       + "will not be enabled. Setting this to a positive integer will enable async processing, but only if "
       + "there is at least one AsyncProcessor in the topology. See javadocs for AsyncProcessorSupplier for details.";
 
-  public static final String ASYNC_MAX_EVENTS_PER_KEY_CONFIG = "responsive.async.max.events.per.key";
-  private static final int ASYNC_MAX_EVENTS_PER_KEY_DEFAULT = 5;
-  private static final String ASYNC_MAX_EVENTS_PER_KEY_DOC = "The maximum number of pending events at a time for "
-      + "a given key. This puts a limit on how many events for each key can be either currently processing or awaiting "
-      + "processing by the async thread pool. Since all events with the same key must be processed in offset order, this"
+  public static final String ASYNC_MAX_EVENTS_QUEUED_PER_ASYNC_THREAD_CONFIG = "responsive.async.max.events.queued.per.async.thread";
```
ableegoldman, line 217: this is the new config -- I thought it made sense to just make it a multiplicative factor instead of the absolute event cap, but I worry this might go over people's heads? Do the docs/config name make this clear enough? Happy to accept naming suggestions

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -214,14 +214,26 @@ public class ResponsiveConfig extends AbstractConfig {
...
+  private static final int ASYNC_MAX_EVENTS_QUEUED_PER_ASYNC_THREAD_DEFAULT = 5;
```
ableegoldman, line 218: @rodesai  As I said earlier, 2 felt really low to me, but I'm open to changing the default. For this and the per-key one

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
rodesai: This makes sense to me - it describes exactly what it is.

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
rodesai: These defaults look fine to me. I don't really see the benefit of accumulating lots of records in the queues. You're going to have to pay the cost at flush time anyway. One potential improvement would be to compute the default for `responsive.async.max.events.queued.per.async.thread` from the size of the thread pool (it could just be the size of the thread pool times some factor (even x1 seems fine - you have one event enqueued for each thread and one being actively processed))

rodesai (approved):
LGTM!

PR #283, opened 2024-04-23
Schedule punctuator to flush async processor

ableegoldman:
Quick PR to schedule regular flushes via the Kafka Streams punctuators. Adds a config for the flushing frequency and then schedules a punctuator in the asyncProcessor that simply invokes the existing non-blocking flush method that  gets called during #process


rodesai (approved):
LGTM!

PR #285, opened 2024-04-26
Re-set the internal "leave.group.on.close" config to true

ableegoldman:
See https://issues.apache.org/jira/browse/KAFKA-16514 for some context

This is a temporary solution for us until that ticket is resolved and we can use the actual KafkaStreams#close(closeOptions) API with correct semantics

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
@@ -159,6 +159,7 @@ public final class ResponsiveKafkaClientSupplier implements KafkaClientSupplier 
     LOG.info("Creating responsive main consumer: {}", clientId);
 
     // Reset the forced override of this config done by Kafka Streams
+    // TODO: remove this and use KafkaStreams#close(closeOptions) once KAFKA-16514 is fixed
     config.put("internal.leave.group.on.close", true);
 
     final String streamThreadName = extractThreadNameFromConsumerClientId(clientId);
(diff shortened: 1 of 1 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveKafkaClientSupplier.java
```diff
@@ -155,15 +155,21 @@ public Producer<byte[], byte[]> getProducer(final Map<String, Object> config) {
 
   @Override
   public Consumer<byte[], byte[]> getConsumer(final Map<String, Object> config) {
-    final String clientId = (String) config.get(ConsumerConfig.CLIENT_ID_CONFIG);
+    final Map<String, Object> consumerConfig = new HashMap<>(config);
+
+    // Reset the forced override of this config done by Kafka Streams
+    // TODO: remove this and use KafkaStreams#close(closeOptions) once KAFKA-16514 is fixed
+    consumerConfig.put("internal.leave.group.on.close", true);
```
rodesai, line 162: what's the behavior going to be like on a roll now? 2 rebalances, the first of which assigns all the rolled nodes tasks to other nodes, and then another rebalance that assigns them back?

also, do we want to do this if static membership is enabled? in that case we'd not want to leave the group immediately right?

ableegoldman: yeah... guess it might not be worth the tradeoff. Hopefully we can just get the actual API fixed in Kafka/Streams

This won't have any impact on static membership though. Basically the only way to leave the group is by sending a special request type, which this config does not affect (ie it has to be done manually). That's what the KafkaStreams#close(closeOptions) API will do if you select "leave group on close" -- and it's why the API only "works" as intended when using static membership. It just adds this "leave group as static member" request to the close process, which is a no-op for non-static members

I suppose if we have any users who do enable static membership then we should already be using the KafkaStreams#close(closeOptions) API. But static membership seems pretty rare in the wild these days

ableegoldman:
Closing this for now, will work on pressuring Kafka to fix this in the consumer and/or Kafka Streams  so we can use the actual Streams close API as intended instead of getting locked into always leaving the group

PR #286, opened 2024-04-27
add support for async event error handling

rodesai:
This patch adds support for async event error handling:
- AsyncEvent now has a new state called PROCESSING_FAILED. When the event is in this state, all calls to normal processing methods like transitionX or fetching writes/forwards throw the processing exception.
- AsyncThreadPool catches all processing exceptions and transitions the event to PROCESSING_FAILED
- It is possible (though unlikely - it's a bug if it happens) that an async thread hits an exception that our thread pool doesnt catch and put on the finalizing queue. In that case, the main stream thread can observe the exception by looking at the futures returned by the pool. In this patch, I've moved the futures to `AsyncProcessor` and also check those when blocking on the finalizing queue and when flushing.
- AsyncProcessorIntegration test is extended to support error injection. Errors are specified on the input records.


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -381,6 +399,17 @@ public void flushAndAwaitPendingEvents() {
     assertQueuesEmpty();
   }
 
+  private void checkUncaughtExceptionsOnAsyncThreads() {
+    final var uncaught = threadPool.checkUncaughtExceptions(asyncProcessorName, taskId.partition());
+    if (uncaught.isPresent()) {
+      if (processingException != null) {
```
ableegoldman, line 405:
```suggestion
      if (processingException == null) {
```

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -112,91 +126,94 @@ private static String generateAsyncThreadName(
...
+    private final ConcurrentMap<AsyncEvent, InFlightEvent> inFlightForTask;
 
     private AsyncEventTask(
         final AsyncEvent event,
         final ProcessingContext originalContext,
         final AsyncUserProcessorContext<KOut, VOut> userContext,
         final FinalizingQueue finalizingQueue,
-        final Map<AsyncEvent, Future<?>> inFlightForTask,
-        final Logger log
+        final ConcurrentMap<AsyncEvent, InFlightEvent> inFlightForTask
```
ableegoldman, line 195: not used?

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -358,13 +367,22 @@ private static void registerFlushListenerForStoreBuilders(
    * each commit, similar to #flushCache
    */
   public void flushAndAwaitPendingEvents() {
+    if (processingException != null) {
+      // if there was a processing exception, processing for the key that hit the processing
+      // exception is blocked, so we don't want to go into the loop below that tries to drain
+      // queues, as the queue for that key will never drain. Instead, just throw here.
```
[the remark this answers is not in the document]
rodesai: yep, fair enough. I agree that we want to throw immediately for all those reasons. I was trying to point out that you don't want to move this check because you'll probably dead hang if you do, which seems really bad - but honestly doing anything other than dropping dead when you're in an undefined state is potentially just as bad and can lead to corruption. I'll clarify.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
rodesai: ah - good catch!

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -112,91 +126,94 @@ private static String generateAsyncThreadName(
...
+              e, taskContext, asyncProcessorContext, finalizingQueue, inFlightForTask),
+          executor
+      );
+      final var inFlightEvent = new InFlightEvent(future);
+      inFlightForTask.put(e, inFlightEvent);
+      future.handle((r, t) -> {
+        if (t != null) {
+          inFlightEvent.setError(t);
+        } else {
+          inFlightForTask.remove(e);
```
ableegoldman, line 150: I guess it doesn't really matter right now, but looking forward to when we might want to allow actually handling processing exceptions and need to keep the internal state consistent even after a failure...we should probably remove it from the inFlight map regardless of whether it failed?

rodesai: Not in this case. This is really just a failsafe. Any exception that we want to handle gracefully should go onto the finalizing queue. The check here is just to catch anything that slips through, and if that happens its a bug and we should exit the main thread (when it calls `checkUncaughtExceptions`)

PR #288, opened 2024-04-27
Expand Async integration test with upstream and downstream stateful processors

ableegoldman:
Tried to flush out the bug with not resetting the processor context after forwarding records but unfortunately(?) this test does seem to still be passing. Will try playing around with some of the parameters to see if I can hit it, but in the meantime, here's the expanded test.

Also has some refactoring/utilities lined up for some of the other testing I've been working on

### DIFF
--- a/kafka-client/src/test/java/dev/responsive/kafka/async/AsyncProcessorIntegrationTest.java
+++ b/kafka-client/src/test/java/dev/responsive/kafka/async/AsyncProcessorIntegrationTest.java
@@ -162,11 +162,12 @@ public class AsyncProcessorIntegrationTest {
                 ResponsiveKeyValueParams.fact(IN_KV_STORE)),
             IN_KV_STORE)
         .processValues(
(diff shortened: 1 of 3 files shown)


kafka-client/src/test/java/dev/responsive/kafka/async/AsyncProcessorIntegrationTest.java
```diff
@@ -177,91 +199,55 @@ public void shouldProcessEventsInOrderByKey() throws Exception {
     assertThat(processed.get(), equalTo(numProcessedRecords));
     assertThat(latestValues, equalTo(finalOutputRecords));
   }
-  
-  private static class UserFixedKeyProcessor implements FixedKeyProcessor<String, String, String> {
```
ableegoldman, line 181: There's a small change to the output format (for obvious reasons) but other than that nothing changed about the way this processor works. I just extracted it out to a standalone class so I can use the framework to plug in stateful processors more easily across integration tests

PR #291, opened 2024-05-07
Top-level error handling

ableegoldman:


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -299,12 +300,23 @@ private void processNewAsyncEvent(final AsyncEvent event) {
...
 
-    executeAvailableEvents();
+    } catch (final StreamsException streamsException) {
+      fatalException = streamsException.taskId().isPresent()
+          ? streamsException
+          : new StreamsException(streamsException, taskId);
+      throw fatalException;
+    } catch (final Throwable t) {
+      fatalException = new StreamsException(t, taskId);
+      throw fatalException;
```
ableegoldman, line 318: This block is repeated a few times -- can put it into a single method if you want to avoid duplicate code but in theory we should never touch this stuff again. 

Basically I extracted all setting and checking of the top-level `fatalException` to the top level of the three "entry points" to the AsyncProcessor class: the `#process` methods, the `#punctuate`, and the `#flushAndAwaitPendingEvents`

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -477,6 +500,25 @@ private void maybeBackOffEnqueuingNewEventWithKey(final KIn key) {
     }
   }
 
+  /**
+   * Invoked by the punctuator for this processor
+   */
+  private void punctuate(final long timestamp) {
```
ableegoldman, line 506: There's nothing new here, we just used to call `executeAvailableEvents` directly and now we have a dedicated method for punctuation (with error handling and logging now, so I guess that part is new)

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -594,21 +639,26 @@ private StreamThreadProcessorContext.PreviousRecordContextAndNode preFinalize(
...
+    final var toClose = streamThreadContext.prepareToFinalizeEvent(event);
+
+    final Optional<Throwable> processingException = event.processingException();
+    if (processingException.isEmpty()) {
+      event.transitionToFinalizing();
+    } else {
       pendingEvents.remove(event);
-      this.fatalException = processingException.get();
-      throw this.fatalException;
+      event.transitionToDone();
```
ableegoldman, line 650: this part is different: not just that we don't set the `fatalException` here anymore, but we also transition events that were FAILED to DONE. Two reasons:
- I think it makes sense to retain DONE as the terminal state that's associated with an event being removed from the `pendingEvents` map
- We can thus use the event's state to make sure that we have checked on an error for a given event. If it's stuck in FAILED that means we haven't ever picked up the error and handled it on the StreamThread end. Once we do this, the event is transitioned from FAILED to DONE. So FAILED is  like an alternative to finalization rather than a terminal state of its own

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -33,6 +33,11 @@ public class AsyncThreadPool {
       = new HashMap<>();
   private final BlockingQueue<Runnable> processingQueue;
 
+  // TODO: we don't really need to track this by processor/partition, technically an
+  //  error anywhere is fatal for the StreamThread and all processors + all further
+  //  processing should shut down ASAP to minimize ALOS overcounting
```
ableegoldman, line 38: This is what I mentioned on slack: we should really consider an exception on any partition to be fatal for all, and furthermore actually suspend the AsyncThreadPool altogether (ie cancel all in-flight events) as soon as we see a fatal error anywhere. But I'll leave this for followup

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -77,31 +82,33 @@ boolean isEmpty(final String processorName, final int partition) {
...
   ) {
-    final var key = InFlightWorkKey.of(processorName, partition);
-    final Map<AsyncEvent, InFlightEvent> inFlightForTask
-        = inFlight.computeIfAbsent(key, k -> new ConcurrentHashMap<>());
-    return inFlightForTask.values().stream()
-        .map(InFlightEvent::error)
-        .filter(Optional::isPresent)
-        .map(Optional::get)
-        .findFirst();
+    return Optional.ofNullable(fatalExceptions.get(new InFlightWorkKey(processorName, partition)));
```
ableegoldman, line 105: This is much simpler now 🙂

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -135,21 +142,36 @@ public <KOut, VOut> void scheduleForProcessing(
...
-        if (t != null) {
-          inFlightEvent.setError(t);
-        } else {
-          inFlightForTask.remove(e);
-        }
-        return null;
-      });
+      inFlightForTask.put(event, inFlightEvent);
+
+      future
```
ableegoldman, line 154: This refactoring of the future/adding stages is one of the main changes in this PR so definitely take a close look.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -180,39 +195,38 @@ public synchronized void setError(final Throwable error) {
...
-      try {
-        wrappingContext.setDelegateForAsyncThread(new AsyncThreadProcessorContext<>(
-                originalContext,
-                event
-        ));
-        event.transitionToProcessing();
-        event.inputRecordProcessor().run();
-        finalizingQueue.scheduleForFinalization(event);
-      } catch (final RuntimeException e) {
-        finalizingQueue.scheduleFailedForFinalization(event, e);
```
ableegoldman, line 211: Moved all the exception handling into the completable future stages

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/queues/FinalizingQueue.java
```diff
@@ -56,37 +53,35 @@ public FinalizingQueue(final String logPrefix, final int partition) {
...
-          "attempted to finalize an event for partition %d on queue for partition %d",
-          processedEvent.partition(),
-          this.partition
-      ));
+      log.error(errorMsg);
+      throw new IllegalStateException(errorMsg);
     }
-    // Transition to OUTPUT_READY to signal that the event is done with processing
-    // and is currently awaiting finalization by the StreamThread
-    processedEvent.transitionToToFinalize();
```
ableegoldman, line 78: Another change of note: made this into a "pure" queue which doesn't do any event manipulations and just holds onto events. These event transitions now take place in the AsyncThreadPool class

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/queues/FinalizingQueue.java
```diff
@@ -56,37 +53,35 @@ public FinalizingQueue(final String logPrefix, final int partition) {
...
-      final AsyncEvent processedEvent,
-      final RuntimeException exception
+  public void addFailedEvent(
+      final AsyncEvent event,
+      final Throwable throwable
   ) {
-    processedEvent.transitionToToFailed(exception);
-    finalizableRecords.add(processedEvent);
+    // Failed events get to jump the line to make sure the StreamThread fails fast
+    finalizableRecords.addFirst(event);
```
ableegoldman, line 84: Another change of interest: made this a dequeue and stashed failed events in the front so they get picked up ASAP, since we want to minimize the number of events that get finalized after a fatal exception is thrown

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -135,21 +142,29 @@ public <KOut, VOut> void scheduleForProcessing(
...
+          .exceptionally(processingException -> {
+            inFlightEvent.setError(processingException);
+            event.transitionToFailed(processingException);
+            finalizingQueue.addFailedEvent(event, processingException);
+            return null;
+          })
+          .exceptionally(processingException -> {
+            // do this in separate stage to ensure we always catch a fatal exception, even
+            // if we somehow hit another exception in the previous exception handling stage
+            fatalExceptions.computeIfAbsent(inFlightKey, k -> processingException);
```
ableegoldman, line 165: This is our "failsafe" for any exceptions that don't end up in the finalizing queue. In theory that should never happen, but it's technically possible for something in the previous `exceptionally` block to throw

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -380,42 +392,55 @@ public void flushAndAwaitPendingEvents() {
...
 
-    // Make a (non-blocking) pass through the finalizing queue up front, to
-    // free up any recently-processed events before we attempt to drain the
-    // scheduling queue
-    drainFinalizingQueue();
-
-    while (!isCleared()) {
-      // we need to check this here because the check in finalizeAtLeastOneEvent only
-      // runs if there are currently slow events being handled by the async threads.
-      checkUncaughtExceptionsOnAsyncThreads();
```
ableegoldman, line 391: I essentially moved this check by putting it at the start of the `drainFinalizingQueue` method. It solves the problem mentioned here by ensuring that it always gets checked in finalizeAtLEastOneEvent.

It also helps with the check during punctuations and flushes, and to make sure that in case of an uncaught error (ie one that didn't make it into the finalizing queue), we abort quickly

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -364,42 +376,55 @@ public void flushAndAwaitPendingEvents() {
...
-        fatalException = uncaught.map(t ->
-          t instanceof RuntimeException ? (RuntimeException) t : new RuntimeException(t)).get();
-      }
+    } catch (final StreamsException streamsException) {
+      fatalException = streamsException.taskId().isPresent()
+        ? streamsException
+        : new StreamsException(streamsException, taskId);
+      throw fatalException;
+    } catch (final Throwable t) {
+      fatalException = new StreamsException(t, taskId);
```
rodesai, line 406: why wrap these in `StreamsException`?

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -364,42 +376,55 @@ public void flushAndAwaitPendingEvents() {
...
+   * Note: since it's important to fail as quickly as possible in the event of a
+   * fatal exception, we don't limit the scope of this check to only errors
+   * tied to events that belong to this specific AsyncProcessor or task.
+   */
+  private void checkUncaughtExceptionsInAsyncThreadPool() throws Throwable {
+    final Optional<Throwable> fatal = threadPool.checkUncaughtExceptions(
+        asyncProcessorName, taskId.partition()
+    );
+    
+    if (fatal.isPresent()) {
```
rodesai, line 423: can we wrap it in a `RuntimeException` if its a checked exception? will help get rid of the `throws Throwable` everywhere which I don't think is adding anything

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
rodesai: makes sense to me!

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
ableegoldman: Fair enough

PR #293, opened 2024-05-07
add first set of metrics to async processor

rodesai:
- do a minor refactor to ResponsiveMetrics. We change the name MetricGroup to MetricScope, and have ResponsiveMetrics construct concrete instances of MetricScope for specific pre-defined scopes like thread, task, etc. The main difference is that the group is now passed in from the caller. So you can have different groups with the same scope, ensuring that different metrics are namespaced but use a common set of tags.
- add processor and thread level metrics scopes
- add a recorder class for async processor metrics that records metrics at processor level scope. For now, it records:
  - the number of pending events
  - the number of events blocked in the scheduling queue
  - the number of events blocked behind the most backed up key in the scheduling queue
  - the max/avg time spent actually processing events
  - the max/avg time spent between state transitions
- add a thread-level recorder class for the thread pool. currently it just reports the thread pool queue size


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/queues/KeyOrderPreservingQueue.java
```diff
@@ -0,0 +1,218 @@
+/*
```
rodesai, line 1: this is just `SchedulingQueue`, renamed so I could have a `SchedulingQueue` interface

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/queues/MeteredSchedulingQueue.java
```diff
@@ -0,0 +1,66 @@
+package dev.responsive.kafka.api.async.internals.queues;
+
+import dev.responsive.kafka.api.async.internals.events.AsyncEvent;
+import dev.responsive.kafka.api.async.internals.metrics.AsyncProcessorMetricsRecorder;
+
+public class MeteredSchedulingQueue<KIn> implements SchedulingQueue<KIn> {
+  private final AsyncProcessorMetricsRecorder metricsRecorder;
+  private final SchedulingQueue<KIn> wrapped;
+
+  public MeteredSchedulingQueue(
```
rodesai, line 10: I wound up measuring the scheduling metrics by wrapping it in this class and calling `recordX` methods. Alternatively, I could have set up gauges that sample the sizes directly. I chose this approach so I wouldn't have to make `SchedulingQueue` thread-safe

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncProcessorMetricsRecorder.java
```diff
@@ -0,0 +1,146 @@
...
+    schedulingQueueLongestQueueSizeSensor.record(size);
+  }
+
+  public void recordStateTransition(
+      final AsyncEvent.State from,
+      final Instant fromTime,
+      final AsyncEvent.State to,
+      final Instant toTime
+  ) {
+    final var innerScope = scope
```
ableegoldman, line 120: This is very nitty so feel free to ignore, but imo some things are bit difficult to follow when everything's declared as a var. Like this line. Probably my own fault for reviewing directly on github instead of pulling in the PR locally but sometimes I'm lazy 🙂

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncProcessorMetricsRecorder.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.kafka.common.metrics.stats.Avg;
+import org.apache.kafka.common.metrics.stats.Max;
+import org.apache.kafka.common.metrics.stats.Value;
+import org.apache.kafka.streams.processor.TaskId;
+
+public class AsyncProcessorMetricsRecorder {
+  public static final String GROUP_NAME = "async-processor-metrics";
+
+  public static final String PENDING_EVENTS = "pending-events";
+  public static final String PENDING_EVENTS_DESC = "events pending in processor";
```
ableegoldman, line 27:
```suggestion
  public static final String PENDING_EVENTS_DESC = "The current number of events pending in processor";
```

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncThreadPoolMetricsRecorder.java
```diff
@@ -0,0 +1,32 @@
+package dev.responsive.kafka.api.async.internals.metrics;
+
+import dev.responsive.kafka.internal.metrics.ResponsiveMetrics;
+import java.util.Objects;
+import java.util.function.Supplier;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.Gauge;
+
+public class AsyncThreadPoolMetricsRecorder {
+  public static final String GROUP_NAME = "async-processor-pool-metrics";
```
ableegoldman, line 10: should this be "async-thread-pool-metrics"?

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncThreadPoolMetricsRecorder.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public static final String QUEUED_EVENTS = "queued-events";
+  public static final String QUEUED_EVENTS_DESC = "events pending in thread pool queues";
```
ableegoldman, line 13:
```suggestion
  public static final String QUEUED_EVENTS_DESC = "The total number of events pending in the async thread pool queue";
```

ableegoldman (commented):
Looks good overall. Planning to redeploy the soak from this tonight so I can check out these metrics myself

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -75,8 +80,10 @@ public class AsyncProcessor<KIn, VIn, KOut, VOut>
   private final Map<String, AbstractAsyncStoreBuilder<?, ?, ?>> connectedStoreBuilders;
 
   // Tracks all pending events, ie from moment of creation to end of finalization/transition
-  // to "DONE" state. Used to make sure all events are flushed during a commit
-  private final Set<AsyncEvent> pendingEvents = new HashSet<>();
+  // to "DONE" state. Used to make sure all events are flushed during a commit.
+  // We use a concurrent map here so metrics readers can query its size. Otherwise, only
+  // the stream thread should access this.
+  private final Map<AsyncEvent, Object> pendingEvents = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
rodesai: yeah that should be the only effect practically - you might get a cached value for `size` so it might flip flop depending on where the metric is being read from (its an int so I don't think you can get a partial read). That said, `HashSet` is technically free to throw if multiple threads access it. It just feels safer to use a concurrent map if it's going to be accessed concurrently.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncProcessorMetricsRecorder.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public static final String PROCESS_TIME = "event-process-time-ns";
```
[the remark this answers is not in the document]
rodesai: That's fair. How about `event-process-duration-ns`? I feel like latency kind of implies e2e latency.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -184,29 +202,34 @@ private static class AsyncEventTask<KOut, VOut> implements Runnable {
...
       try {
+        final Instant start = Instant.now();
         wrappingContext.setDelegateForAsyncThread(new AsyncThreadProcessorContext<>(
                 originalContext,
                 event
         ));
         event.transitionToProcessing();
         event.inputRecordProcessor().run();
         finalizingQueue.scheduleForFinalization(event);
+        metricsRecorder.recordEventProcess(Duration.between(start, Instant.now()));
```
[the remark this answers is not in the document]
rodesai: > another performance issue we encountered in Streams was due to converting between long and Instant/Duration on record fetches.

Interesting. I don't really know anything about the internals with the time api. I figured it was just doing the same math internally. I can switch to manually computed times.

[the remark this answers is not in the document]
rodesai: re performance of timers: there's going to be _some_ performance hit from reading the time (though it should be a lot less than it used to be now that the default time source on ec2 nodes is sane). My thinking for this scenario is that:
1. these metrics are super useful for debugging, especially in the early phases of this feature.
2. in the context of the use cases we know we want to use this for now, I'm confident the overhead won't be meaningful.
3. the metrics aren't part of a public contract like ak metrics are, so we can change them to debug if/when we feel like we don't want to eat the performance penalty.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncProcessorMetricsRecorder.java
rodesai: nah that's fair. Putting a type here doesn't hurt.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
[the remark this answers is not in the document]
ableegoldman: > That said, HashSet is technically free to throw if multiple threads access it.

Ah, didn't know that. Anyways that all makes sense 👍

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/metrics/AsyncProcessorMetricsRecorder.java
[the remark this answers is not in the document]
ableegoldman: yep that SGTM

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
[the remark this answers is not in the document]
ableegoldman: Yep that all sounds fair. Unless the soak drops like a rock when I deploy with these metrics I won't complain (any more :P)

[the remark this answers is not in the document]
ableegoldman: Also it probably all comes back to that same issue with #currentTimeMillis, since `Instant#now` uses that internally. But here we only need #currentTimeNanos which is always fast, since we don't need an exact clock just a relative one. But I'll shut up now

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -153,6 +155,7 @@ public <KOut, VOut> void scheduleForProcessing(
       final AsyncUserProcessorContext<KOut, VOut> asyncProcessorContext,
       final AsyncProcessorMetricsRecorder processorMetricsRecorder
   ) {
+    maybeInitThreadPoolMetrics();
```
ableegoldman, line 158: I don't love doing this in #scheduleForProcessing...putting it in a separate #init method (maybe even waiting to pass the ResponsiveMetrics in until #init is called) seems slightly preferable to me. But honestly it's fine if you want to just merge it as-is

(my main concern is that this feels not very future proof, if we ever added another method to AsyncThreadPool that could potentially be invoked before #scheduleForProcessing is, and has some kind of metric to record)

rodesai: Oh I see what you mean - like an initMetrics that the processor calls from its init. I like that much better.

ableegoldman: exactly

rodesai:
I'm going to merge this because the build failure is in smoke and totally unrelated to this change

PR #295, opened 2024-05-10
wrap thread pool in semaphore to block on full queue

rodesai:
By default, the thread pool executor does not block when the queue is full. Moreover, it does not have an option to block. The best it supports is running the task on the main thread. The risk with blocking is that you could deadlock if all your processing threads are blocked. For streams, that's not as much of a concern because the thread will be removed from the group if that happens, and should only happen if the user's processor blocks. On the other hand, running from the main thread is not ideal because you could get unlucky and run a very slow task on the main thread and starve the topology. So instead, we wrap the pool in a semaphore to block before enqueuing a task

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
@@ -15,6 +15,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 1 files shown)


ableegoldman (commented):
LGTM although we can just remove the `BlockingQueue` that we pass in to the executor now, since the only reason for doing that was to try and get blocking behavior

rodesai:
> LGTM although we can just remove the `BlockingQueue` that we pass in to the executor now, since the only reason for doing that was to try and get blocking behavior

`ThreadPoolExecutor` expects a `BlockingQueue`, so we don't have a choice there. I could create it without a limit though, since the semaphore protects us from over-filling it.

ableegoldman (approved):
> I could create it without a limit though

yeah sorry that's what I meant. LGTM!

PR #296, opened 2024-05-10
Unregister store flush listener in the store's close

ableegoldman:


rodesai (approved):
LGTM!

Are there cases where `close` would not get called? E.g. if some exception gets thrown before the stream thread gets to `close`? I still think this change is worthwhile, but maybe we need to use some thread-unique id to look up the connector as well to deal with that case.

ableegoldman:
That's a reasonable question...I'm pretty sure close always has to be called or else it's a pretty drastic bug in Streams. That said, it makes sense to be as careful as possible and use a unique id. I think that's worth following up on, although I'm just going to go ahead and merge this since I think it's the right thing to do anyways and is cleaner/less error prone even when we do it based on unique id

PR #297, opened 2024-05-14
make sure to throw processing exceptions before prep finalize

rodesai:
Make sure that when finalizing a record, we throw processing exceptions before preping the event for finalization. Also updates the test to check that the causing exception for injected exceptions is the injected exception.

Before this patch, all exceptions would result in an IllegalStateException due to event state machine violation.


ableegoldman (approved):
Ah, good catch, LGTM

PR #298, opened 2024-05-14
Handle TaskMigratedException and clean up Throwables

ableegoldman:
Split up the error handling by separating into StreamsExceptions that arise from processing an event, and fatal exceptions that cover things like errors in the framework itself and checked exceptions.

Modified the async task from a Runnable to a Supplier<StreamsException> so we have a natural way to split up errors based on whether they're "normal" processing errors we can expect and handle (in which case we catch it and return it) vs uncaught/unexpected/checked exceptions that are immediately fatal.

Also split up the error handling path itself, so we don't add processing exceptions to the finalizing queue AND track them as an "uncaught AsyncThreadPool exception", it's either-or. This helps distinguish the fatal errors that should be acted upon immediately vs the non-fatal errors that might just mean rejoining the group (TaskMigratedException) or could, in the future, have some other kind of special error handling


rodesai (approved):
LGTM!

PR #299, opened 2024-05-15
Dont finalize an async event if it failed to process.

rodesai:
Dont finalize an async event if it failed to process. We should only call addFinalizableEvent for successfully processed events. Failed events should be called with addFailedEvent, and be transition to FAILED state.


ableegoldman (approved):
LGTM but it should also be covered by https://github.com/responsivedev/responsive-pub/pull/298 so we can merge this or just wait for that PR to be merged (just need to rebase it first)

PR #300, opened 2024-05-17
deal with unexpected async exceptions from pool correctly

rodesai:
This patch fixes how we deal with unexpected/fatal exception. Previously we were passing them through a completion stage generated by calling `handle` on the event's future. `handle` will not preserve the result and exception type of the original future. Instead, it's result/exception is the return value or exception thrown by the function passed to `handle`. Therefore unexpected exceptions passed to `handle` were being dropped. Instead, we use `whenComplete` which does preserve the exception.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
@@ -192,7 +192,7 @@ public class AsyncThreadPool {
           .whenComplete((r, t) -> {
             inFlightForTask.remove(event);
           })
-          .handle((processingException, fatalException) -> {
+          .whenComplete((processingException, fatalException) -> {
 
             if (fatalException == null) {
               if (processingException == null) {
@@ -206,7 +206,6 @@ public class AsyncThreadPool {
             // Once we've successfully placed a failed event in the finalizing queue
             // then there's nothing more to handle by the async thread pool since the
             // StreamThread will process the exception from here
-            return null;
           })
           .exceptionally(fatalException -> {
             // do this alone & in separate stage to ensure we always catch a fatal exception, even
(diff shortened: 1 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -192,7 +192,7 @@ public <KOut, VOut> void scheduleForProcessing(
           .whenComplete((r, t) -> {
             inFlightForTask.remove(event);
           })
-          .handle((processingException, fatalException) -> {
+          .whenComplete((processingException, fatalException) -> {
```
rodesai, line 195: this is the main fix, as explained in the commit msg

ableegoldman: Good catch, this was my mistake -- I thought `handle` just let you change the return value but didn't realize it would wipe out the exception as well.

FYI you can probably remove the comment on line 206 (or move it up here if you think it has value, I went overboard with the comments I know)

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -206,15 +206,17 @@ public <KOut, VOut> void scheduleForProcessing(
...
             // do this alone & in separate stage to ensure we always catch a fatal exception, even
             // if we somehow hit another exception while handling an exception in a previous stage
             fatalExceptions.computeIfAbsent(
                 inFlightKey,
                 k -> new FatalAsyncException("Uncaught exception while handling", fatalException));
-            return fatalException;
+            if (fatalException instanceof RuntimeException) {
+              throw (RuntimeException) fatalException;
+            }
+            throw new RuntimeException(fatalException);
```
ableegoldman, line 219: nit: extract the `new FatalAsyncException("Uncaught exception while handling", fatalException))` above and then just throw that here.

The main reason I added that was to help clean up all this RuntimeException casting and wrapping

ableegoldman (approved):
some super minor suggestions but LGTM

PR #302, opened 2024-05-23
bump AK to 3.7 and default DSL stores to Responsive

agavra:
This PR allows users to literally migrate with a single line code change by automatically configuring `ResponsiveKafkaStreams` to use the new `ResponsiveDslStoreSuppliers` class. The main change in this PR is in `kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java`. In order to use this we have to depend on apache kafka 3.7.

This PR also wires in Session stores, which previously did not have an equivalent store supplier in `Stores` static class. 

For testing, you can see that the new test doesn't specify any `Materialized` or `Stores` but it automatically populated data in MongoDB. The one drawback of this approach is that it risks collisions with names of stores in the same MongoDB instance. We need to figure out how to prevent that (perhaps by prepending the applicationID to the db name).

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
@@ -34,6 +34,7 @@ import static org.apache.kafka.streams.StreamsConfig.NUM_STREAM_THREADS_CONFIG;
 import dev.responsive.kafka.api.async.internals.AsyncThreadPoolRegistry;
 import dev.responsive.kafka.api.config.CompatibilityMode;
 import dev.responsive.kafka.api.config.ResponsiveConfig;
+import dev.responsive.kafka.api.stores.ResponsiveDslStoreSuppliers;
 import dev.responsive.kafka.internal.clients.ResponsiveKafkaClientSupplier;
 import dev.responsive.kafka.internal.config.ConfigUtils;
 import dev.responsive.kafka.internal.config.InternalSessionConfigs;
@@ -271,7 +272,6 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
   ) {
     final Properties propsWithOverrides = new Properties();
 
-
     final InternalSessionConfigs.Builder internalConfBuilder = new InternalSessionConfigs.Builder()
         .withSessionClients(sessionClients)
         .withStoreRegistry(storeRegistry)
@@ -312,6 +312,11 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
       LOG.error(errorMsg);
       throw new ConfigException(errorMsg);
     }
+
+    propsWithOverrides.putIfAbsent(
+        StreamsConfig.DSL_STORE_SUPPLIERS_CLASS_CONFIG,
+        ResponsiveDslStoreSuppliers.class.getCanonicalName()
+    );
 
     return propsWithOverrides;
   }
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
@@ -315,7 +315,7 @@ public class ResponsiveKafkaStreams extends KafkaStreams {
 
     propsWithOverrides.putIfAbsent(
         StreamsConfig.DSL_STORE_SUPPLIERS_CLASS_CONFIG,
-        ResponsiveDslStoreSuppliers.class.getCanonicalName()
+        ResponsiveDslStoreSuppliers.class.getName()
     );
 
     return propsWithOverrides;
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
@@ -28,25 +28,25 @@ public final class ResponsiveSessionParams {
 
   private final TableName name;
   private final SessionSchema schemaType;
-  private final long inactivityGapMs;
-  private final long gracePeriodMs;
   private final long retentionPeriodMs;
-
-  private long numSegments;
+  private final long numSegments;
 
   private ResponsiveSessionParams(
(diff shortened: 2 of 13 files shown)


kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveKafkaStreamsIntegrationTest.java
```diff
@@ -0,0 +1,171 @@
...
+        new KeyValueTimestamp<>("key", "c", 3_000L),
+        new KeyValueTimestamp<>("STOP", "ignored", 18_000L)
+    );
+    final CountDownLatch outputLatch = new CountDownLatch(1);
+
+    final StreamsBuilder builder = new StreamsBuilder();
+    final KStream<String, String> input = builder.stream(inputTopic());
+    input
+        .groupByKey()
+        .count(Named.as(COUNT_TABLE_NAME))
```
agavra, line 119: *Note:* `COUNT_TABLE_NAME` doesn't actually get used as the store name! Pretty crazy and unfortunate because I think you actually need to use `Materialized.as` in order to get the name.

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveKeyValueStoreIntegrationTest.java
```diff
@@ -187,38 +169,4 @@ public void shouldMatchRocksDB() throws Exception {
     }
   }
 
-  private Map<String, Object> getMutablePropertiesWithStringSerdes() {
```
agavra, line 190: this was just moved to `IntegrationTestUtil` so I could reuse it

apourchet (approved):
👍

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveKafkaStreamsIntegrationTest.java
ableegoldman: Ugh that is unfortunate...why would we not use actual processor names for state stores??? I honestly thought you were mistaken and had to check this for myself, it just seems too stupid. And we can't ever "fix" this without it being a backwards incompatible change. That sucks

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
```diff
@@ -75,10 +76,6 @@ public long numSegments() {
     return this.numSegments;
   }
 
-  public long gracePeriodMs() {
```
ableegoldman, line 78: Why get rid of this? I know it's unused, just thought it would be nice to leave getters for users. (I'm also fine with removing it if you want to, just wanted to explain why it's there despite being unused)

ableegoldman (approved):
LGTM, few nits that you can take or leave as you want

ableegoldman:
To address this (just reiterating a slack conversation for posterity):


> The one drawback of this approach is that it risks collisions with names of stores in the same MongoDB instance. We need to figure out how to prevent that (perhaps by prepending the applicationID to the db name).

Background:
To clarify, the risk of naming collisions was always present -- for example, a user who doesn't care about materialization or IQ might simply pass in the name "store" for multiple stateful apps, and we would see a collision. We would have to trust in users to select a globally unique name. However, after this change, users can inject Responsive stores without explicitly configuring a name at all, in which case it would default to the naming scheme `<operator>-STATE-STORE-<id>` -- for example, `KSTREAM-AGGREGATE-STATE-STORE-0000000001`. 

Note the `<operator>` parameter reflects the DSL operator type and is unaffected by naming the operators themselves. It also does not reflect the literal/logical DSL operator  (eg `count` vs `reduce` vs `aggregate`) but is instead based on the physical operator type (ie `aggregate`). This means there are only a handful of possible values for the `<operator>`  parameter of the store name -- I believe there's only 5: aggregate, s-t join, s-s join, t-t join, and fk join. 

On the other hand, the `<id>` is a unique integer for that node which counts up from the first node of the first subtopology (with an index burned for each node regardless of whether it's already named by the user). So for a naming collision to occur, two applications would have to have the stateful operator at exactly the same index in the overall topology. The odds of this are definitely not zero, but it's not likely to be a real risk until an org starts to have a large number of stateful applications or run with similar logic

Solution thoughts:
Note that we can't just tack on the appId to the state store name, or modify the state store name itself in any way, since this will affect the changelog topic name as well. Any solution will need to modify the table name without affecting the underlying state store name. 

A simple solution would be to prepend the appId to the state store name, but this risks running up against the relatively small character limit for MongoDB table names. Another option would be to generate a globally unique id, perhaps via the operator, which gives us the desired org-level scope. We can inject it during startup by adding a config for table name prefix. Of course, this requires an operator connection, which is not ideal. 

We could implement an incomplete but forwards compatible solution to help in the interim, by going the appId route if and only if the conjugated appId-storeName string is within the known character limits. This  approach could be continued even after transitioning to some other scheme for names that exceed the limit.

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -313,6 +313,11 @@ private static Properties propsWithOverrides(
       throw new ConfigException(errorMsg);
     }
 
+    propsWithOverrides.putIfAbsent(
+        StreamsConfig.DSL_STORE_SUPPLIERS_CLASS_CONFIG,
+        ResponsiveDslStoreSuppliers.class.getCanonicalName()
```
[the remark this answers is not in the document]
agavra: this is something I always forget 😆 I think in this case you're right, `getName` is more appropriate

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveSessionParams.java
agavra: the `DslSessionParams` only specifies retention, not grace+inactivity, meaning I can't compute the grace period in the new constructor

ableegoldman (approved):
LGTM

PR #303, opened 2024-06-01
clear recorded offsets on unsubscribe and handlePartitionsLost

rodesai:
This patch clears all recorded offsets in OffsetRecorder when Consumer.unsubscribe or RebalanceListener.handlePartitionsLost are called. In both cases, we know streams is reinitializing so we should clear any offsets we have recorded. If we don't clear them, then when streams finishes reinitializing and goes to do its first commit, it will try to find stores for those partitions and commit those offsets


ableegoldman (approved):
Good catch -- LGTM

PR #304, opened 2024-06-05
make sure to deregister listener when store init fails

rodesai:
This patch makes store init more "transactional" by deregistering the flush listener if the wrapped store's init method throws. This way it either succeeds and the listener is registered and the store initialized. Or it fails and there is no listener registered.

If we dont do this, the task for which we threw can never be assigned to this stream thread again as it will always raise an IllegalStateException.


ableegoldman (approved):
Good catch, LGTM

PR #305, opened 2024-06-05
make sure we always call postFinalize

rodesai:


agavra (approved):

ableegoldman (approved):
Oh nice, I remember thinking a while back that we should be unblocking keys even in the event of failure, but this slipped my mind. Wasn't thinking about the need to also remove it from the set of pending events but that makes sense too. LGTM

ableegoldman (approved):
LGTM

PR #306, opened 2024-06-07
WIP: add wrapper around AsyncProcessor to enable async processing engine only with async threads configured

ableegoldman:
Add a wrapper around the AsyncProcessor class and only instantiate it if the user has configured a nonzero thread pool size. This allows turning async on/off with just a config, without needing to add/remove the wrapper around the ProcessorSupplier as well.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncFixedKeyProcessorSupplier.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncFixedKeyProcessorSupplier.java
@@ -16,10 +16,9 @@
 
 package dev.responsive.kafka.api.async;
 
-import static dev.responsive.kafka.api.async.internals.AsyncProcessor.createAsyncFixedKeyProcessor;
 import static dev.responsive.kafka.api.async.internals.AsyncUtils.initializeAsyncBuilders;
 
-import dev.responsive.kafka.api.async.internals.AsyncProcessor;
+import dev.responsive.kafka.api.async.internals.MaybeAsyncProcessor;
 import dev.responsive.kafka.api.async.internals.stores.AbstractAsyncStoreBuilder;
 import java.util.HashSet;
 import java.util.Map;
@@ -75,8 +74,8 @@ public class AsyncFixedKeyProcessorSupplier<KIn, VIn, VOut>
   }
 
   @Override
-  public AsyncProcessor<KIn, VIn, KIn, VOut> get() {
-    return createAsyncFixedKeyProcessor(userProcessorSupplier.get(), asyncStoreBuilders);
+  public MaybeAsyncProcessor<KIn, VIn, KIn, VOut> get() {
+    return MaybeAsyncProcessor.createFixedKeyProcessor(userProcessorSupplier.get(), asyncStoreBuilders);
   }
 
   @Override
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
@@ -69,12 +69,7 @@ import org.slf4j.Logger;
  * -Coordinates the handoff of records between the StreamThread and AyncThreads
  * -The starting and ending point of all async events -- see {@link AsyncEvent}
  */
-public class AsyncProcessor<KIn, VIn, KOut, VOut>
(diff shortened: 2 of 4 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -69,12 +66,18 @@
...
+  private final String logPrefix;
+  private final Logger log;
+
+  private final String streamThreadName;
+  private final String asyncProcessorName;
+  private final TaskId taskId;
+
+  private final AsyncThreadPool threadPool;
+  private final SchedulingQueue<KIn> schedulingQueue;
+  private final FinalizingQueue finalizingQueue;
```
ableegoldman, line 80: The nice thing about this change is that we only instantiate the AsyncProcessor after the processor is initialized, so we can actually make all these variables final. I was hoping to do this anyway

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -69,12 +66,18 @@
  * -Coordinates the handoff of records between the StreamThread and AyncThreads
  * -The starting and ending point of all async events -- see {@link AsyncEvent}
  */
-public class AsyncProcessor<KIn, VIn, KOut, VOut>
-    implements Processor<KIn, VIn, KOut, VOut>, FixedKeyProcessor<KIn, VIn, VOut> {
+public class AsyncProcessor<KIn, VIn, KOut, VOut> {
```
ableegoldman, line 69: Since this no longer implements the `Processor` interfaces, I think I'll rename it to "AsyncProcessingEngine" or something to that effect (suggestions welcome). And then rename the new wrapper class that creates/delegates to this class to "AsyncProcessor". But I'll wait until the PR has been reviewed/approved to do the renamings since that might impact a bunch of code

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/MaybeAsyncProcessor.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.kafka.streams.processor.api.ProcessorContext;
+import org.apache.kafka.streams.processor.api.Record;
+import org.apache.kafka.streams.processor.internals.InternalProcessorContext;
+
+/**
+ * This class is in charge of two things:
+ * 1. deduplicating and delegating code between the regular and fixed-key processor versions
+ * 2. disabling async processing altogether if the thread pool size is 0
+ */
+public class MaybeAsyncProcessor<KIn, VIn, KOut, VOut>
```
ableegoldman, line 38: not the final name for this -- plan to rename it to "AsyncProcessor" or similar. See [this comment](https://github.com/responsivedev/responsive-pub/pull/306/files#r1630614463)

kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncFixedKeyProcessorSupplier.java
```diff
@@ -75,8 +74,8 @@ private AsyncFixedKeyProcessorSupplier(
   }
 
   @Override
-  public AsyncProcessor<KIn, VIn, KIn, VOut> get() {
-    return createAsyncFixedKeyProcessor(userProcessorSupplier.get(), asyncStoreBuilders);
+  public MaybeAsyncProcessor<KIn, VIn, KIn, VOut> get() {
+    return MaybeAsyncProcessor.createFixedKeyProcessor(userProcessorSupplier.get(), asyncStoreBuilders);
```
ableegoldman, line 78: the one thing I haven't implemented yet is that we initialize the store builders as AsyncStoreBuilders right from the start. I still need to push that back so that we don't end up wrapping the user's store suppleirs in the async stuff if that's not enabled

ableegoldman:
Shelving this for now since the approach becomes rather messy once we fix the AsyncStoreBuilder issue (essentially we'll need to provide wrapper classes for every single async store type, which are already wrappers for each of the actual StateStore types)

Will probably pick this back up again at some point in the future, but in the meantime we advise users to simply branch their DSL operators based on the configuration in their own code. For example:

```
boolean enableAsync = asyncThreadPoolSize > 0;
if (enableAsync) {
    kstream
        .process(createAsyncProcessorSupplier(new MyProcessorSupplier))
        .to(OUTPUT_TOPIC);
} else {
    kstream
        .process(new MyProcessorSupplier)
        .to(OUTPUT_TOPIC);
}
```

PR #307, opened 2024-06-07
Disable async processing when thread pool size is 0 via static gate

ableegoldman:
Alternative to [pull/306](https://github.com/responsivedev/responsive-pub/pull/306)

Unfortunately the approach taken in the above PR gets considerably messier than the first draft showed, since we need to "shut off" the async store wrappers in addition to just the async processor itself. And unfortunately, we can't delay that until #init -- at least not as easily. 

This PR is definitely very hacky and probably not a great long-term solution for a real product -- but it does solve the immediate problem in a quick and neat way.


ableegoldman:
Too risky -- we're advise users branch their DSL operators based on their configuration in their own code

PR #308, opened 2024-06-07
handle cancelled tasks in thread pool future handlers

rodesai:
We shouldn't record a CancellationException as a fatal async exception. The Async Thread Pool's future handler is called with a CancellationException whenever a task is cancelled, which is done on the normal path when a task is closed dirty. If we record the exception, then the next time the task is assigned we will immediately throw.


ableegoldman (approved):
Makes sense

PR #312, opened 2024-06-20
add an integration test for foreign key joins

rodesai:


rodesai:
I'm going to merge this to get the other fk join patch unstuck

PR #313, opened 2024-06-21
Implement window stores with retainDuplicates for stream-stream joins

ableegoldman:
We need to support the `retainDuplicates` configuration of window stores to enable stream-stream joins, which essentially means storing multiple values for the same key (or technically, the same key+timestamp combination).

This required quite a few changes in these areas:
-Parameterize the `CommitBuffer` on `retainDuplicates` by changing the `SizeTrackingBuffer` from a class to an interface, and introducing a new `DuplicateKeyBuffer` implementation for the duplicates case. Also had to do some refactoring of the buffer APIs to make sure duplicates didn't get swallowed up
-Parametrize the `LocalRemoteKvIterator` on `retainDuplicates` so that it doesn't apply CommitBuffer changes on top of remote records with the same key, but rather returns them all
-Add a `values` field to `WindowDoc` that stores multiple values in a list. When `retainDuplicates` is true, the regular `byte[] value` field is null, and when it's false the new `List<byte[] values` field is null instead. Use Mongo's `Updates.push` operation to simply append to the list on write so we don't have to read the entire thing
-Introduce a new iterator type to parse the `List<byte[]> values` field and flatten the results into a sequence of KVs for each value in the list (I called this "DuplicateKeyListValueIterator" but better naming suggestions welcome)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -375,17 +388,31 @@ public WriteModel<WindowDoc> insert(
...
-            Filters.lte(WindowDoc.EPOCH, epoch)
-        ),
-        Updates.combine(
-            Updates.set(WindowDoc.VALUE, value),
-            Updates.set(WindowDoc.EPOCH, epoch)
-        ),
-        UPSERT_OPTIONS
-    );
+    if (retainDuplicates) {
+      return new UpdateOneModel<>(
```
rodesai, line 392: I think you want to do an insert, not an update here. This will update the existing timestamp+key if it exists, when really we want to only ever insert.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -403,7 +411,13 @@ public WriteModel<WindowDoc> delete(
       final int kafkaPartition,
       final WindowedKey windowedKey
   ) {
+    if (retainDuplicates) {
+      throw new IllegalStateException("Should never attempt to issue a delete with duplicates");
```
agavra, line 415: (not a blocker) why? couldn't the user have a custom processor that does this and expects it to delete all the values associated with the same key?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/DuplicateKeyBuffer.java
```diff
@@ -0,0 +1,134 @@
...
+    size = 0;
+    buffer.clear();
+  }
+
+  @Override
+  public Result<K> get(final K key) {
+    if (!buffer.containsKey(key)) {
+      return null;
+    } else {
+      // point lookups are defined for stores with duplicates, so just return the first element
```
agavra, line 76: did you mean to say "are _not_ defined"?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/DuplicateKeyBuffer.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Override
+  public void put(final K key, final Result<K> value) {
+    // ignore tombstones since deletes are a no-op for duplicate stores
+    if (!value.isTombstone) {
+      bytes += value.size(extractor);
+      ++size;
+
+      buffer.computeIfAbsent(key, k -> new LinkedList<>()).add(value);
```
agavra, line 60: Typically unless there are going to be insertions/deletions from the middle of the list `ArrayList` is almost always a better choice. The memory overhead per-element in a LinkedList is so significant that the memory savings of unused elements in the array list are shadowed (not to mention performance of iterating a LinkedList is significantly worse since there's no necessary memory cohesion)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/LocalRemoteKvIterator.java
```diff
@@ -121,13 +133,21 @@ private KeyValue<K, byte[]> advance() {
...
-      // value is more recent, so we return that and advance
-      // both - either way, the value from remote.next()
-      // should not be returned
-      remote.next();
+
+      if (retainDuplicates) {
+        // if keys are the same and duplicates are allowed, then we arbitrarily
+        // pick one of them to return (in this case the remote) but it doesn't
+        // really matter which
+        return remote.next();
```
agavra, line 141: Took me a while to realize that we're supposed to return _both_ it's just the order that doesn't matter. Can we add that as a comment?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveWindowStore.java
```diff
@@ -110,7 +110,9 @@ public void init(final StateStoreContext storeContext, final StateStore root) {
           window -> window.windowStartMs >= minValidTimestamp()
       );
 
-      numBloomFilterWindows = config.getInt(WINDOW_BLOOM_FILTER_COUNT_CONFIG);
+      numBloomFilterWindows = params.retainDuplicates()
+          ? 0 // disable bloom filters for stream-stream join/duplicate stores
```
agavra, line 114: why don't bloom filters work in this scenario? we should be able to know if there are no values in the remote, right?

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/DuplicateKeyListValueIterator.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private static class WindowResult {
+    private final WindowedKey key;
+    private final List<byte[]> values;
+    private int valueIndex = 0;
+
+    public WindowResult(final WindowDoc windowDoc) {
+      this.key = windowedKey(windowDoc.getKey());
+      this.values = windowDoc.getValues();
+      System.out.printf("SOPHIE: WindowDoc result for key=%s has %d values\n", key, values.size());
```
agavra, line 84: 😉

agavra (approved):
LGTM! Nice to complete the parity with streams :)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -43,6 +43,7 @@
 import dev.responsive.kafka.internal.db.partitioning.Segmenter.SegmentPartition;
 import dev.responsive.kafka.internal.db.partitioning.WindowSegmentPartitioner;
 import dev.responsive.kafka.internal.stores.RemoteWriteResult;
+import dev.responsive.kafka.internal.utils.DuplicateKeyListValueIterator;
```
rodesai, line 46: `MongoWindowedTable::fetch` for point lookups returns null for duplicate stores - is that what we want? Or should we have the same behaviour as CommitBuffer (return the first value)

rodesai (commented):
Overall LGTM. Couple questions inline

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
```diff
@@ -198,6 +201,7 @@ public MongoWindowedTable(
     this.name = name;
     this.partitioner = partitioner;
     this.timestampFirstOrder = timestampFirstOrder;
+    this.retainDuplicates = partitioner.retainDuplicates();
```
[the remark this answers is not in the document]
ableegoldman: short answer: yes, it should be part of the table not the partitioner
long answer: right now it's a huge pain to get anything passed into the tables because of the way they get built from the table specs. It takes a LOT to add even one parameter this way -- see for example this PR where I just wanted the changelog topic name but piping it through was such a huge mess that [Almog suggested just tacking it onto the partitioner ](https://github.com/responsivedev/responsive-pub/pull/251#discussion_r1504573937). Obviously this situation is untenable since we can't throw everything into the partitioner, but it's nontrivial tech debt for another time. Hence I stuck with the partitioner approach for now

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: Ah, good catch, no we should just return the first value (though arguably it's not really a defined API for duplicate stores and I would not recommend anyone use it. Of course I'm sure some people still do anyways 🤷‍♀️ )

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/DuplicateKeyListValueIterator.java
ableegoldman: whoops haha

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveWindowStore.java
ableegoldman: there's a few reasons but the main one is that stream-stream joins only use range queries whereas the bloom filters are only implemented for/help with the point lookups.

I suppose you could argue we should still enable them in case someone is using a window store with duplicates from a custom processor, which is rare but theoretically valid. However the point lookup API is essentially not defined for duplicate window stores due to the restrictions of the store interface, so I really doubt anyone is ever using point lookups in this case (and if they are, I'm highly skeptical of what they're doing and whether it should just be a regular/non-duplicate store)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/LocalRemoteKvIterator.java
ableegoldman: ah, yes that's fair, will do

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/DuplicateKeyBuffer.java
ableegoldman: good points, thanks for catching this

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoWindowedTable.java
ableegoldman: I mean...they can expect that, and it does make sense, but it's not how the built-in stores in Kafka Streams work so I figured we should follow the existing semantics instead of correcting them

In general the window store with duplicates isn't very well defined or thought out since it was never really supposed to be for users to plug into their processors but rather a hack to allow reusing the WindowStore for the new (at the time) stream-stream join operator. We don't explicitly forbid anyone from trying to use this but there's not much interest out there. In the end deletes are sort of like point lookups, in that the API is built around the non-duplicate stores and just happened to be inherited without any kind of semantic guarantees.

ableegoldman: Btw we throw an exception here because the delete should be "stopped" in one of the above layers like the ResponsiveWindowStore, so the call should never reach the MongoWindowedTable. ie it's not that it's invalid for a user to attempt a delete (the store just treats it as a no-op), rather it's invalid for our code to attempt this since we expect deletes to be intercepted before this

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/WindowDoc.java
```diff
@@ -47,10 +50,12 @@ public WindowDoc() {
   public WindowDoc(
       @BsonProperty(ID) BasicDBObject id,
       @BsonProperty(VALUE) byte[] value,
+      @BsonProperty(VALUES) List<byte[]> values,
```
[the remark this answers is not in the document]
ableegoldman: I was less concerned with the overhead of the list and more about maintaining a clean separation so we can make sure the correct column has data for the given store type

For one thing, if we use a list for both, then for non-duplicates we presumably want to overwrite the entire list on each update or pop the current value then push the new one, which both feel kind of silly/inefficient (just my intuition, maybe it's not that bad 🤷‍♀️ ). But we definitely don't want an ever-increasing list of values for non-duplicate stores, since it can get quite long with many frequent updates in an aggregation window. So we'd still have to branch on `retainDuplicates` on the write path. 

For another thing, and maybe this is a little paranoid, but what if someone has a store with many duplicates and at some point they accidentally run their app with `retainDuplicates` set to false? That could easily wipe out their existing data if it starts overwriting these lists rather than pushing new items. Keeping the fields separate prevents this from happening since each store type gets its own field

PR #314, opened 2024-06-24
model mongodb keyvalue keys as strings to support range scans

rodesai:
This patch changes the mongodb key representation to a string from a byte array. The string is a base64 encoded representation of the key bytes. The specific encoding uses a character table that guarantess that the byte-order of the keys is preserved in the mongo string sort order.

This change is required to support range scans in mongodb. MongoDB orders binary fields first by length, and then by content for fields of the same length. This is not the order required/assumed by Kafka Streams.

There is likely a performance and space penalty for this change, as the base64 encoded keys are 8/6 the size of the binary keys. If its a problem, then for use cases that do not require range scans we can add a config that toggles the key representation. However I don't think its worth the added complexity just yet. We can revisit if it actually manifests as a problem.

Extending this change to other store types is deferred to a follow-on patch.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -21,6 +21,7 @@ import static dev.responsive.kafka.internal.stores.ResponsiveStoreRegistration.N
 import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
 import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
 
+import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
@@ -36,12 +37,14 @@ import com.mongodb.client.model.WriteModel;
 import com.mongodb.client.result.UpdateResult;
 import dev.responsive.kafka.internal.db.MongoKVFlushManager;
 import dev.responsive.kafka.internal.db.RemoteKVTable;
+import dev.responsive.kafka.internal.utils.Iterators;
 import java.time.Instant;
 import java.util.Date;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.configuration.CodecRegistry;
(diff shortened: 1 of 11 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/OrderPreservingBase64Encoder.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.commons.codec.binary.Base64;
+
+public class OrderPreservingBase64Encoder {
+  private static final Base64 encoder = Base64.builder()
+      .setEncodeTable(
+          encoderCharset("+/0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"))
+      .get();
```
agavra, line 11: since this is a custom b64 encoding anyway (meaning we can't use online tools to decode it), I'm wondering why we're limiting ourselves to just b64? couldn't we do the same thing with a wider charset?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -68,6 +71,7 @@ public MongoKVTable(
       final CollectionCreationOptions collectionCreationOptions
   ) {
     this.name = name;
+    this.keyCodec = new StringKeyCodec();
```
agavra, line 74: let's add comments re: why we decided to add this and why we're not using byte[]

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -140,8 +144,19 @@ public KeyValueIterator<Bytes, byte[]> range(
       final Bytes to,
       final long minValidTs
   ) {
-    // NOTE: Make sure to handle filtering of tombstones in the remote range scan.
-    throw new UnsupportedOperationException();
+    final FindIterable<KVDoc> result = docs.find(
```
agavra, line 147: I suspect tests for this are in your FK join PR?

agavra (approved):
LGTM! Thanks @rodesai

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/OrderPreservingBase64Encoder.java
rodesai: Mongo strings are UTF-8 encoded. So the only charset size that would actually reduce overhead is 7 bits (basically full ascii). Larger bit-sizes would use enough additional bits to do the utf-8 encoding that the overhead is actually worse (for example, to encode 8-11 bit characters in utf-8 you have to use 16 bits). I'm hesitant to use the full ascii charset because
1. I'm not sure how mongo's find will or wont interpret the special characters.
2. We'd have to write the encoding ourself, which isn't that hard, but its nice to just have a library to use instead.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
rodesai: I didn't see that we had kvtable unit tests. I'm adding some cases now.

PR #315, opened 2024-06-26
handle expiry for mongo store

rodesai:
- Add a timestamp field to the key value document
- Set the timestamp on put and unset on delete
- Filter get/range/all using the provided min valid time


agavra (approved):
Thanks @rodesai this looks good to me, though I was looking at the code and I couldn't find where we actually create a general TTL index (we have a tombstone TTL index) -- we should make sure that we implement that as well, which can just be an index on this new timestamp field.

PR #316, opened 2024-06-27
Use string keys for session and window stores

rodesai:
This patch changes window/session stores to use string keys like we did in https://github.com/responsivedev/responsive-pub/pull/314 to support correct range scans across keys. Note that currently range scans across keys are only used by kafka streams for IQ, and are currently not supported by responsive. However it would be good to make this change now to keep the key format consistent and use a baseline that will be functionally correct if we want to turn iq on. If the format causes performance problems we can always add a config to revert back to the binary key format.

This patch also fixes a bug in the window range scan implementation when timestamp order is enabled. In particular, when using timestamp order we need to filter on the specific key being scanned.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KeyCodec.java
```diff
@@ -0,0 +1,9 @@
+package dev.responsive.kafka.internal.db.mongo;
+
+import org.apache.kafka.common.utils.Bytes;
+
+interface KeyCodec {
```
agavra, line 5: we should consider leveraging custom codecs in the Mongo Java driver (https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/data-formats/codecs/) so that we just automatically get this and don't forget using it somewhere.

I haven't explored whether this is possible yet - but if it's possible could be a worthwhile change

agavra (approved):
LGTM, one comment to see if we can leverage MongoDB changelogs

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/KeyCodec.java
rodesai: I'll tackle this in a follow-up. It covers both kv and window stores.

rodesai: I'll tackle this in a follow-up. It covers both kv and window stores.

PR #317, opened 2024-06-27
add an in-memory responsive store tye

rodesai:
Adds an in-memory responsive store type that stores all data in memory. It never stores an offset, so on a restore the entire store is restored from offset 0. This is meant to be used for test environments where we dont want to run a remote store.


agavra (approved):
Skimmed -- since this is used mostly for testing I didn't do a deep review. Let me know if there's anything in particular you want me to review.

PR #328, opened 2024-07-02
Remove state store requirement for async processing

ableegoldman:
The general approach approach here is fairly straightforward: when there are no state stores we can use to hook into the `#flushCache` call before a commit, we simply hook into the actual clients' commit methods and flush the processors before proceeding with the commit. For ALOS apps, we flush via the consumer's `#commitSync` while with EOS we flush via the producer's `#commitTransaction`

To make this work cleanly, I extracted the async thread pool management from the ResponsiveKafkaClientSupplier/ResponsiveConsumer and put it into a new layer for all the async logic which only gets created when async processing is enabled. Similarly, we only create the async version of the relevant client and have it handle the async thread pool lifecycle -- in others, for EOS apps everything happens in the AsyncStreamsProducer and there is no AsyncStreamsConsumer, and vice versa for ALOS apps.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2024 Responsive Computing, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package dev.responsive.kafka.api.async.internals;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class AsyncThreadPoolRegistration {
+
+  private final AsyncThreadPool threadPool;
+
+  private final Map<AsyncProcessorId, Runnable> asyncProcessorsToFlush = new HashMap<>();
+
+  public AsyncThreadPoolRegistration(
+      final AsyncThreadPool threadPool
+  ) {
+    this.threadPool = threadPool;
+  }
+
+  public AsyncThreadPool threadPool() {
+    return threadPool;
+  }
+
+  public void registerAsyncProcessor(final AsyncProcessorId id, final Runnable flushProcessor) {
+    asyncProcessorsToFlush.put(id, flushProcessor);
+  }
+
+  public void unregisterAsyncProcessor(final AsyncProcessorId id) {
+    asyncProcessorsToFlush.remove(id);
+    threadPool.removeProcessor(id);
+  }
+
+  public void flush() {
+    // TODO: this is non-ideal for stateless apps since we're flush each processor instance
+    //  sequentially and might end up waiting for async results from one processor while there
+    //  are others ready and waiting on other processors/partitions.
+    //  We should loop over all processors and only wait/block when we really have nothing to do
+    //  (doesn't matter for stateful apps since this method is a no-op due to flushing
+    //  everything during #flushCache which comes earlier in the commit process)
+    asyncProcessorsToFlush.values().forEach(Runnable::run);
+  }
+
+  public void close() {
+    threadPool.shutdown();
+  }
+}
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
@@ -16,14 +16,18 @@
 
 package dev.responsive.kafka.api.async.internals;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import org.apache.kafka.streams.processor.TaskId;
 
 public class AsyncThreadPoolRegistration {
 
   private final AsyncThreadPool threadPool;
 
-  private final Map<AsyncProcessorId, Runnable> asyncProcessorsToFlush = new HashMap<>();
+  // Processors are maintained in topological order for each partition/task
+  private final Map<TaskId, List<Runnable>> taskToAsyncProcessorFlushers = new HashMap<>();
 
   public AsyncThreadPoolRegistration(
       final AsyncThreadPool threadPool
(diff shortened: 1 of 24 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/clients/AsyncStreamsProducer.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public void commitTransaction() throws ProducerFencedException {
+    flushAsyncProcessors.run();
+    super.commitTransaction();
+  }
+
+  @Override
+  public void close() {
+    // TODO: Producer is closed on reset under EOSv2 :/
```
ableegoldman, line 61: One thing I haven't decided yet is how we want to handle the case of a TaskMigrated/ProducerFenced/etc under EOSv2 in Streams. When these occur we "reinitialize" the StreamsProducer which actually calls close on the old Producer and creates a new one. This feels kind of silly to do for the async thread pool as well, so in theory we could try to avoid closing it and creating a new one somehow. Haven't yet thought through whether there are any correctness implications there (certainly we would need to at least flush/clear any pending work before recycling the async thread pool across a producer restart?)

ableegoldman: On second thought I think it's clear we do want to just shut down the current async pool and get a new one in case of a TaskMigratedException. So not an issue

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistry.java
```diff
@@ -89,7 +96,7 @@ public void shutdownAsyncThreadPool(final String streamThreadName) {
   }
 
   private void shutdownAsyncThreadPool(final String streamThreadName, final boolean fromStart) {
-    final AsyncThreadPool threadPool = streamThreadToAsyncPool.remove(streamThreadName);
+    final AsyncThreadPoolRegistration threadPool = streamThreadToAsyncPool.remove(streamThreadName);
```
rodesai, line 99: nit: rename to `registration`

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public void flush() {
+    // TODO: this is non-ideal for stateless apps since we're flush each processor instance
+    //  sequentially and might end up waiting for async results from one processor while there
+    //  are others ready and waiting on other processors/partitions.
+    //  We should loop over all processors and only wait/block when we really have nothing to do
+    //  (doesn't matter for stateful apps since this method is a no-op due to flushing
+    //  everything during #flushCache which comes earlier in the commit process)
+    asyncProcessorsToFlush.values().forEach(Runnable::run);
```
rodesai, line 54: we need to iterate on these in topology order right? Also, assuming we do that, then what happens if there is a stateless async processor before a stateful one? I think in that case the stateful processor may do some flush work when it's flushed here. Which I think is fine - just making sure I understand.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/AsyncStreamsKafkaClientSupplier.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Map;
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.streams.KafkaClientSupplier;
+import org.apache.kafka.streams.StreamsConfig;
+
+public class AsyncStreamsKafkaClientSupplier implements KafkaClientSupplier {
```
rodesai, line 31: this class is probably worth unit testing

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/AsyncStreamsConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+    this.asyncThreadPoolRegistry = asyncThreadPoolRegistry;
+    final var asyncThreadPoolRegistration = asyncThreadPoolRegistry
+        .startNewAsyncThreadPool(streamThreadName);
+    this.flushAsyncProcessors = asyncThreadPoolRegistration::flush;
+
+  }
+
+  @Override
+  public void commitSync(final Map<TopicPartition, OffsetAndMetadata> offsets) {
+    flushAsyncProcessors.run();
```
rodesai, line 62: I think the approach here breaks ALOS, because the producer has already been flushed at this point. I think we need to call producer.flush here.

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
ableegoldman: well technically no because these are all the same processor just from different partitions, but yeah we should probably make sure to flush in topological order to prepare for when we support more than one async processor per topology

As for how to deal with mixed stateless and stateful async processors, that's a good question which I'll want to think through more carefully but will punt on that here since it's really part of the "support multiple async processors" work and not in scope here. I will file a ticket or jot down this question to make sure we address it once we get to that 

(my gut reaction is that it's mostly fine since we re-flush the stateful processor here anyway, but we might also need to re-flush the Streams cache after/as part of flushing the stateful processor, since that might send some records downstream)

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/AsyncStreamsConsumer.java
ableegoldman: Good catch. Changed it to just do the async flushing inside/before the Producer#flush instead of the consumer/producer commitOffsets/commitTxn. Also simplified things so that the async threadpool lifecycle is always managed by the AsyncConsumer and the flushing is always managed by the AsyncProducer, and we always create both clients, regardless of eos.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/AsyncStreamsKafkaClientSupplier.java
ableegoldman: Since I ended up taking out the whole "create different clients based on eos" thing, it's pretty dead simple now. Can still add a unit test for it if you'd like but idk what's left that's worth testing

kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPoolRegistration.java
ableegoldman: Alright I made sure that we'll now flush processors within a task in topological order (luckily they're initialized in order so we can just throw them in a list). As for processors across tasks, they can go in any order (and eventually we could consider parallelizing these if it becomes a bottleneck)

PR #331, opened 2024-07-30
Support multiple/mixed async processors

ableegoldman:
Adds support for using async with any number and type of processors in the same topology, and adds an integration test with multiple stateful and stateless async processors in one app.


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncProcessor.java
```diff
@@ -254,7 +254,7 @@ private void completeInitialization() {
         streamThreadName,
         taskId.partition(),
         connectedStoreBuilders.values(),
-        this::flushPendingEventsForCommit
+        asyncThreadPoolRegistration::flushAllAsyncEvents
```
ableegoldman, line 257: This is actually the only change that was strictly necessary for supporting async, and it addresses the issue raised in this earlier comment by simply flushing all the async processors (stateless and stateful) in topological order from inside #flushCache, rather than flushing the stateful async processors in their individual #flushCache calls and then later flushing all processors during the producer#flush

rodesai (approved):
LGTM!

PR #332, opened 2024-08-05
allow deploying operators with labelselectors

agavra:
this patch allows running operator with `--labelSelector <label selector>` to specify which policies it should reconcile

### DIFF
--- a/operator/build.gradle.kts
+++ b/operator/build.gradle.kts
@@ -40,9 +40,12 @@ dependencies {
     annotationProcessor(libs.crd.generator.atp)
 
     testImplementation(testlibs.bundles.base)
+    testImplementation(testlibs.bundles.testcontainers)
     testImplementation("io.grpc:grpc-testing:1.53.0")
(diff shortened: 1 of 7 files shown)


operator/build.gradle.kts
```diff
@@ -40,9 +40,12 @@ dependencies {
     annotationProcessor(libs.crd.generator.atp)
 
     testImplementation(testlibs.bundles.base)
+    testImplementation(testlibs.bundles.testcontainers)
     testImplementation("io.grpc:grpc-testing:1.53.0")
     testImplementation("io.grpc:grpc-examples:0.15.0")
 
+    testImplementation("org.bouncycastle:bcprov-jdk15on:1.70")
+    testImplementation("org.bouncycastle:bcpkix-jdk15on:1.70")
```
agavra, line 48: these are needed to run a k8s client connected to k3s (they use EC certs)

rodesai (approved):
LGTM! Thanks for adding an integration test!

PR #343, opened 2024-09-06
fix mongo KV range scans

agavra:
range scans on KV stores didn't consider kafka partition - this patch fixes it by adding the kafka partition into the data model. while I don't love this since it'll make it more difficult to repartition a kafka topic, it will only be the case for topologies that use range or all on KV stores (something the DSL does not leverage, so I consider it an advanced use case)

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
@@ -170,7 +170,8 @@ public class MongoKVTable implements RemoteKVTable<WriteModel<KVDoc>> {
             Filters.gte(KVDoc.ID, keyCodec.encode(from)),
             Filters.lte(KVDoc.ID, keyCodec.encode(to)),
             Filters.not(Filters.exists(KVDoc.TOMBSTONE_TS)),
-            Filters.gte(KVDoc.TIMESTAMP, minValidTs)
+            Filters.gte(KVDoc.TIMESTAMP, minValidTs),
+            Filters.eq(KVDoc.KAFKA_PARTITION, kafkaPartition)
         )
     );
     return Iterators.kv(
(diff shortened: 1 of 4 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -185,7 +186,8 @@ public KeyValueIterator<Bytes, byte[]> range(
   public KeyValueIterator<Bytes, byte[]> all(final int kafkaPartition, final long minValidTs) {
     final FindIterable<KVDoc> result = docs.find(Filters.and(
         Filters.not(Filters.exists(KVDoc.TOMBSTONE_TS)),
-        Filters.gte(KVDoc.TIMESTAMP, minValidTs)
+        Filters.gte(KVDoc.TIMESTAMP, minValidTs),
+        Filters.eq(KVDoc.KAFKA_PARTITION, kafkaPartition)
```
agavra, line 190: we should consider adding a configuration that will create an index in MongoDB on the kafka partition

agavra: #344

rodesai (approved):
LGTM!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
ableegoldman: well we're going to have to get rid of this if we want to support partition scaling. That's why I went with the approach (long ago) of using the partitioner to figure out the kafka partition based on the key

Unfortunately I never found the time to finish that PR, so it's on me 

Well technically the PR was finished but you requested a change that was completely reasonable and that's when I had to pivot to whatever it was that was more important -- point being I can dig that old PR up and finally finish it if we want to do this and it needs to happen ASAP. Thoughts?

apurvam: An aside: how much work is left on that PR @ableegoldman ?

PR #345, opened 2024-09-09
add a new regression testing framework

agavra:
### Overview

This PR introduces the structure for a new regression testing framework that we plan to run with antithesis testing. This strategy of testing compares the output of a `ResponsiveKafkaStreams` application to the output of a `KafkaStreams` application running the same topology against the same input topics.

A quick overview of the components:

1. `OrderAndCustomerDriver` produces the data to the input topics
2. `ResultsComparatorService` compares the results of the output topics (more on methodology later)
3. `OrderProcessingExample` has the topology itself

Quite a few pieces of code in this PR are just moved out from nested classes in the old `E2E` test suite to top level components, so feel free to just review the three classes listed above -- the rest is just boilerplate.

**I've highlighted some places I want more detailed reviews on**

-------

### Workload / Comparison

The data is modeled as customers & orders, where orders are generated at a 8:1 ratio and reference already created customers.

As far as the actual workload, this tests a simple stream-table join. In order to make sure that the results are 100% deterministic, the original stream is keyed on the table key (this avoids any repartition which could lead to results that aren't exactly 1:1 in the output topics). 

The comparison methodology is pretty straightforward as well: we just read from each topic partition and buffer results. When we have results from both output topics, we compare them to make sure they're the same. If the buffer builds up significantly so that one topic is way ahead of the other, we pause consuming from that output topic to avoid overloading memory. 

----

### Running the workload locally

Apply the diff below to make sure that docker compose is setup correctly to run with localhost. Then run `docker compose -f docker-compose-mongo.yml up` to get the infra setup. From that point on you can modify the run configuration in IntelliJ to just run it directly in the IDE just make sure to add `<path>/<to>/responsive-pub/kafka-client-examples/e2e-test/docker/antithesis/config/volumes/app-mongo.properties` as the program argument and `E2E_APP_MODE=<MODE>` as the env variable. Also tick "allow multiple instances" so IntelliJ doesn't kill the old run when you start a new one.

```
diff --git a/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-base.yml b/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-base.yml
index 1732978..cd6fa1c 100644
--- a/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-base.yml
+++ b/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-base.yml
@@ -29,7 +29,7 @@ services:
       KAFKA_BROKER_ID: 1
       KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
       KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_INTERNAL:PLAINTEXT
-      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://broker:9092,PLAINTEXT_INTERNAL://broker:29092
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092,PLAINTEXT_INTERNAL://broker:29092
       KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
       KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
       KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
diff --git a/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-mongo.yml b/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-mongo.yml
index fe52226..f0ded12 100644
--- a/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-mongo.yml
+++ b/kafka-client-examples/e2e-test/docker/antithesis/config/docker-compose-mongo.yml
@@ -9,8 +9,6 @@ services:
     command: mongod --replSet rs0
     ports:
       - "27017:27017"
-    volumes:
-      - ./volumes/mongo/data:/data/db
     healthcheck:
       test: echo "try { rs.status() } catch (err) { rs.initiate({_id:'rs0',members:[{_id:0,host:'mongo:27017',priority:1},{_id:1,host:'mongo2:27017',priority:0.5},{_id:2,host:'mongo3:27017',priority:0.5}]}) }" | mongosh --port 27017 --quiet
       interval: 5s
@@ -27,8 +25,6 @@ services:
     command: mongod --replSet rs0
     ports:
       - "27018:27017"
-    volumes:
-      - ./volumes/mongo2/data:/data/db
     networks:
       backend:
         ipv4_address: 10.0.0.31
@@ -40,8 +36,6 @@ services:
     command: mongod --replSet rs0
     ports:
       - "27019:27017"
-    volumes:
-      - ./volumes/mongo3/data:/data/db
     networks:
       backend:
         ipv4_address: 10.0.0.32
\ No newline at end of file
diff --git a/kafka-client-examples/e2e-test/docker/antithesis/config/volumes/app-mongo.properties b/kafka-client-examples/e2e-test/docker/antithesis/config/volumes/app-mongo.properties
index 8b1a54b..265e48c 100644
--- a/kafka-client-examples/e2e-test/docker/antithesis/config/volumes/app-mongo.properties
+++ b/kafka-client-examples/e2e-test/docker/antithesis/config/volumes/app-mongo.properties
@@ -14,9 +14,9 @@
 # limitations under the License.
 #
 
-bootstrap.servers=broker:9092
+bootstrap.servers=localhost:9092
 responsive.org=responsive
 responsive.env=test
 
 responsive.storage.backend.type=MONGO_DB
-responsive.mongo.endpoint=mongodb://mongo:27017
\ No newline at end of file
+responsive.mongo.endpoint=mongodb://localhost:27017
\ No newline at end of file
```


kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/ResultsComparatorService.java
```diff
@@ -0,0 +1,184 @@
...
+            new TopicPartition(resultsTopic(false), p)
+        )).toList();
+  }
+
+  private record Record(ConsumerRecord<String, EnrichedOrder> record)
+      implements Comparable<Record> {
+
+    @Override
+    public int compareTo(final Record o) {
+      return record.value().compareTo(o.record().value());
```
agavra, line 181: this is the logic that compares two records -- we effectively only call this method to compare two output records. It would be good to review and see if we want to compare anything else (e.g. offset, but I don't think offset is necessarily deterministic)

PR #347, opened 2024-09-11
change mongo scans to filter client-side

agavra:
Turns out this was a bit more of a refactoring nightmare than I wanted it to be! Strategy was to pass in alongside the `TablePartitioner` a method that allows us to determine whether or not a key belongs to a Kafka partition. We can consider refactoring that to return a specific partition instead of whether or not a key "belongs" in a given one, but this makes it simpler to bypass the method in cases where it doesn't need to be implemented.

90% of this PR is just adding generics all over the place to somewhere that were necessary to get the `TablePartitioner` piped into the tables instead of being created in the flush manager

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/TablePartitioner.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/TablePartitioner.java
@@ -16,6 +16,8 @@
 
 package dev.responsive.kafka.internal.db.partitioning;
 
+import org.apache.kafka.common.utils.Bytes;
+
 public interface TablePartitioner<K, P> {
 
   /**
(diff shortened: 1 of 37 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/TablePartitioner.java
```diff
@@ -31,27 +33,15 @@ public interface TablePartitioner<K, P> {
...
-    @Override
-    public Integer tablePartition(final int kafkaPartition, final K key) {
-      return kafkaPartition;
-    }
-
-    @Override
-    public Integer metadataTablePartition(final int kafkaPartition) {
-      return kafkaPartition;
-    }
+  boolean belongs(final Bytes key, final int kafkaPartition);
```
rodesai, line 41: seems cleaner to just have a method that returns the partition for a key, or the number of partitions so then we don't have a bunch of impls that throw unsupportedoperationexception than to have those impls understand the context of why it's being called and throw accordingly.

rodesai (approved):
Mostly LGTM, one structural comment inline.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/partitioning/TablePartitioner.java
agavra: discussed offline, the exceptions are less about the API and more a shortcut in order to avoid needing to pass in the numPartitions to all the other implementations. eventually it would make sense to implement this method in those as well.

PR #350, opened 2024-09-15
Revert "change mongo scans to filter client-side (#347)"

rodesai:
This reverts commit 8e9b0d8ab83cea894d2e6a5d76bb3d57e1c26918 which removed the partition from the mongo value schema and used client-side filtering that computed the partition from the mongo key.

It turns out that this approach won't actually work because the mongo key may not be the original record key that was used to compute the changelog partition. We therefore cannot use the mongo key to compute the partition.:
- some dsl operators include a timestamp in the key
- a user writing their own PAPI processor is free to construct their own key, which we cannot predict.


agavra (commented):
Posthumous approval ✅

PR #351, opened 2024-09-24
fix bugs with default windowed store

agavra:
There were two bugs with the `ResponsiveDefaultDslStoreSuppliers`:

1. for some reason I thought it was appropriate to subtract the window size from retention
2. we didn't properly pass through `retainDuplicates` (it wasn't supported when we first created these classes)
 
This PR fixes both of these issues and changes the tests to leverage the default suppliers instead of passing in custom stores to make sure we have coverage there.


rodesai (approved):
LGTM

PR #352, opened 2024-09-25
initial commit for spring boot compatbility (JDK 17)

agavra:
See `KafkaStreamsApp` for how it's used. I tried many approaches to avoid copying a bunch of code over from the `spring-kafka` package and this was the best I could come up with.

I tried using bytecode manipulation, but turns out it's really difficult to replace a constructor call in Java -- so if I was going to override the entire `start` method in the bean I may as well just do what I did here so we have full control over it going forward.

I tested the basic functionality as well as the integration with other beans (like the DLQ bean) and it seems to work... difficult to tell what will break and what won't though!

### DIFF
--- a/responsive-spring/build.gradle.kts
+++ b/responsive-spring/build.gradle.kts
@@ -1,0 +1,76 @@
+import java.io.ByteArrayOutputStream
+
+/*
+ * Copyright 2023 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("responsive.java-library-conventions")
+    id("java")
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    }
+}
+
+/*********** Generated Resources ***********/
+
+val gitCommitId: String by lazy {
+    val stdout = ByteArrayOutputStream()
+    //rootProject.exec {
+    exec {
+        commandLine("git", "rev-parse", "--verify", "--short", "HEAD")
+        standardOutput = stdout
+    }
+    stdout.toString().trim()
+}
+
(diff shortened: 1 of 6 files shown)


responsive-spring/src/main/java/dev/responsive/spring/config/ResponsiveFactoryBean.java
```diff
@@ -0,0 +1,353 @@
...
+
+  @Override
+  @Nullable
+  public Properties getStreamsConfiguration() {
+    return properties;
+  }
+
+  @Override
+  public void setClientSupplier(final KafkaClientSupplier clientSupplier) {
+    Assert.notNull(clientSupplier, "'clientSupplier' must not be null");
```
ableegoldman, line 112: guess we'll need to remember to update this when the new KafkaClientInterceptor replaces the KafkaClientSupplier -- don't suppose we'll get the deprecation warning to remind us (eg from deprecating the old KafkaStreams constructors which accept the KafkaClientSupplier)?

responsive-spring/src/test/java/dev/responsive/spring/KafkaStreamsApp.java
```diff
@@ -0,0 +1,151 @@
...
+        RecoveringDeserializationExceptionHandler.class);
+    props.put(
+        RecoveringDeserializationExceptionHandler.KSTREAM_DESERIALIZATION_RECOVERER,
+        recoverer()
+    );
+    return new KafkaStreamsConfiguration(props);
+  }
+
+  @Bean
+  public StreamsBuilderFactoryBeanConfigurer moreMagic() {
```
ableegoldman, line 85: 😬

ableegoldman: Was laughing at the method name but also just noticed the return type as well -- if that isn't the most Java nonsense I've ever seen...

ableegoldman (commented):
Reviewed to the best of my ability...frankly I hardly understand what's going on here but I think what's important is the approach, so I will comment on that: my only concern is whether this will be difficult to maintain as things change in the Streams API. See for example the comment about KafkaClientSupplier

responsive-spring/src/test/java/dev/responsive/spring/KafkaStreamsApp.java
agavra: haha I thought it was a little cheeky ha!

responsive-spring/build.gradle.kts
```diff
@@ -0,0 +1,78 @@
+import java.io.ByteArrayOutputStream
```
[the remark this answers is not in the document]
agavra: I did notice this and also that the other build.gradle.kts files match this pattern 🤷 not sure if this is canonical or not

PR #354, opened 2024-09-26
Run regression test in antithesis

ableegoldman:


ableegoldman:
Going to just merge this so I can test it without dealing with GHA bs

PR #356, opened 2024-09-27
Allow configuration of generic Mongo params via connection string

ableegoldman:
Wanted to be able to set a higher connection idle timeout. Might as well take the opportunity to allow configuration of all the Mongo params via the connection string, rather than adding them one by one.

Context: saw this in an Antithesis run of the regression test:

```
org.apache.kafka.streams.errors.StreamsException: com.mongodb.MongoSocketReadException: Prematurely reached end of stream
```

According to some brief googling, this suggests the Mongo connection closed due to being idle. I'm guessing this is an effect of the chaos testing

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/utils/SessionUtil.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/utils/SessionUtil.java
@@ -42,6 +42,7 @@ import dev.responsive.kafka.internal.db.ResponsiveRetryPolicy;
 import java.net.InetSocketAddress;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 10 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/utils/SessionUtil.java
```diff
@@ -121,6 +122,7 @@ public static MongoClient connect(
 
     MongoClientSettings settings = MongoClientSettings.builder()
         .applyConnectionString(new ConnectionString(connectionString))
+        .applyToConnectionPoolSettings(b -> b.maxConnectionIdleTime(60000, TimeUnit.MILLISECONDS))
```
ableegoldman, line 125: How long are the Antithesis test interruptions? I guess we should probably make this a config so we can tune it at will?

rodesai (approved):
LGTM!

PR #357, opened 2024-09-27
Add Antithesis event signals and error injection for regression test

ableegoldman:
Moved common starter signal `"Consumed N output records" from e2e test driver to generic utils class so we can regex match for the common Logger prefix instead of looking for test-specific class names like E2eTestDriver

Also added e2e error injection to the regression test topologies (for the responsive app only)

After this we should change the start trigger from 
```E2ETestDriver - consumed 1 records```
to
```EventSignals - consumed 1 records```


rodesai (approved):
LGTM

PR #358, opened 2024-09-30
Collect producer transaction init/abort metrics

hachikuji:
We are already collecting commit and send-offsets times, so we may as well include transaction initialization and abort.


rodesai (approved):
LGTM!

PR #360, opened 2024-10-01
Use SDK for Antithesis tests

ableegoldman:
Beginning the migration to the new SDK for Antithesis. Mostly focusing on the new regression test in this PR and the `"ANTITHESIS NEVER"` string literals, will complete the migration in a followup PR to catch any remaining assertions from the async test

### DIFF
--- a/kafka-client-examples/e2e-test/build.gradle.kts
+++ b/kafka-client-examples/e2e-test/build.gradle.kts
@@ -10,16 +10,17 @@ application {
 dependencies {
     // todo: how to set the version here?
     implementation(project(":kafka-client"))
-    implementation("com.google.guava:guava:32.1.1-jre")
-    implementation("org.apache.kafka:kafka-clients:3.7.0")
-    implementation("org.apache.kafka:kafka-streams:3.7.0")
-    implementation("org.apache.logging.log4j:log4j-slf4j-impl:2.20.0")
+    implementation(libs.kafka.clients)
+    implementation(libs.kafka.streams)
+
+    implementation("com.antithesis:sdk:1.3.1")
     implementation("org.apache.commons:commons-text:1.10.0")
-    implementation("com.scylladb:java-driver-core:4.15.0.0")
-    implementation("com.scylladb:java-driver-query-builder:4.15.0.0")
-    implementation("com.scylladb:java-driver-mapper-runtime:4.15.0.0")
-    implementation("com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.15.2")
-    implementation("org.mongodb:mongodb-driver-core:4.10.2")
+
+    implementation(libs.guava)
+    implementation(libs.slf4j.log4j2)
+    implementation(libs.bundles.scylla)
+    implementation(libs.jackson)
+    implementation(libs.mongodb.driver.core)
 
     testImplementation(testlibs.bundles.base)
     testImplementation(testlibs.bundles.testcontainers)
--- a/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
+++ b/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
@@ -27,6 +27,7 @@ import com.datastax.oss.driver.api.core.servererrors.WriteFailureException;
 import com.datastax.oss.driver.api.core.servererrors.WriteTimeoutException;
 import com.mongodb.MongoNotPrimaryException;
 import com.mongodb.MongoQueryException;
+import com.mongodb.MongoSocketReadException;
 import com.mongodb.MongoTimeoutException;
 import java.net.ConnectException;
 import java.util.LinkedList;
@@ -77,6 +78,7 @@ public class UncaughtStreamsAntithesisHandler implements StreamsUncaughtExceptio
         InjectedE2ETestException.class,
         InvalidProducerEpochException.class,
         MongoNotPrimaryException.class,
+        MongoSocketReadException.class,
         MongoTimeoutException.class,
         ProducerFencedException.class,
         ReadFailureException.class,
--- a/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
+++ b/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
@@ -16,6 +16,7 @@
 
 package dev.responsive.examples.common;
 
+import com.antithesis.sdk.Assert;
 import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.DriverTimeoutException;
 import com.datastax.oss.driver.api.core.connection.ConnectionInitException;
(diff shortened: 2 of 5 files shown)


kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/ResultsComparatorService.java
```diff
@@ -118,13 +127,13 @@ protected void run() throws Exception {
             baseline.remove();
             matches++;
           } else if (responsive.hasNext() && baseline.hasNext()) {
-            LOG.error(
-                "ANTITHESIS NEVER: Expected to consume records in order, but they were not"
-                    + "in order. Most recent record from responsive is {} and most recent up"
-                    + "record from baseline is {}",
+            Assert.unreachable(String.format(
+                "Expected to see identical output records in identical order, but the next set "
```
ableegoldman, line 131: @agavra is this a fair re-wording of this error message? It's not just about checking ordering, it's about correctness, right?

Also, it seems like it would be possible for the test to "pass" when we actually never see any output from one side at all. I did add the `Assert#sometimes` cases above to help so now it should enforce that we see output from each side at least once, but in theory the Responsive side could stall after the first record and we wouldn't catch this right?

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
```diff
@@ -77,12 +77,10 @@ private boolean shouldLogError(final Throwable throwable, List<Throwable> seen)
         InjectedE2ETestException.class,
         InvalidProducerEpochException.class,
         MongoNotPrimaryException.class,
+        MongoTimeoutException.class,
         ProducerFencedException.class,
         ReadFailureException.class,
         ReadTimeoutException.class,
-        RebalanceInProgressException.class,
-        TaskCorruptedException.class,
-        TaskMigratedException.class,
```
ableegoldman, line 85: Sanity check me here I guess, but none of these should ever get thrown all the way up to the uncaught exception handler in Streams. So if we do ever see something like these it would be a bug that we should look into (whether in Responsive code or AK)

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
```diff
@@ -77,12 +78,11 @@ private boolean shouldLogError(final Throwable throwable, List<Throwable> seen)
         InjectedE2ETestException.class,
         InvalidProducerEpochException.class,
         MongoNotPrimaryException.class,
+        MongoSocketReadException.class,
+        MongoTimeoutException.class,
```
ableegoldman, line 82: I wonder if it might make sense to always inject a StreamsUncaughtExceptionHandler that checks for storage/Responsive-specific exceptions like this and opts to replace the thread if it sees a known exception. Obviously for anything else we delegate to the user's handler (if they have one) -- thoguhts?

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/ResultsComparatorService.java
agavra: correct and correct 👍 I hadn't added the "sometimes" assertion since the SDK wasn't available but that was my intention.

kafka-client-examples/e2e-test/build.gradle.kts
```diff
@@ -10,16 +10,17 @@ application {
 dependencies {
     // todo: how to set the version here?
     implementation(project(":kafka-client"))
-    implementation("com.google.guava:guava:32.1.1-jre")
-    implementation("org.apache.kafka:kafka-clients:3.7.0")
-    implementation("org.apache.kafka:kafka-streams:3.7.0")
-    implementation("org.apache.logging.log4j:log4j-slf4j-impl:2.20.0")
+    implementation(libs.kafka.clients)
```
agavra, line 13: ❤️ thank you for making this use libs!

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
agavra: I think it's better to bubble them up - we should leverage the db client's native retry mechanisms instead of having two layers of retry mechanisms. Those clients have more fine-grained mechanisms (e.g. Cassandra knows what you can retry based on consistency guarantees, I assume Mongo has something similar).

The risk of just replacing the thread is that if there's actually a problem the backoff mechanisms won't be as effective if there's two layers of retries and we may exacerbate the issue.

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
agavra: cc @rodesai - I believe you added these? What was the reason for that?

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/e2etest/E2ETestDriver.java
```diff
@@ -243,12 +246,20 @@ private void pollOnce() {
...
       maybeLogConsumed();
       maybeLogAllConsumed();
       cs.updateReceived(poppedOffsets, ps.partition(), cr.value().digest());
     }
   }
 
+  private void maybeFireSetupCompleteSignal() {
+    if (!setupCompleteSignalFired && recordsProcessed > 0) {
+      LOG.info("Received at least one output record, setup is complete");
+      Lifecycle.setupComplete(null);
```
agavra, line 259: what's `null` here? can we give antithesis more info? same with most other calls to the antithesis SDK, seems like we're passing in lots of `null`

agavra (approved):
LGTM!

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
rodesai: Yeah I tend to agree. I think we should do some level of retries at a level that's close to the actual error (e.g. by configuring the client or from the code that calls the client if the client doesn't support retries), and then just treat them like any other exception if they persist.

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
rodesai: I took them all to be things that you'd expect to see in the presence of faults.

I may have misunderstood `RebalanceInProgressException` - I thought you'd hit that if the group hits a rebalance while the application is processing and it calls something other than `poll`.

From what I see `TaskCorruptedException` can be thrown for various transient errors like the broker not being available or if `process` hits a timeout. Ideally we can differentiate these from the state actually being corrupted, but all the cases I saw were transient errors.

`TaskMigratedException` seems pretty expected right? Streams seems to throw it whenever it detects the producer's been fenced. And it gets thrown when by us when we fail to reserve an epoch because the task got fenced.

rodesai (approved):
LGTM!

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/common/UncaughtStreamsAntithesisHandler.java
rodesai: Ohhhh I see what you're saying - streams should catch these and handle them transparently. That makes sense to me. Lets drop them and see what happens.

ableegoldman: yeah, these are "expected" exceptions but they should never be thrown all the way up to the user's exception handler, and should be handled by Streams

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/e2etest/E2ETestDriver.java
ableegoldman: I'm going to merge this for now to unblock the nightlies but I'll follow up on this

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/ResultsComparatorService.java
ableegoldman: Do you think we should try to assert that a certain number of records minimum are processed? Or that one side doesn't lag for too long? Not quite sure what a good trigger would be but we want to make sure some kind of alarm is raised if one side only ever processes a single record and then gets stuck for the rest of time.

Might not be a good way to enforce this since the Responsive side can lag the Streams side, especially with chaos testing, and I'm not sure we can put a bound on either. Did you have any thoughts about this when you wrote the test?

agavra: if I understand antithesis correctly, it will continuously run the test until it doesn't really notice any new scenarios being hit so it's just important that the setup ran. from that point on code coverage and randomness will handle the rest.

cc @rodesai

PR #361, opened 2024-10-01
Collect MongoDB command execution metrics

hachikuji:
This patch adds collection of MongoDB command metrics. We implement a `CommandListener` as documented [here](https://www.mongodb.com/docs/drivers/java/sync/v4.6/fundamentals/monitoring/) which records counts and cumulative latencies for command successes and failures.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListener.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListener.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2024 Responsive Computing, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.responsive.kafka.internal.db.mongo;
+
+import com.mongodb.event.CommandFailedEvent;
+import com.mongodb.event.CommandListener;
+import com.mongodb.event.CommandSucceededEvent;
+import dev.responsive.kafka.internal.metrics.ResponsiveMetrics;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.common.metrics.stats.CumulativeCount;
+import org.apache.kafka.common.metrics.stats.CumulativeSum;
+
+/**
+ * Listener which reports MongoDB command failures and successes. For each result,
+ * we record a total count representing the number of commands executed which had that
+ * result (success or failure), and a cumulative latency. Reported metrics are tagged by
+ * the MongoDB command name.
+ */
+public class MongoTelemetryListener implements CommandListener {
+  private static final String MONGODB_METRICS_GROUP = "responsive-mongodb";
+
+  static final MetricName COMMANDS_SUCCEEDED_LATENCY = new MetricName(
+      "commands-succeeded-cumulative-latency",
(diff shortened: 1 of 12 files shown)


kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListenerTest.java
```diff
@@ -0,0 +1,164 @@
...
+    assertEquals(Collections.singletonMap("command", "find"), findFailureCountName.tags());
+
+    final MetricName findFailureLatencyName = commandsFailedLatency("find");
+    assertEquals("responsive-mongodb", findFailureLatencyName.group());
+    assertEquals("commands-failed-cumulative-latency", findFailureLatencyName.name());
+    assertEquals(Collections.singletonMap("command", "find"), findFailureLatencyName.tags());
+  }
+
+  @Test
+  public void testSuccessfulCommand() {
```
agavra, line 65: (nit) we don't have to change it for this PR, but stylistically for future tests prefer

1. naming something `shouldXYZ` instead of `testXYZ` since it encourages not only what functionality its testing but also describes the success criteria. we noticed this results in more targeted tests that are easier to debug
2. add comments for Given, When, Then -- this structures tests in a way that's easy to read and consistent throughout the codebase (setup, the test, the assertions)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListener.java
```diff
@@ -0,0 +1,139 @@
...
+    String sensorName = "mongodb-commands" + "-" + command
+        + (isSuccess ? "-succeeded" : "-failed");
+
+    Sensor sensor = metrics.getSensor(sensorName);
+    if (sensor == null) {
+      sensor = metrics.addSensor(sensorName);
+
+      if (isSuccess) {
+        sensor.add(commandsSucceededCount(command), new CumulativeCount());
+        sensor.add(commandsSucceededLatency(command), new CumulativeSum());
```
agavra, line 97: I really need to brush up on my stats, but can we compute any histogram using htis or just average? I feel like p90/95/99 is helpful as well

agavra (approved):
LGTM! Some comments inline

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListener.java
```diff
@@ -0,0 +1,138 @@
...
+  static MetricName commandsFailedCount(String command) {
+    return metricName(COMMANDS_FAILED_COUNT, buildTags(command));
+  }
+
+  static MetricName commandsFailedLatency(String command) {
+    return metricName(COMMANDS_FAILED_LATENCY, buildTags(command));
+  }
+
+  private static Map<String, String> buildTags(String command) {
+    return Collections.singletonMap("command", command);
```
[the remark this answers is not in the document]
hachikuji: Yeah, I was thinking about that. At first I had a single metric for all commands, but it felt like it would be too noisy to be useful. I imagine our usage of the commands is pretty limited. I was planning on starting up an application and looking at JMX to get a feel for it. I'll do that before we merge here.

kafka-client/src/test/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListenerTest.java
hachikuji: Sounds fair to me.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoTelemetryListener.java
hachikuji: There is a `Percentiles` class included with Kafka metrics, but it is almost never used. Most of our histograms in Kafka tend to use the old yammer metrics implementation. Max is a straightforward (and cheap) addition though if you think we could get away with just that.

agavra: I don't imagine `max` is super helpful for most of these - we can punt for now and see how far we get with average, that's what matters for throughput anyway.

hachikuji:
@agavra Here is a snapshot of jconsole with these metrics enabled. This is using our example app. Only four commands are used. I am not sure whether a more general application would use others.
Screenshot 2024-10-02 at 12 02 18 PM

agavra:
Perfect, let's not worry about it then and just ship it as is. If it gets out of hand we can always whitelist later. 🫡

PR #362, opened 2024-10-02
add kv store implementation for rs3

rodesai:
This patch adds a key-value store implementation for RS3: Adds the rs3 repo as a submodule so that we can get to the rs3 proto file

Extends the flush writing path to pass the relevant changelog offset to the flush manager. RS3 uses this to specify the end offset of the WAL segments that it writes during flush.

Adds the RS3 RemoteKVTable implementation. The RS3 table uses the RS3 grpc protocol to do writes/reads to/from RS3. The grpc protocol is behind a more generic RS3Client type that exposes apis for writing wal segments, reading keys, and getting the current written/flushed offsets.

When the table is initialized, it fetches the current written offset for all PSSs, and uses the lowest of these as the restore point.

On a write, RS3KVFlushManager builds up a WAL segment for each PSS for a given LSS/partition, adding entires to the segment based on the record key. This mapping (from key to PSS) is provided by PssPartitioner. When it has all the writes in the current flush, it completes the WAL segment write rpcs to RS3.

On a read, RS3KVTable fetches the value from the PSS that corresponds to the requested key. Range scans are currently not supported.

Adds a new store type called RS3 for specifying an RS3 store.

Adds some simple config for talking to the rs3 server that allows specifying the hostname/port. The store->PSS mapping is still hard-coded. I'll make it configurable in a follow-up.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVTable.java
```diff
@@ -0,0 +1,139 @@
...
+    this.name = Objects.requireNonNull(name);
+    this.storeId = Objects.requireNonNull(storeId);
+    this.rs3Client = Objects.requireNonNull(rs3Client);
+    this.pssPartitioner = Objects.requireNonNull(pssPartitioner);
+  }
+
+  @Override
+  public KVFlushManager init(final int kafkaPartition) {
+    if (flushManager != null) {
+      throw new IllegalStateException("already initialized");
```
agavra, line 45:
```suggestion
      throw new IllegalStateException("already initialized for partition " + kafkaPartition);
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVTable.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    this.lssId = new LssId(kafkaPartition);
+
+    // TODO: we should write an empty segment periodically to any PSS that we haven't
+    //       written to to bump the written offset
+    final HashMap<Integer, Optional<Long>> lastWrittenOffset = new HashMap<>();
```
agavra, line 52: might be nice to have a strongly typed `Pss` class just for readability throughout the code even if it just wraps an integer (though it could encapsulate some of the bytestring prefix stuff too)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVTable.java
```diff
@@ -0,0 +1,139 @@
...
+        .map(v -> v.orElse(-1L))
+        .min(Long::compare)
+        .orElse(-1L);
+    if (fetchOffsetOrMinusOne == -1) {
+      this.fetchOffset = ResponsiveStoreRegistration.NO_COMMITTED_OFFSET;
+    } else {
+      this.fetchOffset = fetchOffsetOrMinusOne;
+    }
+
+    LOG.info("restore rs3 kv table from offset {} for {}", fetchOffset, kafkaPartition);
```
agavra, line 67: would it help to log all the pss written offsets? at a minimum I imagine a separate log that logs `kafka partition -> [pss]` would be helpful

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/CurrentOffsets.java
```diff
@@ -0,0 +1,21 @@
+package dev.responsive.kafka.internal.db.rs3.client;
+
+import java.util.Optional;
+
+public class CurrentOffsets {
```
agavra, line 5: can we document the difference between these?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3Client.java
```diff
@@ -0,0 +1,234 @@
...
+              .setExpectedWrittenOffset(expectedWrittenOffset.orElse(WAL_OFFSET_NONE));
+          if (entry instanceof Put) {
+            final var put = (Put) entry;
+            final var putBuilder = Rs3.WriteWALSegmentRequest.Put.newBuilder()
+                .setKey(ByteString.copyFrom(put.key()));
+            if (put.value().isPresent()) {
+              putBuilder.setValue(ByteString.copyFrom(put.value().get()));
+            }
+            entryBuilder.setPut(putBuilder.build());
+          }
```
agavra, line 115: might make sense to implement a function `WalEntry#addTo(WriteWALSegmentRequestBuilder builder)` and implement it on each implementation of `WalEntry` instead of doing the check here

agavra (commented):
This makes sense to me. Just to make sure I understand it there's basically just two things in this PR:

1. a simple client interface that writes KV pairs (potentially in an async fashion) to RS3
2. the PSS range-mapping logic

The first is just a bunch of boilerplate, nothing concerning to me (some minor comments inline). The latter is a bit more complicated, I think it might be good to separate that into a different PR and just use a simple static partitioner for v0.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/PssRangePartitioner.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.apache.kafka.streams.state.internals.Murmur3;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PssRangePartitioner implements PssPartitioner {
```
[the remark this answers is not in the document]
ableegoldman: FWIW I agree with Almog on this, would be nice to split out into a followup PR

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -0,0 +1,217 @@
...
+        consumedOffset,
+        writtenOffsets.get(pssId),
+        kafkaPartition
+    );
+    writers.put(pssId, writer);
+    return writer;
+  }
+
+  @Override
+  public RemoteWriteResult<Integer> postFlush(long consumedOffset) {
```
ableegoldman, line 107: nit: 
```suggestion
  public RemoteWriteResult<Integer> postFlush(final long consumedOffset) {
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -0,0 +1,217 @@
...
+    writers.clear();
+    return super.postFlush(consumedOffset);
+  }
+
+  Optional<Long> writtenOffset(final int pssId) {
+    return writtenOffsets.get(pssId);
+  }
+
+  @Override
+  public RemoteWriteResult<Integer> updateOffset(long consumedOffset) {
```
ableegoldman, line 120:
```suggestion
  public RemoteWriteResult<Integer> updateOffset(final long consumedOffset) {
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -0,0 +1,217 @@
...
+    // no op for rs3 store. offsets are updated by RS3Client.writeWalSegment
+    return RemoteWriteResult.success(null);
+  }
+
+  @Override
+  public String failedFlushInfo(long batchOffset, Integer failedTablePartition) {
```
ableegoldman, line 126:
```suggestion
  public String failedFlushInfo(final long batchOffset, final Integer failedTablePartition) {
```

ableegoldman: etc throughout the PR. 

We should probably just add a checkstyle rule for this

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVTable.java
ableegoldman: Let's include the table name here too (can we also log an error before throwing)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVTable.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Override
+  public KeyValueIterator<Bytes, byte[]> all(final int kafkaPartition, final long minValidTs) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public long approximateNumEntries(final int kafkaPartition) {
+    LOG.warn("approximateNumEntries not implemented for RS3");
+    return 0;
```
ableegoldman, line 106: since we're not throwing an exception let's return -1 so it's obvious the store isn't empty, the API just isn't implemented

ableegoldman:
FYI this needs to be rebased

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = controller-api/src/main/external-protos/opentelemetry-proto
 	url = https://github.com/open-telemetry/opentelemetry-proto.git
         branch = main
+[submodule "kafka-client/src/main/external-protos/rs3"]
+	path = kafka-client/src/main/external-protos/rs3
+	url = https://github.com/responsivedev/rs3.git
```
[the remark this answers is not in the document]
rodesai: yeah this just occurred to me as well. They would not be able to. We can move the protocol spec to a public repo if we want to support that.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/PssRangePartitioner.java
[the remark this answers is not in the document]
rodesai: sounds good - I'll split that out and we can discuss the mapping strategy in another forum.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/PssPartitioner.java
```diff
@@ -0,0 +1,9 @@
+package dev.responsive.kafka.internal.db.rs3;
+
+import java.util.List;
+
+public interface PssPartitioner {
```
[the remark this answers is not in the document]
rodesai: Yeah that's fair. I can write something up to go along with a follow-up PR. Since we're all here, a quick breakdown:
LS - stands for logical store. This is 1:1 with a Kafka Streams store at the processor level - meaning it encompasses the entire keyspace (not just one task). We could probably do away with this term since we don't actually use it anywhere in the implementation.
LSS - stands for logical store shard. This is 1:1 with a Kafka Streams store at the task level, and therefore 1:1 with a Kafka partition. The expectation from RS3 is that each LSS has it's own WAL, which is the changelog topic-partition for the store. 
PS - Stands for physical store. This is a single store in RS3 that spans the whole RS3 cluster and stores data from one LS.
PSS - Stands for physical store shard. This is a single partition/shard of the PS that runs on one RS3 node as a slatedb instance. This is 1:1 with a TablePartition.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  @Override
+  public String tableName() {
+    return table.name();
+  }
+
+  @Override
+  public TablePartitioner<Bytes, Integer> partitioner() {
+    return new TablePartitioner<Bytes, Integer>() {
```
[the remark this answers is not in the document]
rodesai: yep, I'll fix.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -0,0 +1,217 @@
...
+    if (maybeWrittenOffset.isPresent()) {
+      final var writtenOffset = maybeWrittenOffset.get();
+      if (writtenOffset >= consumedOffset) {
+        // This can happen after a restart, during restore because we restore from
+        // the earliest written offset over all pss. For most pss we are restoring
+        // from an offset earlier than the pss's last written offset. In that case,
+        // we wait until we have crossed the written offset for the pss before sending
+        // actual records.
+        // TODO: we should be stricter when checking this case
+        return new NoopWriter(kafkaPartition);
```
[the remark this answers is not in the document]
rodesai: No, the `writers` map is strictly internal to this class and is just used to record the written offsets after flush completes. By excluding it here we just don't include the writer in the set we use to update the written offsets (because we aren't actually writing anything for this pss for this batch)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3Client.java
rodesai: I'd rather not put in the `WalEntry` type, as it's not aware of the specific format used by the underlying protocol. I added a helper method to this class.

.gitmodules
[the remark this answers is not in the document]
agavra: let's not worry about it now, especially with the potential changes to product packaging coming up

agavra (approved):
🚀 let's go!

PR #363, opened 2024-10-04
Collect useful Cassandra driver metrics

hachikuji:
This patch adds a custom Cassandra `MetricsFactory` so that the native metrics shipped by the driver can be adapted for reporting through the Kafka Metrics API. See [here](https://docs.datastax.com/en/developer/java-driver/4.15/manual/core/metrics/index.html) for the driver's documentation on metrics. 

We collect the following session metrics:
- bytes-sent
- bytes-received
- cql-requests-count
- cql-requests-cumulative-latency
- cql-request-timeouts-count
- throttling-cumulative-delay
- throttling-errors-count

Additionally we collect the metrics which are aggregated from node-level metrics:
- cql-request-ignores-count
- cql-request-retries-count
- cql-request-errors-count
- connection-errors-count


hachikuji:
Here's what this looks like through JConsole.
Screenshot 2024-10-04 at 12 40 15 PM

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/CassandraMetricsFactory.java
```diff
@@ -0,0 +1,251 @@
...
+      getOrCreateSensor(sessionMetric)
+          .record(TimeUnit.MILLISECONDS.convert(duration, timeUnit));
+    }
+
+    @Override
+    public boolean isEnabled(
+        final SessionMetric metric,
+        @Nullable final String profileName
+    ) {
+      switch (asDefaultSessionMetric(metric)) {
```
agavra, line 140: do you know if there are any general error/success metrics or latency metrics?

agavra (approved):
LGTM! Would love to know if there are latency/error metrics, as those have been the most useful server-side metrics that I look at when we host the Cassandra cluster. Otherwise the infra looks good.

kafka-client/src/main/java/dev/responsive/kafka/internal/metrics/CassandraMetricsFactory.java
hachikuji: I could not find any general success/error metrics. There are some more error metrics defined under `DefaultNodeMetric`. I was a little concerned about exposing node level metrics, but the `MetricsFactory` would allow us to return a session-level metric as a node metric. In other words, we could just ignore the node. That would give us these metrics:
```java
  ABORTED_REQUESTS("errors.request.aborted"),
  WRITE_TIMEOUTS("errors.request.write-timeouts"),
  READ_TIMEOUTS("errors.request.read-timeouts"),
  UNAVAILABLES("errors.request.unavailables"),
  OTHER_ERRORS("errors.request.others"),
  CONNECTION_INIT_ERRORS("errors.connection.init"),
  AUTHENTICATION_ERRORS("errors.connection.auth");
```
What do you think?

agavra: Ugh that's pretty verbose... let's just keep it as you have in your PR for now and if we end up having any issues we can work them in later. Of those I think timeouts are the most interesting.

agavra: Also not sure how I missed `cql-requests-cumulative-latency` but looks like that's already there.

hachikuji: It would be to roll up all of the error metrics into a single `errors.total` metric. We could also expose these:
```
        # The total number of errors on this node that caused the RetryPolicy to trigger a retry
        # (exposed as a Counter).
        #
        # This is a sum of all the other retries.* metrics.
        // retries.total,
        # The total number of errors on this node that were ignored by the RetryPolicy (exposed as
        # a Counter).
        #
        # This is a sum of all the other ignores.* metrics.
        // ignores.total,
```
That would give us a pretty good overall picture about error and retry behavior.

agavra: 💯 that would be perfect

hachikuji: Sounds good. I'll update the patch.

hachikuji:
@agavra Thanks for the review. I pushed an update which adds the following metrics which are aggregated from the node-level metrics:
- cql-request-ignores-count
- cql-request-retries-count
- cql-request-errors-count
- connection-errors-count

agavra (approved):
Nice! Thanks for the super comprehensive tests as well. LGTM

PR #368, opened 2024-10-14
clean up the commitbuffer and store registration if store init fails

rodesai:
This patch cleans up the commit buffer and store registration if a store's operations provider fails to initialize. If we don't clean this up, we leave around stuff like metrics and stale registrations, which can cause downstream errors like failure to register store metrics.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
@@ -123,48 +123,62 @@ public class PartitionedOperations implements KeyValueOperations {
         flushManager
     );
 
-    final CommitBuffer<Bytes, ?> buffer = CommitBuffer.from(
-        batchFlusher,
-        sessionClients,
-        changelog,
-        keySpec,
-        params.name(),
-        false,
-        config
-    );
+    // these objects need to be cleaned up once they're created. If this method fails
+    // then we should make sure we do the cleanup here.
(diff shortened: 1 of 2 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -123,48 +123,62 @@ public static PartitionedOperations create(
...
-    final boolean migrationMode = ConfigUtils.responsiveMode(config) == ResponsiveMode.MIGRATE;
-    long startingTimestamp = -1;
-    final Optional<Duration> ttl = params.timeToLive();
-    if (migrationMode && ttl.isPresent()) {
-      startingTimestamp = Instant.now().minus(ttl.get()).toEpochMilli();
+    // these objects need to be cleaned up once they're created. If this method fails
+    // then we should make sure we do the cleanup here.
+    CommitBuffer<Bytes, ?> buffer = null;
+    ResponsiveStoreRegistration registration = null;
+    try {
```
agavra, line 130: thoughts on making them implement `Closeable` and then using try-with-resources? that way we'll also get IDEA warnings if we forget to do it later

agavra (approved):
LGTM, one code quality suggestion

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
agavra: ah yes, that makes sense to me!

PR #369, opened 2024-10-16
Row-level ttl for KV stores

ableegoldman:


kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -116,10 +116,12 @@ public static KeyValueBytesStoreSupplier factStore(final String name) {
    *         that uses Responsive's storage for its backend
    */
   public static <K, V> StoreBuilder<KeyValueStore<K, V>> keyValueStoreBuilder(
-      final KeyValueBytesStoreSupplier storeSupplier,
+      final ResponsiveKeyValueBytesStoreSupplier storeSupplier,
       final Serde<K> keySerde,
       final Serde<V> valueSerde
   ) {
+    storeSupplier.params().setKeyAndValueSerdes(keySerde, valueSerde);
```
ableegoldman, line 123: @agavra it turns out we only have users pass in the serdes via the StoreBuilder, so we can only automatically grab serdes for them when they're using the PAPI

If we want to support row-level ttl for the DSL then we'll need to add these params either to the ResponsiveStores#Materialized or to the ResponsiveKeyValueParams itself. Would prefer not requiring it for the params personally but thoughts?

kafka-client/src/test/java/dev/responsive/kafka/integration/RowLevelTtlIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+    public Processor<String, String, String, String> get() {
+      return new TtlProcessor();
+    }
+
+    @Override
+    public Set<StoreBuilder<?>> stores() {
+      return Collections.singleton(ResponsiveStores.keyValueStoreBuilder(
+          ResponsiveStores.keyValueStore(
+              ResponsiveKeyValueParams.fact(STORE_NAME).withTtlProvider(
+                  TtlProvider.withDefaultTtl(DEFAULT_TTL).withKeyBasedTtl(k -> ttlForKey(k))
```
agavra, line 161: nit: we can probably just have `withDefault` and `withKeyBased` to avoid some verbosity

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
agavra: 😢 there was no other way to automatically get them? that's a real bummer - I feel like that's a worthwhile improvement to contribute to KS.

I think it's fine not requiring them, the API is already available in `ResponsiveKeyValueParams`, right? They could call `setKeyAndValueSerdes` themselves? We should just make sure to throw with a good exception message if they configure row level TTL and haven't passed in serdes.

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
```diff
@@ -46,7 +49,12 @@ public static ResponsiveKeyValueParams fact(final String name) {
   }
 
   public ResponsiveKeyValueParams withTimeToLive(final Duration timeToLive) {
-    this.timeToLive = timeToLive;
+    this.ttlProvider = TtlProvider.withDefaultTtl(timeToLive);
+    return this;
```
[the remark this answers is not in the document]
ableegoldman: Ah actually I was planning to remove this, I just didn't get around to cleaning it up everywhere yet. Sounds like this would be your preference as well?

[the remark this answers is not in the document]
agavra: that was, but now I'm doubting myself. I think most situations that want TTL will be OK with just a default ttl 🤷 it's not like it adds that much surface area to the API since underneath the hood it uses the same mechanism.

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
ableegoldman: SG

And yeah, I agree we should just expose the serdes in Kafka Streams itself -- this should at least be a relatively short and easy KIP, whereas I can think of some hacky ways to expose the original deserialized key & value but it would probably be a longer & more difficult KIP discussion for that

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    if (ttlForKey != null || ttlForValue != null) {
+      throw new IllegalArgumentException("Must choose only one of key, value, or key-and-value ttl");
+    }
+    return new TtlProvider<>(TtlType.KEY_AND_VALUE, defaultTtl, null, null, ttlForKeyAndValue);
+  }
+
+  public static class TtlDuration {
+
+    /* Equivalent to no ttl -- records are never expired */
+    public static final TtlDuration INFINITE_TTL = new TtlDuration(Duration.ZERO);
```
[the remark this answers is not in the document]
ableegoldman: Can you have an enum with a variable value for the non-infinite ttls? I'm probably lacking in java expertise and/or imagination so pretend I'm 5 and spell this suggestion out for me please 😛

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -100,16 +104,8 @@ public static CassandraFactTable create(
...
-            .insertInto(name)
-            .value(ROW_TYPE.column(), RowType.DATA_ROW.literal())
-            .value(DATA_KEY.column(), bindMarker(DATA_KEY.bind()))
-            .value(TIMESTAMP.column(), bindMarker(TIMESTAMP.bind()))
-            .value(DATA_VALUE.column(), bindMarker(DATA_VALUE.bind()))
-            .build(),
-        QueryOp.WRITE
-    );
+    final PreparedStatementBuilder insert = (k, v) ->
+        buildInsertQuery(client, name, spec.ttlResolver(), k, v);
```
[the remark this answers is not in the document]
ableegoldman: yep, on my TODO list!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -298,4 +296,39 @@ private static String metadataTable(final String tableName) {
...
+    final TtlDuration resolvedTtl = ttlResolver.resolveTtl(key, value);
+    final var baseQuery = com.datastax.oss.driver.api.querybuilder.QueryBuilder
+        .insertInto(name)
+        .value(ROW_TYPE.column(), RowType.DATA_ROW.literal())
+        .value(DATA_KEY.column(), bindMarker(DATA_KEY.bind()))
+        .value(TIMESTAMP.column(), bindMarker(TIMESTAMP.bind()))
+        .value(DATA_VALUE.column(), bindMarker(DATA_VALUE.bind()));
+
+    if (resolvedTtl.equals(TtlDuration.INFINITE_TTL)) {
+      return client.prepare(
```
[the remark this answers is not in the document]
ableegoldman: ah, good to know. will do

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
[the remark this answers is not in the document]
ableegoldman: I mean you can do something like this, but then it's possible for users to mess it up by just returning the `FINITE` enum without setting the actual ttl value. Is there a better way to do? If not, do you still prefer this enum approach?

```
public enum Ttl {
    INFINITE,
    FINITE;

    private Duration ttlDuration;
    
    public void withDuration(final Duration ttlDuration) {
      this.ttlDuration = ttlDuration;
    }

    public Duration ttl() {
      if (isInfiniteRetention()) {
        throw new IllegalStateException("Attempted to retrieve ttl value for infinite retention");
      } else if (ttlDuration == null) {
        throw new IllegalArgumentException("Must specify the ttl duration for FINITE ttl");
      } 
      
      return ttlDuration;
    }
    
    public boolean isInfiniteRetention() {
      return this.ordinal() == INFINITE.ordinal();
    }
  }
```

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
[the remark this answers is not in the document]
ableegoldman: fine with me 🙂

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
ableegoldman: If we are going to have to require the key and value serdes for key/value based TtlProviders, then rather than adding them as optional (and easily missable) params in the ResponsiveKeyValueParams, we might as well just add these to the `.withKey/ValueBased` builder methods for TtlProvider. I'll make this change but lmk if you prefer otherwise or have a different idea

kafka-client/src/test/java/dev/responsive/kafka/integration/RowLevelTtlIntegrationTest.java
ableegoldman: SGTM. Though while we're on the subject, I don't really love the `withKeyBased`/`.withValueBased`/etc method names. Especially if we strip away the "ttl" suffix, it just sounds awkward. Any suggestions?

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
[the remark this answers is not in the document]
agavra: > Can you have an enum with a variable value for the non-infinite ttls?

sadly no 😢 you can in Rust though! But we're in javaland so...

My idea was for something like this:

```java
public class Ttl {
  
  enum Type { INFINITE, FINITE };
  
  private Type type;
  private Duration duration; // null if INFINITE

  public static Ttl of(final Duration duration) {
     return Ttl(duration, FINITE);
  }

  public static Ttl noTtl() {
    return Ttl(null, INFINITE);
  }

  private Ttl(Duration duration, Type type) { ... }

}
```

kafka-client/src/test/java/dev/responsive/kafka/integration/RowLevelTtlIntegrationTest.java
agavra: maybe `fromKey`? so `TtlProvider.withDefault(DEFAULT_TTL).fromKey(k -> k.ttl)`

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
agavra: > whereas I can think of some hacky ways to expose the original deserialized key & value

ah this is what I was hoping the KIP would do... (not expose the Serdes directly)

> just add these to the .withKey/ValueBased builder methods for TtlProvider.

yeah let's just do that

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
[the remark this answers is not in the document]
ableegoldman: ohhh I see. I was hoping there was a way to do it via just the enum, didn't realize you meant an enum within the class. Thanks for clarifying, I'll adopt this!

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
ableegoldman: > ah this is what I was hoping the KIP would do... (not expose the Serdes directly)

ah I shouldn't discourage you from trying this, I guess it might be possible to add the deserialized key/value to the StateStoreContext? I'm not entirely sure this is possible but I didn't put much thought into this so I wouldn't say I'm sure it's impossible either

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
```diff
@@ -116,56 +150,78 @@ private enum TtlType {
...
+  public Serde<V> valueSerde() {
+    return valueSerde;
   }
 
   public TtlDuration defaultTtl() {
     return defaultTtl;
   }
 
-  public TtlDuration computeTtl(
+  public boolean hasTtl() {
```
agavra, line 185: nit: this method name is a little confusing, you could have "no ttl" with a key-provider if the key provider returns INFINITE, right? perhaps we should invert usage and name it `hasNoTtl` (which is guaranteed if the condition is flipped)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -160,10 +158,10 @@ public static PartitionedOperations create(
...
 
       final boolean migrationMode = ConfigUtils.responsiveMode(config) == ResponsiveMode.MIGRATE;
-      long startingTimestamp = -1;
-      final Optional<Duration> ttl = params.timeToLive();
-      if (migrationMode && ttl.isPresent()) {
-        startingTimestamp = Instant.now().minus(ttl.get()).toEpochMilli();
+      long startTimeMs = -1;
+      if (migrationMode && params.ttlProvider().hasTtl()) {
+        // TODO(sophie): figure out how to account for row-level ttl in migration mode
+        startTimeMs = System.currentTimeMillis() - params.ttlProvider().defaultTtl().toMillis();
```
agavra, line 164: let's create a follow-up ticket, but I think what we can do is just have them pass in the "start time" to the tool, and then just apply individual TTls as necessary

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StateDeserializer.java
```diff
@@ -16,28 +16,45 @@
...
+    if (isTimestamped && valueSerde != null) {
+      timestampedValueDeserializer =
+          Optional.of(new ValueAndTimestampSerde<>(valueSerde).deserializer());
+    } else {
+      timestampedValueDeserializer = Optional.empty();
+    }
   }
 
   public K keyFrom(final byte[] keyBytes) {
     return keyDeserializer.deserialize(changelogTopic, keyBytes);
```
agavra, line 50: should we throw a custom exception if `keyDeserializer` is null instead of an NPE? (same for below)

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/Utils.java
```diff
@@ -122,4 +136,53 @@ public static String extractThreadIdFromThreadName(final String threadName) {
     return threadName;
   }
 
+  public static <K, V> byte[] getValueWithTtl(
```
agavra, line 139: nice 👍

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/Utils.java
```diff
@@ -122,4 +136,53 @@ public static String extractThreadIdFromThreadName(final String threadName) {
...
+      }
+    }
+    return result.value();
+  }
+
+  public static <D> byte[] serialize(final D data, final Serde<D> serde) {
+    return serde.serializer().serialize("ignored", data);
+  }
+
+  public static Bytes serializedKey(final String key) {
```
agavra, line 175: what are these used for?

kafka-client/src/test/java/dev/responsive/kafka/internal/db/AbstractCassandraKVTableIntegrationTest.java
```diff
@@ -0,0 +1,23 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package dev.responsive.kafka.internal.db;
+
+public class AbstractCassandraKVTableIntegrationTest {
```
agavra, line 19: used?

responsive-test-utils/src/main/java/dev/responsive/kafka/internal/db/TTDKeyValueTable.java
```diff
@@ -34,26 +33,22 @@ public class TTDKeyValueTable extends TTDTable<Bytes> implements RemoteKVTable<B
...
 
-    while (maybeTtlSpec instanceof DelegatingTableSpec) {
-      if (maybeTtlSpec instanceof TtlTableSpec) {
-        ttl = ((TtlTableSpec) maybeTtlSpec).ttl();
-        break;
-      }
-
-      maybeTtlSpec = ((DelegatingTableSpec) maybeTtlSpec).delegate();
+    if (!spec.ttlResolver().hasConstantTtl()) {
+      throw new UnsupportedOperationException("The ResponsiveTopologyTestDriver does not yet "
```
agavra, line 50: let's create a ticket for this, I think we have customers that use this

agavra (approved):
It's looking good! Thanks for putting this together so fast. No major comments.

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
ableegoldman: Actually I don't know what I was talking about, it should be pretty straightforward. Just add APIs for deserialized key/value to the StateStoreContext and then we can set these in the store layer right before we serialize them (which IIRC is the MeteredXYZStore layer)

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
ableegoldman: Agreed, I actually removed this method entirely in my last update

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/Utils.java
ableegoldman: tests...guess these should be in the test utils, I'll move them

kafka-client/src/test/java/dev/responsive/kafka/internal/db/AbstractCassandraKVTableIntegrationTest.java
ableegoldman: nope this is TODO -(going to combine some of the common KV vs Fact table tests, especially the ttl unit tests)

agavra: if possible, I'd prefer we used static helper methods to abstract classes for tests 🤷

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -253,44 +304,98 @@ public BoundStatement insert(
...
+        .setLong(TIMESTAMP.bind(), epochMillis);
   }
 
   @Override
-  public byte[] get(final int kafkaPartition, final Bytes key, long minValidTs) {
-    final BoundStatement get = this.get
-        .bind()
-        .setByteBuffer(DATA_KEY.bind(), ByteBuffer.wrap(key.get()))
-        .setInstant(TIMESTAMP.bind(), Instant.ofEpochMilli(minValidTs));
+  public byte[] get(final int kafkaPartition, final Bytes key, long streamTimeMs) {
```
agavra, line 328: the conditional complexity of this method is pretty high -- would be good to split it up into individual methods, perhaps follow the form:

```
if !ttlResolver { simpleGet() }
else if !ttlResolver.needsValueToComputeTtl() { getWithMinTtl() }
else { getWithPostFilterForTtl() }
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -338,69 +391,116 @@ public CassandraKVFlushManager init(
...
           .setInt(PARTITION_KEY.bind(), partition)
           .setByteBuffer(FROM_BIND, ByteBuffer.wrap(from.get()))
           .setByteBuffer(TO_BIND, ByteBuffer.wrap(to.get()))
-          .setInstant(TIMESTAMP.bind(), Instant.ofEpochMilli(minValidTs));
+          .setLong(TIMESTAMP.bind(), minValidTs);
 
       final ResultSet result = client.execute(range);
       resultsPerPartition.add(Iterators.kv(result.iterator(), CassandraKeyValueTable::rows));
     }
+    // TODO(sophie): filter by minValidTs if based on value
```
agavra, line 482: we should make sure to address these TODOs before cutting a release since this will result in wrong data, I know it's already a big PR so just make sure we're tracking it somewhere

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -109,13 +116,20 @@ public MongoKVTable(
         Indexes.descending(KVDoc.TOMBSTONE_TS),
         new IndexOptions().expireAfter(12L, TimeUnit.HOURS)
     );
-    if (ttl != null) {
-      final Duration expireAfter = ttl.plus(Duration.ofHours(12));
-      final long expireAfterSeconds = expireAfter.getSeconds();
+    if (ttlResolver.isPresent()) {
+      // if the default ttl is infinite  we still have to define the ttl index for
+      // the table since the ttlProvider may apply row-level overrides. To approximate
+      // an "infinite" default retention, we just set the default ttl to the maximum value
```
agavra, line 122: I think we should change the approach here to be more like the Tombstones: we would create the index on a new field `expire_ts` (instead of the insertion ts), which could be just be null/empty for documents that don't have a TTL. And instead of having `expireAfter = ttl + 12h` we should have `expireAfter = 12h` (which by the way, it looks like this change dropped the 12h, which was there to give a grace to make sure that stream time advances past the physical time).  This would dramatically reduce the size of the index in situations where many rows do not have a TTL.

From https://www.mongodb.com/docs/manual/core/index-ttl/#expiration-of-data: 

> If the indexed field in a document doesn't contain one or more date values, the document will not expire.
> 
> If a document does not contain the indexed field, the document will not expire.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -150,21 +164,63 @@ public MongoKVFlushManager init(final int kafkaPartition) {
...
   @Override
-  public byte[] get(final int kafkaPartition, final Bytes key, final long minValidTs) {
-    final KVDoc v = docs.find(Filters.and(
-        Filters.eq(KVDoc.ID, keyCodec.encode(key)),
-        Filters.gte(KVDoc.TIMESTAMP, minValidTs)
-    )).first();
-    return v == null ? null : v.getValue();
+  public byte[] get(final int kafkaPartition, final Bytes key, final long streamTimeMs) {
+
+    // Need to post-filter if value is needed to compute ttl
```
agavra, line 169: I actually think with the approach suggested above, we can do it all server side because `expire_ts` is already a dedicated column so we don't need to recompute it. We could actually consider doing the same thing in Scylla by inserting an extra column for expire_ts (they just wouldn't create an index on it).

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -150,21 +164,63 @@ public MongoKVFlushManager init(final int kafkaPartition) {
...
 
   @Override
   public KeyValueIterator<Bytes, byte[]> range(
       final int kafkaPartition,
       final Bytes from,
       final Bytes to,
-      final long minValidTs
+      final long streamTimeMs
   ) {
+    // TODO(sophie): filter by minValidTs if based on key or default only
```
agavra, line 222: Same comment about TODOs

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -151,10 +156,15 @@ public static PartitionedOperations create(
       storeRegistry.registerStore(registration);
 
       final boolean migrationMode = ConfigUtils.responsiveMode(config) == ResponsiveMode.MIGRATE;
-      long startingTimestamp = -1;
-      final Optional<Duration> ttl = params.timeToLive();
-      if (migrationMode && ttl.isPresent()) {
-        startingTimestamp = Instant.now().minus(ttl.get()).toEpochMilli();
+      long startTimeMs = -1;
+      if (migrationMode && params.ttlProvider().isPresent()) {
+        // TODO(sophie): figure out how to account for row-level ttl in migration mode
```
agavra, line 161: let's remove this TODO and just create a ticket since the code isn't wrong (it's just an unsupported feature at the moment)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -182,14 +192,16 @@ public static PartitionedOperations create(
   }
 
   private static RemoteKVTable<?> createInMemory(final ResponsiveKeyValueParams params) {
+    // TODO(sophie): implement ttl for in-memory store?
```
agavra, line 195: same as above (remove TODO and create a ticket) we should probably throw an exception though if there is a non-empty TTL Resolver

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StateDeserializer.java
```diff
@@ -0,0 +1,68 @@
...
+      timestampedValueDeserializer =
+          Optional.of(new ValueAndTimestampSerde<>(valueSerde).deserializer());
+    } else {
+      timestampedValueDeserializer = Optional.empty();
+    }
+  }
+
+  public K keyFrom(final byte[] keyBytes) {
+    if (keyBytes == null || keyDeserializer == null) {
+      throw new IllegalStateException("Attempted to deserialize unprovided key");
```
agavra, line 51: Let's add (a) which one was null (I know it should be both, but still makes it easier to debug) and (b) an actionable error message, something like "you must specify the keyDeserializer when creating the TtlProvider"

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/Utils.java
```diff
@@ -122,4 +122,8 @@ public static String extractThreadIdFromThreadName(final String threadName) {
     return threadName;
   }
 
+  public static long millisToSeconds(final long millis) {
+    return millis / 1000L;
+  }
```
agavra, line 127: nit we should prefer `TimeUnit.MILLISECONDS.toSeconds(millis)` and just remove this method (it handles more edge cases that I don't believe we can ever hit but it's good to have)

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
```diff
@@ -0,0 +1,285 @@
...
+      default:
+        throw new IllegalStateException("Unrecognized ttl type: " + ttlType);
+    }
+    final Optional<TtlDuration> rowTtlOverride = computeTtl.apply(key, value);
+
+    // Just return Optional.empty if the row override is actually equal to the default
+    // to help simplify handling logic for tables
+    if (rowTtlOverride.isPresent() && rowTtlOverride.get().equals(defaultTtl)) {
+      return Optional.empty();
+    }
```
agavra, line 281: I'd prefer we removed this optimization unless it proves necessary -- does it actually simplify things? I imagine we shouldn't be comparing the result of this to the default anywhere in the code

It works with today's implementation, but if we ever have partial updates it's risky (as in I inserted a row with a non-default TTL and then I update it to be the default TTL I should make sure that every scenario would override the old TTL)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/mongo/MongoKVTable.java
```diff
@@ -150,21 +164,63 @@ public MongoKVFlushManager init(final int kafkaPartition) {
   }
 
   @Override
-  public byte[] get(final int kafkaPartition, final Bytes key, final long minValidTs) {
-    final KVDoc v = docs.find(Filters.and(
-        Filters.eq(KVDoc.ID, keyCodec.encode(key)),
-        Filters.gte(KVDoc.TIMESTAMP, minValidTs)
-    )).first();
-    return v == null ? null : v.getValue();
+  public byte[] get(final int kafkaPartition, final Bytes key, final long streamTimeMs) {
```
agavra, line 167: Let's make sure to add tests for MongoDB as well. I'd actually prefer if we just threw an exception for this PR if there's anything but a static default TTL so we can have all the mongo changes be part of a future PR.

kafka-client/src/test/java/dev/responsive/kafka/integration/RowLevelTtlIntegrationTest.java
```diff
@@ -0,0 +1,220 @@
...
+  private String inputTopic() {
+    return name + "." + INPUT_TOPIC;
+  }
+
+  private String outputTopic() {
+    return name + "." + OUTPUT_TOPIC;
+  }
+
+  @Test
+  public void test() throws Exception {
```
agavra, line 120: is this test complete? (also re: TODO below)

kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraFactTableIntegrationTest.java
```diff
@@ -141,57 +150,225 @@ public void shouldUseTwcsWithoutTtl() throws Exception {
...
+    final long insertTimeMs = 0L;
+    client.execute(table.insert(1, noTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, defaultTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, tenMinTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, fiftyMinTtlKey, val, insertTimeMs));
 
     // Then
-    assertThat(valAt0, is(val));
-    assertThat(valAt1, nullValue());
+    long lookupTime = Duration.ofMinutes(11).toMillis();
```
agavra, line 199: no need to change it for this PR, but just in the future you can use `TimeUnit.MINUTES.toMillis(11)` etc.. for all of these. IMO a little easier to read and has no object allocations 😅

kafka-client/src/test/java/dev/responsive/kafka/internal/db/AbstractCassandraKVTableIntegrationTest.java
ableegoldman: It's not about the static helper methods it's about the test themselves. We want to run largely the same set of tests for both of these (same insert semantics, same ttl semantics, etc) so I was going to extract all the common tests into this class. 

If you'd prefer the @Parametrize approach I can take a look at that, but since not all the tests are going to be the same for the kv vs fact tables, I thought the abstract test format would be cleaner. Does that make sense?

(That said, ideally we can share the same tests for the Mongo table as well, in which case another approach might be easier in the end 🤷‍♀️ )

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
ableegoldman: https://www.notion.so/responsivedev/Support-row-level-ttl-in-migration-mode-12927549be4c8065b9e1c92ed0f9030f?pvs=4

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
ableegoldman: https://www.notion.so/responsivedev/Support-ttl-for-in-memory-store-12927549be4c80199207f894273c0b03?pvs=4

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/StateDeserializer.java
ableegoldman: This should indicate an internal/Responsive error, since we should never attempt to deserialize a null key/value where the key/value deserializer is null. So there is no actionable message for users, it's for us. But ack, I'll log which was null

ableegoldman: Just to clarify since I know we discussed some alternatives where this wouldn't be true, it's not possible to set a key-based ttlProvider without specifying the keyDeserializer. That's why this is an internal error not a user error

agavra: yup, I did understand it was impossible! just would rather make it future proof in case anything changes. good erorr messages are always good :)

PR #370, opened 2024-10-24
Row-level TTL PR 1: new API

ableegoldman:
First PR for row-level ttl. Includes only the API changes and supporting code

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
@@ -50,12 +50,12 @@ public final class ResponsiveKeyValueParams {
   }
 
   public ResponsiveKeyValueParams withTtlProvider(final TtlProvider<?, ?> ttlProvider) {
-    if (ttlProvider.hasConstantTtl() && !ttlProvider.defaultTtl().isFinite()) {
-      throw new IllegalArgumentException("Passed in ttlProvider with infinite ttl, "
-                                             + "this is equivalent to no ttl and you "
-                                             + "should not set the ttlProvider for this");
+    // If ttl is constant and infinite, it's equivalent to having no ttl at all
+    if (ttlProvider.hasDefaultOnly() && !ttlProvider.defaultTtl().isFinite()) {
+      this.ttlProvider = Optional.empty();
+    } else {
+      this.ttlProvider = Optional.of(ttlProvider);
     }
-    this.ttlProvider = Optional.of(ttlProvider);
     return this;
   }
 
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
@@ -16,18 +16,18 @@
 
 package dev.responsive.kafka.api.stores;
 
+import dev.responsive.kafka.api.stores.TtlProvider.TtlDuration;
 import dev.responsive.kafka.internal.stores.SchemaTypes.KVSchema;
 import dev.responsive.kafka.internal.utils.TableName;
 import java.time.Duration;
 import java.util.Optional;
-import javax.annotation.Nullable;
 
 public final class ResponsiveKeyValueParams {
 
   private final TableName name;
   private final KVSchema schema;
 
-  @Nullable private Duration timeToLive = null;
+  private Optional<TtlProvider<?, ?>> ttlProvider = Optional.empty();
 
   private ResponsiveKeyValueParams(
       final String name,
@@ -46,7 +46,16 @@ public final class ResponsiveKeyValueParams {
   }
 
   public ResponsiveKeyValueParams withTimeToLive(final Duration timeToLive) {
-    this.timeToLive = timeToLive;
+    return withTtlProvider(TtlProvider.withDefault(timeToLive));
+  }
+
+  public ResponsiveKeyValueParams withTtlProvider(final TtlProvider<?, ?> ttlProvider) {
+    if (ttlProvider.hasConstantTtl() && !ttlProvider.defaultTtl().isFinite()) {
+      throw new IllegalArgumentException("Passed in ttlProvider with infinite ttl, "
+                                             + "this is equivalent to no ttl and you "
+                                             + "should not set the ttlProvider for this");
+    }
+    this.ttlProvider = Optional.of(ttlProvider);
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -56,7 +56,9 @@ public final class ResponsiveStores {
    * @return a supplier for a key-value store with the given options
    *         that uses Responsive's storage for its backend
    */
-  public static KeyValueBytesStoreSupplier keyValueStore(final ResponsiveKeyValueParams params) {
+  public static ResponsiveKeyValueBytesStoreSupplier keyValueStore(
```
ableegoldman, line 59: @agavra FYI all the changes in ResponsiveStores are technically unrelated -- just holdovers from an alternative API approach I abandoned -- but I think it's the right thing to do anyways so I kept it in. Hope you don't mind

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
```diff
@@ -0,0 +1,295 @@
...
+      }
+      return new TtlDuration(ttl, Ttl.FINITE);
+    }
+
+    // No ttl will be applied, in other words infinite retention
+    public static TtlDuration noTtl() {
+      return new TtlDuration(Duration.ZERO, Ttl.INFINITE);
+    }
+
+    // TODO(sophie): store ttl as long to avoid Duration conversions on the hot path
```
ableegoldman, line 162: @agavra I'm actually a bit concerned about this. Even if we store it as millis we'll still have a Duration on the hot path since this is the return type. What if we just bite the bullet and have the user return a TTL in terms of millis rather than a Duration?

I know it's not as nice API-wise but To be honest I think most Streams users are probably pretty used to reasoning about time in terms of milliseconds. So I'm personally leaning towards replacing Duration with long millis entirely. Thoughts?

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
```diff
@@ -46,7 +46,16 @@ public static ResponsiveKeyValueParams fact(final String name) {
   }
 
   public ResponsiveKeyValueParams withTimeToLive(final Duration timeToLive) {
-    this.timeToLive = timeToLive;
+    return withTtlProvider(TtlProvider.withDefault(timeToLive));
+  }
+
+  public ResponsiveKeyValueParams withTtlProvider(final TtlProvider<?, ?> ttlProvider) {
+    if (ttlProvider.hasConstantTtl() && !ttlProvider.defaultTtl().isFinite()) {
```
agavra, line 53:
```suggestion
    if (ttlProvider.hasDefaultOnly() && !ttlProvider.defaultTtl().isFinite()) {
```
nit: `hasConstantTtl` trips me up a little bit, might as well just directly say what it is

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
```diff
@@ -46,7 +46,16 @@ public static ResponsiveKeyValueParams fact(final String name) {
...
+      throw new IllegalArgumentException("Passed in ttlProvider with infinite ttl, "
+                                             + "this is equivalent to no ttl and you "
+                                             + "should not set the ttlProvider for this");
```
agavra, line 56: just wondering, why not just do nothing (i.e. leave it as Optional.empty and just return early)? I'd rather not throw a bunch of errors as that makes it stricter in the way we allow usage (I actually like having "NOOP" instances instead of nulls/empty in Java, so as a user, especially in tests, I may consider doing that)

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
```diff
@@ -0,0 +1,295 @@
...
+
+  /**
+   * Creates a new TtlProvider that has no default (equivalent to infinite retention for
+   * all records unless an override is specified).  Must be used in combination with
+   * exactly one of the {@link #fromKey(Function, Serde)}, {@link #fromValue(Function, Serde)},
+   * and {@link #fromKeyAndValue(BiFunction, Serde, Serde)} methods to define the row-level
+   * override function.
+   * <p>
+   * If no ttl whatsoever is desired for this store, simply don't configure a TtlProvider
+   * to begin with.
```
agavra, line 55: same comment as above, just leaving this here as a reminder to change the docs if you agree

kafka-client/src/main/java/dev/responsive/kafka/api/stores/TtlProvider.java
agavra: Do we have evidence that this is actually significant on a per-retrieved doc basis? 

A few bitwise CPU operations seems to be dwarfed by the potential network call + the re-serialization in my opinion. Even if hits the in-memory cache, we're talking about loads from main memory into L1 cache.

The cost of creating a `Duration` is ~15ns (thanks ChatGPT for the estimation, looks legit to me, and half of that cost is just object allocation) and the cost of calling `toMillis` is ~1ns (assuming there's no nanosecond component), so we're talking 16ns per-key returned overhead for using Duration on a 3.2Ghz CPU.

agavra (approved):
LGTM! Just nits inline

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
ableegoldman: I guess to me, "has constant" is more "what it is" than "default only" hahaha. But I can also see it your way. I'll change it

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveKeyValueParams.java
ableegoldman: oh I actually would prefer to just silently convert it! I had the impression from previous interactions that you strongly preferred to throw an exception when the user does something that doesn't make sense rather than just correct it? Anyways I'm happy to change it

agavra: Now that you mention it... 😆 In cases where it's ambiguous I'd prefer throwing an error, but in this case I think we _are_ doing what they want, we're just optimizing it away underneath the hood. (i.e. it's not invalid to set a default + infinite TTL, it's just the same as the default behavior)

PR #371, opened 2024-10-24
Row-level TTL PR 2: TtlResolver and RemoteTableSpec cleanup

ableegoldman:
Introduces the TtlResolver used to hook everything up within Responsive. Also cleans up the RemoteTableSpec

PR 1: API https://github.com/responsivedev/responsive-pub/pull/370

Future TODO (in order)
1. Cassandra Fact tables (insert/get)
2. TopologyTestDriver
3. Mongo KV tables (insert/get/range/all)
4. Cassandra KVTables (insert/get/range/all)

PR #373, opened 2024-10-25
Row-level TTL PR 2: compute minValidTs from TtlResolver

ableegoldman:
Short PR to move the computation of `minValidTs` from the upper layer of the KV stores to the inner Table implementations where we can account for the TtlResolver

Introduces the TtlResolver used to hook everything up within Responsive. Also cleans up the RemoteTableSpec

Future TODO (in order)
1. Cassandra Fact tables (insert/get) <--- this is the next PR
2. TopologyTestDriver
3. Mongo KV tables (insert/get/range/all)
4. Cassandra KVTables (insert/get/range/all)

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
@@ -194,10 +194,11 @@ public class PartitionedOperations implements KeyValueOperations {
       final ResponsiveKeyValueParams params,
       final Optional<TtlResolver<?, ?>> ttlResolver
   ) {
-    if (ttlResolver.isPresent()) {
-      throw new UnsupportedOperationException("ttl is not yet supported for in-memory stores");
+    if (ttlResolver.isPresent() && !ttlResolver.get().hasDefaultOnly()) {
+      throw new UnsupportedOperationException("Row-level ttl is not yet supported "
+                                                  + "for in-memory stores");
     }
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -316,7 +317,7 @@ public byte[] get(final Bytes key) {
     return table.get(
         changelog.partition(),
         key,
-        minValidTimestamp()
+        currentRecordTimestamp()
```
ableegoldman, line 320: @agavra hey I was wondering if this should really be the current record timestamp or if streamTime is more accurate?

Or does it not really matter since we're approximating streamTime = currentTimestamp = wallclock time for the ttl semantics anyways

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveSessionStore.java
```diff
@@ -228,4 +231,11 @@ public KeyValueIterator<Windowed<Bytes>, byte[]> fetch(final Bytes from, final B
   private long minValidEndTimestamp() {
     return observedStreamTime - params.retentionPeriod() + 1;
   }
+
+  public void restoreBatch(final Collection<ConsumerRecord<byte[], byte[]>> records) {
```
ableegoldman, line 235: happened to notice we weren't updating the streamTime during restoration for window or session stores either

agavra (approved):

PR #374, opened 2024-10-28
Row-level TTL PR 4: CassandraFactTable implementation

ableegoldman:
Implement row-level ttl for Scylla fact tables

PR 3: compute minValidTs from TtlResolver https://github.com/responsivedev/responsive-pub/pull/373

Planned followup PRs (in order)
1. TopologyTestDriver
2. Mongo KV tables (insert/get/range/all)

Unplanned future work (can be picked up ad-hoc as needed)
1. Row-level ttl for Cassandra KVTables (insert/get/range/all)
2. Row-level ttl for in-memory KV tables
3. General ttl for global tables
4. Row-level ttl with migration mode

### DIFF
--- a/kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraFactTableIntegrationTest.java
+++ b/kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraFactTableIntegrationTest.java
@@ -18,6 +18,8 @@ package dev.responsive.kafka.internal.db;
 
 import static dev.responsive.kafka.internal.db.partitioning.TablePartitioner.defaultPartitioner;
 import static dev.responsive.kafka.internal.stores.TtlResolver.NO_TTL;
+import static dev.responsive.kafka.testutils.SerdeUtils.serializedKey;
+import static dev.responsive.kafka.testutils.SerdeUtils.serializedValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.containsStringIgnoringCase;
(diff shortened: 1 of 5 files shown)


agavra (approved):
nice, LGTM! missing tests, though I think that'll come in a future PR? you had them in your initial big PR

ableegoldman:
@agavra  you got in with the review faster than I could copy over and push the tests 😛  -- they're in this PR

kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraFactTableIntegrationTest.java
```diff
@@ -184,4 +187,220 @@ public void shouldRespectSemanticDefaultOnlyTtlForLookups() throws Exception {
...
+
+    // When:
+    final long insertTimeMs = 0L;
+    client.execute(table.insert(1, noTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, defaultTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, tenMsTtlKey, val, insertTimeMs));
+    client.execute(table.insert(1, fiftyMsTtlKey, val, insertTimeMs));
+
+    // Then
+    long lookupTime = Duration.ofMillis(11).toMillis();
```
agavra, line 232: hmm, now that I'm thinking about it we may want to use two orders of magnitude larger TTLs than ~10ms because (a) Cassandra only has TTL on order of seconds, so these would all translate to 0 seconds IIRC and (b) we don't want the TTl to pass and the row to be physically expired before the lookup happens.

ableegoldman: mm good call

ableegoldman: @agavra since the build already finished/passed on this PR I'm going to go ahead and merge it as-is, but will address this in the next PR if that's ok

PR #375, opened 2024-10-29
Row-level TTL PR 5: automatically extract StateSerdes

ableegoldman:
Avoid forcing the user to pass in their serdes again by extracting the StateSerdes from the MeteredKeyValueStore layer of the StateStore hierarchy.

Also addresses feedback from PR 4 to increase the ttl times used in the CassandraKV/FactTableIntegrationTest tests

PR 1: API https://github.com/responsivedev/responsive-pub/pull/370
PR 2: TtlResolver https://github.com/responsivedev/responsive-pub/pull/371
PR 3: compute minValidTs from TtlResolver https://github.com/responsivedev/responsive-pub/pull/373
PR 4: CassandraFactTable implementation https://github.com/responsivedev/responsive-pub/pull/374

Planned followup PRs (in order)
1. TopologyTestDriver <-- next/that before release cut
2. Mongo KV tables (insert/get/range/all)

Unplanned future work (can be picked up ad-hoc as needed)
1. Row-level ttl for Cassandra KVTables (insert/get/range/all)
2. Row-level ttl for in-memory KV tables
3. General ttl for global tables
4. Row-level ttl with migration mode

PR #376, opened 2024-10-29
Row-level TTL PR 6: TopologyTestDriver

ableegoldman:
Support for row-level ttl in the TopologyTestDriver, as well as the InMemoryKeyValueTable which we now use as the backing kv store for the Responsive TTD.

PR 3: compute minValidTs from TtlResolver https://github.com/responsivedev/responsive-pub/pull/373
PR 4: CassandraFactTable implementation https://github.com/responsivedev/responsive-pub/pull/374
PR 5: extract StateSerdes https://github.com/responsivedev/responsive-pub/pull/375

Planned followup PRs
1. Row-level ttl for all APIs in MongoKVTables: insert/get/range/all

Unplanned future work (can be picked up ad-hoc as needed)
1. Row-level ttl for all APIs in Cassandra KVTable: insert/get/range/all
2. Row-level ttl for range APIs in TTDKeyValueTable/InMemoryKVTable: range/all
3. General ttl for global tables
4. Row-level ttl with migration mode


responsive-test-utils/src/main/java/dev/responsive/kafka/internal/db/TTDKeyValueTable.java
```diff
@@ -16,152 +16,48 @@
...
+    this.client = client;
   }
 
   @Override
-  public BoundStatement delete(final int kafkaPartition, final Bytes key) {
-    stub.delete(key);
-    return null;
-  }
+  public byte[] get(final int kafkaPartition, final Bytes key, final long streamTimeMs) {
+    // trigger flush before lookup since CommitBuffer doesn't apply ttl
```
ableegoldman, line 36: this is a bit hacky, but...it works

responsive-test-utils/src/main/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriver.java
```diff
@@ -97,22 +97,21 @@ public ResponsiveTopologyTestDriver(
...
+        new TTDCassandraClient(new TTDMockAdmin(baseProps(config), topology))
     );
   }
 
   /**
-   * Advances the internal mock time used for Responsive-specific features such as ttl, as well
-   * as the mock time used for various Kafka Streams functionality such as wall-clock punctuators.
+   * Advances the internal mock time which can be used to trigger some Kafka Streams
+   * functionality such as wall-clock punctuators, or force Responsive to flush
+   * its internal buffers.
```
ableegoldman, line 107: Thanks to [this ](https://github.com/responsivedev/responsive-pub/pull/376/files#r1820067634)"hack" we no longer require users to use this method to advance wall clock time and manually flush the CommitBuffer in their TTD apps. Generally this is only used for triggering punctuation in Streams so most people wouldn't already be using this in their apps, and will advance time via record timestamps

responsive-test-utils/src/test/java/dev/responsive/kafka/api/ResponsiveTopologyTestDriverKeyValueStoreTest.java
```diff
@@ -98,32 +109,36 @@ public void shouldRunAllKVStoreTypesWithTTLWithoutResponsiveConnection(final KVS
         "output", new StringDeserializer(), new StringDeserializer());
 
     // When:
-    people.pipeInput("1", "1,alice,CA"); // time = 0
-    driver.advanceWallClockTime(Duration.ofMillis(5));
```
ableegoldman, line 102: This test now reflects how a regular Streams app would function and how a regular TTD test would be written

kafka-client/src/main/java/dev/responsive/kafka/internal/db/inmemory/InMemoryKVTable.java
```diff
@@ -22,7 +24,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class InMemoryKVTable implements RemoteKVTable<Object> {
+// Note: this class doesn't actually use BoundStatements and applies all operations immediately,
+// we just stub the BoundStatement type so we can reuse this table for the TTD
+public class InMemoryKVTable implements RemoteKVTable<BoundStatement> {
```
ableegoldman, line 29: I suppose it's worth noting that this table implementation doesn't actually apply ttl in terms of retention, only in terms of filtering results. I figure this is fine for the TTD since such test should be short-lived, but good to note in case we ever want to do anything else with this class 🤷‍♀️

agavra (approved):
I can't say I love the hack to flush on get, and it may come back to bite us if we want to expand the usage of TTD in the future, but I also think we might want to just totally rethink testing harnesses for Kafka Streams in the future so I'm OK to just get this out the door for now and revisit it later.

That being said, should CommitBuffer respect TTL? I'm OK if it doesn't because flushes generally happen pretty quickly, but that's not something I thought about until now.

PR #383, opened 2024-11-11
Change test harness default to Mongo and reduce integration test runtime

ableegoldman:
Changes the default storage backend type to Mongo for both the ResponsiveConfig property and the ResponsiveExtension test harness

Also tries to reduce testing runtime in two ways:
1.  Reusing the ResponsiveExtension for all integration tests so we don't have to start and stop all the containers for each new test class. From my local testing this seems to cut the total test runtime down by about 3 minutes
2. Start up containers in parallel, which has a minimal impact on total runtime now that we only start them up once, but reduces the startup for individual tests by about 10s which is nice for debugging cases where you're running a test multiple times


agavra:
image
Did you mean to commit these files?

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -66,7 +66,7 @@ public class ResponsiveConfig extends AbstractConfig {
 
   public static final String STORAGE_BACKEND_TYPE_CONFIG = "responsive.storage.backend.type";
   private static final String STORAGE_BACKEND_TYPE_DOC = "The storage backend";
-  private static final StorageBackend STORAGE_BACKEND_TYPE_DEFAULT = StorageBackend.CASSANDRA;
+  private static final StorageBackend STORAGE_BACKEND_TYPE_DEFAULT = StorageBackend.MONGO_DB;
```
agavra, line 69: oh I'm a little anxious about this, not sure whether or not existing customers set this explicitly or not. please check with them first!

I originally thought you were just going to change the default in the test configurations.

kafka-client/src/test/java/dev/responsive/kafka/api/ResponsiveKafkaStreamsTest.java
```diff
@@ -90,7 +91,7 @@ public Consumer<byte[], byte[]> getRestoreConsumer(final Map<String, Object> con
     }
   };
 
-  private final CassandraClientFactory mockCassandryFactory = new CassandraClientFactory() {
+  private final CassandraClientFactory mockCassandraFactory = new CassandraClientFactory() {
```
agavra, line 94: 😆

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveKeyValueStoreRestoreIntegrationTest.java
```diff
@@ -362,35 +357,20 @@ private RemoteKVTable<?> remoteKVTable(
       final TopicPartition changelog
   ) throws InterruptedException, TimeoutException {
     final RemoteKVTable<?> table;
-    if (EXTENSION.backend == StorageBackend.CASSANDRA) {
+
+    if (type == KVSchema.FACT) {
```
agavra, line 361: it looks like we're removing the ability to test with cassandra and instead tying that to the KVSchema type? Am I reading that correctly?

kafka-client/src/test/java/dev/responsive/kafka/api/ResponsiveKafkaStreamsTest.java
```diff
@@ -126,6 +127,7 @@ public void setUp() {
 
     properties.put(RESPONSIVE_ORG_CONFIG, "responsive");
     properties.put(RESPONSIVE_ENV_CONFIG, "test");
+    properties.put(ResponsiveConfig.STORAGE_BACKEND_TYPE_CONFIG, StorageBackend.CASSANDRA.name());
```
agavra, line 130: I thought the point of this PR is to change our tests to run with Mongo?

agavra (commented):
I'm a little confused about this PR, it seems like we're changing the default to run with Mongo but having all our integration tests specifically test Cassandra? I feel like they should be aligned to make sure we're testing our defaults

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
ableegoldman: ah sorry, good call. We should check with them though because I think we do want to shift to Mongo as the default (or do we?)

I'll extract this out into a separate PR that we can merge if/when we're ready. Does it make sense to change the default for the tests before we change the default itself?

kafka-client/src/test/java/dev/responsive/kafka/integration/ResponsiveKeyValueStoreRestoreIntegrationTest.java
ableegoldman: Just for this specific test (sorry I meant to leave a comment to clarify before you read this)

My thought was, since we still have at least one customer running Cassandra fact stores then it makes sense to keep some tests parametrized to run that. But no one is running Cassandra KV stores, we would recommend Mongo for that, so we might as well change this to run the KV schema variation against mongo.

Is the method of parametrizing confusing though? I can change it so that the storage backend is the parameter and we determine the schema type based on that.

kafka-client/src/test/java/dev/responsive/kafka/api/ResponsiveKafkaStreamsTest.java
ableegoldman: Some of them seemed to rely on Cassandra (eg for fault injection, like in this case) and I didn't want to block the entire PR on updating every test like this. So the ones you see where I set Cassandra explicitly were the ones that failed because they did something Cassandra specific.

Usually the test had some hook through a mock cassandra client, so we'd "just" need to switch that setup over to use mongo. I say "just" in quotes because my impression is this might be a nontrivial effort

kafka-client/src/test/java/dev/responsive/kafka/integration/TablePartitionerIntegrationTest.java
```diff
@@ -84,12 +85,14 @@
...
 import org.junit.jupiter.api.TestInfo;
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.testcontainers.containers.CassandraContainer;
 
-@ExtendWith(ResponsiveExtension.class)
 public class TablePartitionerIntegrationTest {
 
+  @RegisterExtension
+  static ResponsiveExtension EXTENSION = new ResponsiveExtension(StorageBackend.CASSANDRA);
```
ableegoldman, line 94: The table partitioner is an exclusively cassandra concept

kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraFactTableIntegrationTest.java
```diff
@@ -50,12 +51,14 @@
...
 import org.junit.jupiter.api.TestInfo;
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.testcontainers.containers.CassandraContainer;
 
-@ExtendWith(ResponsiveExtension.class)
 class CassandraFactTableIntegrationTest {
 
+  @RegisterExtension
+  static ResponsiveExtension EXTENSION = new ResponsiveExtension(StorageBackend.CASSANDRA);
```
ableegoldman, line 60: Obviously this test is cassandra specific 😄 (ditto the one below)

kafka-client/src/test/java/dev/responsive/kafka/internal/stores/CommitBufferTest.java
```diff
@@ -101,18 +102,22 @@
...
 import org.testcontainers.containers.CassandraContainer;
 
-@ExtendWith({MockitoExtension.class, ResponsiveExtension.class})
+@ExtendWith({MockitoExtension.class})
 @MockitoSettings(strictness = Strictness.LENIENT)
 
 public class CommitBufferTest {
 
+  @RegisterExtension
+  static ResponsiveExtension EXTENSION = new ResponsiveExtension(StorageBackend.CASSANDRA);
```
ableegoldman, line 119: We should eventually migrate this test to Mongo but that will be a pretty huge shift, so imo it makes sense to leave this one as Cassandra for now since it's pretty much hardcoded into the test setup

kafka-client/src/test/java/dev/responsive/kafka/integration/GlobalStoreIntegrationTest.java
```diff
@@ -56,11 +57,13 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInfo;
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-@ExtendWith(ResponsiveExtension.class)
 public class GlobalStoreIntegrationTest {
 
+  @RegisterExtension
+  static ResponsiveExtension EXTENSION = new ResponsiveExtension(StorageBackend.CASSANDRA);
```
ableegoldman, line 65: global stores are incompatible with Mongo(I guess we should fix this...?)

PR #385, opened 2024-11-15
add support types for license verification

rodesai:
This patch adds some infra to support license verification:

The main change is the addition of LicenseAuthenticator and LicenseChecker. LicenseAuthenticator is responsible for taking a license document, verifying that it is signed correctly, and then returning the contained signed LicenseInfo. LicenseChecker takes a LicenseInfo and validates that the license it describes can be used.

These new classes operate on the pojo types defined under the model package that define the structure of the license documents, and the key registry (SigningKeys) that's to be packaged into the jar.

Finally, this patch also includes a utility class for parsing pem files, which we'll use to store public keys. I opted to write our own parser (rather than using something like bouncycastle) because the format it's simple and I'd rather not include a dependency on a foundational artifact like bouncycastle that is likely to conflict with a user's dependencies.

### DIFF
--- a/kafka-client/src/test/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStoreTest.java
+++ b/kafka-client/src/test/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStoreTest.java
@@ -45,7 +45,7 @@ class ResponsiveKeyValueStoreTest {
   public void shouldPutIfAbsent() {
     // Given:
     final var store = new ResponsiveKeyValueStore(
-        ResponsiveKeyValueParams.keyValue("license-test"),
+        ResponsiveKeyValueParams.keyValue("test"),
         (params, ttlResolver, context, type) -> ops
     );
     store.init((StateStoreContext) context, root);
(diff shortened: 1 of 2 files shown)


kafka-client/src/test/java/dev/responsive/kafka/internal/stores/ResponsiveKeyValueStoreTest.java
```diff
@@ -45,7 +45,7 @@ class ResponsiveKeyValueStoreTest {
   public void shouldPutIfAbsent() {
     // Given:
     final var store = new ResponsiveKeyValueStore(
-        ResponsiveKeyValueParams.keyValue("test"),
+        ResponsiveKeyValueParams.keyValue("license-test"),
```
rodesai, line 48: damn it intellij you're not as smart as you think

kafka-client/src/main/java/dev/responsive/kafka/internal/license/PublicKeyPemFileParser.java
```diff
@@ -0,0 +1,56 @@
...
+
+package dev.responsive.kafka.internal.license;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Base64;
+import java.util.List;
+
+public class PublicKeyPemFileParser {
```
rodesai, line 25: I could have used bouncycastle to parse these, but I didn't want to bring it in as a dependency just to do that because it's likely the user will also depend on bouncycastle (possibly at a conflicting version).

agavra (approved):
LGTM

PR #386, opened 2024-11-15
add configs to accept license and verify it

rodesai:
This patch makes the following changes to integrate license verification:
1. Adds responsive.license and responsive.license.file configs to allow the user to configure their license.
2. ResponsiveKafkaStreams validates the license if the user has not configured a cloud API key. The user could technically get around the license verification by configuring a bogus key. We assume that we won't have malicious users for now, and defer a more rigorous check to a future PR. We should do the check in the background and not block starting the app so that we don't depend on our cloud for the user's application to run.
3. Moves around some of the test license files so that we can use them from different tests.


kafka-client/src/main/resources/license-keys/license-keys.json
```diff
@@ -0,0 +1 @@
+{"keys":  []}
```
agavra, line 1: I guess when we create the keys well update it here?

agavra (approved):
I feel like now that we have this all of our integration tests will fail, we probably need to add something to `ResponsiveExtension` so that it will use a test licensek ey

kafka-client/src/main/resources/license-keys/license-keys.json
rodesai: yep, that's right!

PR #388, opened 2024-11-16
Row-level TTL 10: expire based on event time

ableegoldman:
Offset the actual row-level ttl set in the storage backend by the difference between the current system time and the record timestamp or "event time". This allows users to process historical data without holding seeing a spike in storage as the ttl will now expire records based on their timestamp rather than insertion time


ableegoldman:
Closing this PR since we decided against implementing this feature for the time being. We can revive it if/when we want to allow expiring based on event time

PR #392, opened 2024-11-19
DSL w/ async: e2e implementation for stream-table join 

ableegoldman:
This PR is mainly a POC to show that it can be done -- we'll need to wait for 4.0 to be released before merging this. It also includes the changes we'll need to make for that upgrade (eg removal of deprecated APIs)

Until we can merge and release this new feature, it has to be run on top of my WIP KIP-1112 branch ([here](https://github.com/apache/kafka/pull/17833)).

Remaining work beyond this POC:
-AK: converting all of the other stateful DSL operators to implement the ProcessorSupplier#stores method
-Responsive: fill in the async store builders with all the store types (eg versioned stores, window and session stores)

### DIFF
--- a/kafka-client-bootstrap/build.gradle.kts
+++ b/kafka-client-bootstrap/build.gradle.kts
@@ -15,6 +15,10 @@ plugins {
     id("responsive.docker")
 }
 
+repositories {
+    mavenLocal()
+}
+
 application {
     mainClass.set("dev.responsive.kafka.bootstrap.main.Main")
 }
--- a/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/tests/KeyBatchExample.java
+++ b/kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/tests/KeyBatchExample.java
@@ -32,9 +32,10 @@ import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.kstream.Consumed;
 import org.apache.kafka.streams.kstream.Produced;
-import org.apache.kafka.streams.kstream.Transformer;
-import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.processor.PunctuationType;
+import org.apache.kafka.streams.processor.api.Processor;
+import org.apache.kafka.streams.processor.api.ProcessorContext;
+import org.apache.kafka.streams.processor.api.Record;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.Stores;
@@ -70,7 +71,7 @@ public class KeyBatchExample extends AbstractKSExampleService {
     }
 
     builder.stream(ORDERS, Consumed.with(Serdes.String(), RegressionSchema.orderSerde()))
-        .transform(BatchTransformer::new, "grouped-orders-store")
+        .process(BatchTransformer::new, "grouped-orders-store")
         .peek((k, v) -> {
           if (responsive) {
             final var random = Math.abs(randomGenerator.nextLong() % 10000);
@@ -84,14 +85,13 @@ public class KeyBatchExample extends AbstractKSExampleService {
     return builder.build();
   }
 
-  private static class BatchTransformer
-      implements Transformer<String, Order, KeyValue<String, GroupedOrder>> {
+  private static class BatchTransformer implements Processor<String, Order, String, GroupedOrder> {
 
-    private ProcessorContext context;
+    private ProcessorContext<String, GroupedOrder> context;
     private KeyValueStore<String, StoredOrder> store;
 
     @Override
-    public void init(final ProcessorContext context) {
+    public void init(final ProcessorContext<String, GroupedOrder> context) {
       this.context = context;
       this.store = context.getStateStore("grouped-orders-store");
       this.context.schedule(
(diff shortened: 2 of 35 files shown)


kafka-client-bootstrap/build.gradle.kts
```diff
@@ -15,6 +15,10 @@ plugins {
     id("responsive.docker")
 }
 
+repositories {
```
ableegoldman, line 18: Had to put this into all the build.gradle files so I could compile from my local AK branch. I can remove all of these before merging if anyone feels strongly about it, but if it doesn't do any harm then I'd rather keep these for the next time I'm working off of a dev branch

kafka-client-examples/e2e-test/src/main/java/dev/responsive/examples/regression/tests/KeyBatchExample.java
```diff
@@ -70,7 +71,7 @@ protected Topology buildTopology() {
     }
 
     builder.stream(ORDERS, Consumed.with(Serdes.String(), RegressionSchema.orderSerde()))
-        .transform(BatchTransformer::new, "grouped-orders-store")
+        .process(BatchTransformer::new, "grouped-orders-store")
```
ableegoldman, line 74: Not part of async -- related to the 4.0 upgrade

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -284,16 +284,16 @@ private static Properties propsWithOverrides(
       return propsWithOverrides;
     }
 
-    final Object o = configs.originals().get(InternalConfig.INTERNAL_TASK_ASSIGNOR_CLASS);
+    final Object o = configs.originals().get(StreamsConfig.TASK_ASSIGNOR_CLASS_CONFIG);
```
ableegoldman, line 287: Not part of async -- related to the 4.0 upgrade

kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncProcessorSupplier.java
```diff
@@ -132,32 +133,30 @@ public final class AsyncProcessorSupplier<KIn, VIn, KOut, VOut>
...
 
   @Override
   public Set<StoreBuilder<?>> stores() {
+    maybeInitializeAsyncStoreBuilders();
     return new HashSet<>(asyncStoreBuilders.values());
   }
 
+  private void maybeInitializeAsyncStoreBuilders() {
+    if (asyncStoreBuilders == null) {
+      asyncStoreBuilders = initializeAsyncBuilders(userProcessorSupplier.stores());
```
ableegoldman, line 159: We need to delay the initialization of the async store builders because the call to the inner processor's #stores method might return different things at different points in the build process (for example, if a downstream operator forces materialization, the #stores will return null/empty until that downstream operator is processed)

kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncProcessorWrapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class AsyncProcessorWrapper implements ProcessorWrapper {
+
+
+  @Override
+  public <KIn, VIn, KOut, VOut> ProcessorSupplier<KIn, VIn, KOut, VOut> wrapProcessorSupplier(
+      final String processorName,
+      final ProcessorSupplier<KIn, VIn, KOut, VOut> processorSupplier
+  ) {
+    return AsyncProcessorSupplier.createAsyncProcessorSupplier(processorSupplier);
```
ableegoldman, line 27: I'm wondering if we really want to wrap every processor as async -- WDYT about only wrapping stateful operators by default?

(We can of course still allow users to inject their own custom async processors if they have some heavier processing without state like an RPC)

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/ResponsiveGlobalConsumer.java
```diff
@@ -182,7 +183,8 @@ static SingletonConsumerRecords of(final ConsumerRecords<byte[], byte[]> records
     public SingletonConsumerRecords(
         final Map<TopicPartition, List<ConsumerRecord<byte[], byte[]>>> records
     ) {
-      super(records);
+      super(records, Collections.emptyMap());
+      // TODO(sophie): need to pass in the actual next offsets here?
```
ableegoldman, line 187: Related to the upgrade -- a recent KIP added offset tracking to the consumer records, haven't looked into it much yet but I suspect Streams may be using this in some way and so we might need to actually implement this

kafka-client/src/main/java/org/apache/kafka/streams/state/internals/DelayedAsyncStoreBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.processor.StateStore;
+import org.apache.kafka.streams.processor.internals.StoreFactory.FactoryWrappingStoreBuilder;
+import org.apache.kafka.streams.state.KeyValueBytesStoreSupplier;
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.StoreBuilder;
+import org.apache.kafka.streams.state.TimestampedBytesStore;
+
+public class DelayedAsyncStoreBuilder<K, V, T extends StateStore>
```
ableegoldman, line 26: this class is basically where all the nasty hacky stuff is confined to. For now there are two main culprits:

1. checking and unwrapping the `FactoryWrappingStoreBuilder`, an internal AK class I had to add while converting DSL operators to use #stores, so that the DSLStoreSuppliers are resolved correctly
2. some "light" reflection to grab the storeSupplier from a storeBuilder instance

kafka-client/src/main/java/dev/responsive/kafka/api/async/AsyncProcessorWrapper.java
agavra: I was thinking we'd leverage `configure()` to allow us to plug into exactly this consideration. Not sure how we want to expose that as an API, but we can figure that out later. For now we can do just stateful ones. Like you said, if it's something stateless they want to make async it's pretty easy to write a processor for it manually.

PR #393, opened 2024-11-19
use cluster default for compaction strategy

agavra:
There was a regression in #371 where we started specifying LCS as the default compaction strategy. This change uses the cluster default instead, so that it will use ICS in scylla enterprise clusters (which ScyllaDB cloud are by default).

We can't test this easily since the scyllaDB public docker images don't have ICS, but I verified manually that this works.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/spec/DefaultTableSpec.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/spec/DefaultTableSpec.java
@@ -13,8 +13,6 @@
 package dev.responsive.kafka.internal.db.spec;
 
 import com.datastax.oss.driver.api.querybuilder.schema.CreateTableWithOptions;
-import com.datastax.oss.driver.api.querybuilder.schema.compaction.CompactionStrategy;
-import com.datastax.oss.driver.internal.querybuilder.schema.compaction.DefaultLeveledCompactionStrategy;
 import dev.responsive.kafka.internal.db.partitioning.TablePartitioner;
 import dev.responsive.kafka.internal.stores.TtlResolver;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/spec/DefaultTableSpec.java
```diff
@@ -13,17 +13,12 @@
...
-import com.datastax.oss.driver.api.querybuilder.schema.compaction.CompactionStrategy;
-import com.datastax.oss.driver.internal.querybuilder.schema.compaction.DefaultLeveledCompactionStrategy;
 import dev.responsive.kafka.internal.db.partitioning.TablePartitioner;
 import dev.responsive.kafka.internal.stores.TtlResolver;
 import java.util.Optional;
 
 public class DefaultTableSpec implements RemoteTableSpec {
 
-  private static final CompactionStrategy<?> DEFAULT_CASSANDRA_COMPACTION_STRATEGY =
-      new DefaultLeveledCompactionStrategy();
```
ableegoldman, line 25: The reason I chose this was because I saw it was used as the default in a few places (maybe not everywhere like I thought) -- we should make sure we keep this in those places right?

ableegoldman: Sadly I don't remember off the top of my head but I think if you do a search for the DefaultLeveledCompactionStrategy in my first TTL PR it should show where it used to be used

ableegoldman: (lmk if you want my help in finding that)

agavra: I think ICS is generally a good general starting point, and since we don't have prod deployments of other compaction strategies I'm keen to just keep the default.

ableegoldman (approved):

PR #394, opened 2024-11-19
API: fill in missing Mongo Window range queries

ableegoldman:
Fills in some of the missing range queries we had left out of the Mongo implementation due to performance concerns. Also adds a compound index on the windowed key which should alleviate the concerns for all but one of the range query types. For this last type, we log a warning and include a suggestion for flipping a config to trade off between different range query's performance


agavra (approved):
LGTM, though some sort of testing to make sure it's at least functional would be good 🌷

PR #396, opened 2024-11-22
Fix for Antithesis regression test

ableegoldman:
Played myself here by making the #1 most common mistake for Kafka Streams users: not realizing when/where you need to explicitly set serdes


ableegoldman:
Just gonna merge this right right away since it's a simple one-liner

Might be nice to have an integration test or something that runs regularly (or is Disabled but can be run when needed) so we can debug/avoid breaking the Antithesis stuff. Not a huge priority right now though 🙃

PR #398, opened 2024-12-06
fix async ttd code and add test

agavra:


kafka-client/src/main/java/dev/responsive/kafka/api/async/internals/AsyncThreadPool.java
```diff
@@ -182,6 +182,7 @@ public <KOut, VOut> void scheduleForProcessing(
         = inFlight.computeIfAbsent(asyncProcessorId, k -> new ConcurrentHashMap<>());
 
     for (final AsyncEvent event : events) {
+      log.info("Scheduled event {}", event.inputRecord());
```
ableegoldman, line 185: this really should be at DEBUG at the very least...

PR #401, opened 2024-12-10
Replace metrics compatibility mode with StorageBackend.NONE

ableegoldman:
Adds a new StorageBackend option of `NONE` for using Responsive SDK without having to configure Responsive storage. Useful for running things like async processing.

This replaces the old metrics compatibility mode, so we remove that config in favor of `StorageBackend.NONE`


kafka-client/src/test/java/dev/responsive/kafka/api/ResponsiveKafkaStreamsTest.java
```diff
@@ -154,8 +155,8 @@ public void shouldInvalidateBadConfigs() {
     builder.stream("foo").to("bar");
 
     // Then:
-    final StreamsException e = assertThrows(
-        StreamsException.class,
+    final ConfigException e = assertThrows(
```
ableegoldman, line 158: I intentionally removed an outer layer of wrapping ConfigException in a StreamsException since it didn't really add any additional info and just obfuscated the actual error

kafka-client/src/test/java/dev/responsive/kafka/async/AsyncProcessorIntegrationTest.java
```diff
@@ -261,7 +260,7 @@ public void shouldExecuteMultipleMixedAsyncProcessorsNoCaching() throws Exceptio
                       sleepForMs(1L);
                       return new SimpleProcessorOutput<>(val.getValue() + "-S1:" + newSum, newSum);
                     },
-                    ResponsiveKeyValueParams.fact(asyncStore1),
+                    ResponsiveStores.factStore(asyncStore1),
```
ableegoldman, line 263: I ran a full end-to-end test to make sure async works without a Responsive storage backend by plugging in `StorageBackend.NONE` and flipping all the stores from a Responsive store supplier to a regular Streams store supplier.

Ultimately I didn't feel it was worth parametrizing the test on this since these async tests have fairly long runtimes and there's not much value add, so I didn't commit this change. The problem is that the existing tests are all fairly heavy. I do think that once we can upgrade to 4.0 and get async working with the DSL, it could be worth running a short test with `NONE`, because I was going to write a large number of much lighter tests for the spread of DSL operators, so that would be a good opportunity for this as well

kafka-client/src/main/java/dev/responsive/kafka/api/ResponsiveKafkaStreams.java
```diff
@@ -526,29 +522,28 @@ public Params build() {
           streamsConfig,
           storeRegistry,
           metrics,
-          compatibilityMode
+          storageBackend
       );
 
       final var admin = responsiveKafkaClientSupplier.getAdmin(responsiveConfig.originals());
-      if (compatibilityMode == CompatibilityMode.METRICS_ONLY) {
```
agavra, line 533: we should leave this config for now and just have it alias to storage.backend=NONE since some customers and POCs use this

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
```diff
@@ -141,15 +142,18 @@ public static <K, V> StoreBuilder<KeyValueStore<K, V>> keyValueStoreBuilder(
    * @param storeSupplier the key-value store supplier
    * @param keySerde the key serde. If null, the default.key.serde config will be used
    * @param valueSerde the value serde. If null, the default.value.serde config will be used
-   * @return a store builder that can be used to build a key-value store with the given options
-   *         that uses Responsive's storage for its backend
+   * @return a store builder that can be used to build a key-value store that is compatible
+   *         with the Responsive framework
```
agavra, line 146: what's the behavior if they specify a non-NONE storage.backend and they pass in a non-responsive store supplier?

agavra (approved):
LGTM, note on backwards compatibility

kafka-client/src/main/java/dev/responsive/kafka/api/stores/ResponsiveStores.java
ableegoldman: this is, and should be, allowed -- for example someone might want to use async for all their stateful apps, but have one Responsive store and one rocksdb (or other  non-responsive) store.

Note that at the moment, you need to use the ResponsiveStoreBuilder for async, but not necessarily a Responsive store. Hence this change.

But I'm working on removing this requirement (in fact I already have in my draft PR for KIP-1112). I'll probably scrape that part of the draft PR out into a separate PR so we can get rid of this requirement without waiting for 4.0.

ableegoldman:
Ran the async integration tests with the old CompatibilityMode.METRICS_ONLY and took out all the Responsive stores in one of them and it passed, so I think we're good here (and the ones that kept the Responsive stores failed, as expected)

Will merge when the tests pass

PR #409, opened 2025-01-13
support configuring tombstone retention in Mongo

agavra:
Note that we still need to fix https://github.com/responsivedev/responsive-pub/issues/410


kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -111,6 +111,15 @@ public class ResponsiveConfig extends AbstractConfig {
       + "with for best results. However it is important to note that this cannot be changed for "
       + "an active application. Messing with this can corrupt existing state!";
 
+  public static final String MONGO_TOMBSTONE_RETENTION_MS_CONFIG = "responsive.mongo.tombstone.retention.ms";
```
rodesai, line 114: bit of a nit that I'll leave up to you but seconds or minutes is probably a more friendly unit for this config.

rodesai (approved):
LGTM. One suggestion inline

PR #411, opened 2025-01-13
support prefix scans in ResponsiveKeyValueStore

### LINKED ISSUE #408: Support prefixScan on ResponsiveKeyValueStore
Today we support `range` on `ResponsiveKeyValueStore` but we neglected to implement `prefixScan` (which can be implemented similarly except setting to/from to be one byte incremented).

agavra:
fixes #408 

Support prefix scans on:
- `CassandraKeyValueTable`
- `MongoKeyValueTable`
- `InMemoryKeyValueTable`

We don't support range scans on Global tables (yet), `CassandraFactTable` or `RS3KeyValueTable` yet.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
@@ -38,6 +38,7 @@ import dev.responsive.kafka.internal.db.partitioning.SubPartitioner;
 import dev.responsive.kafka.internal.db.spec.RemoteTableSpec;
 import dev.responsive.kafka.internal.stores.TtlResolver;
 import dev.responsive.kafka.internal.utils.Iterators;
+import dev.responsive.kafka.internal.utils.Utils;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.time.Instant;
@@ -46,6 +47,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.TimeoutException;
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.errors.TaskMigratedException;
@@ -74,6 +76,7 @@ public class CassandraKeyValueTable implements RemoteKVTable<BoundStatement> {
   private final PreparedStatement fetchEpoch;
   private final PreparedStatement reserveEpoch;
   private final PreparedStatement ensureEpoch;
+  private final PreparedStatement prefix;
 
   public static CassandraKeyValueTable create(
       final RemoteTableSpec spec,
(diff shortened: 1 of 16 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -383,6 +402,27 @@ public KeyValueIterator<Bytes, byte[]> range(
...
+  @Override
+  public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefix(
+      final P prefix,
+      final PS prefixKeySerializer,
+      final int kafkaPartition,
+      final long streamTimeMs
+  ) {
+    final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+    final Bytes to = Utils.incrementWithoutOverflow(from);
+    return doRange(kafkaPartition, from, to, streamTimeMs, this.prefix);
```
ableegoldman, line 417: very nitpicky but why use `range` above but `this.prefix` here?  Can only afford one `this` ? 😛

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
```diff
@@ -383,6 +402,27 @@ public KeyValueIterator<Bytes, byte[]> range(
...
+  }
+
+  private KeyValueIterator<Bytes, byte[]> doRange(
+      final int kafkaPartition,
+      final Bytes from,
+      final Bytes to,
+      final long streamTimeMs,
+      final PreparedStatement rangeStatement) {
```
ableegoldman, line 425:
```suggestion
      final PreparedStatement rangeStatement
    ) {
```

kafka-client/src/test/java/dev/responsive/kafka/internal/db/CassandraKVTableIntegrationTest.java
```diff
@@ -196,6 +197,40 @@ public void shouldReturnRangeKeysInLexicalOrderAcrossMultipleSubPartitions() {
     }
   }
 
+  @Test
```
ableegoldman, line 200: Can we add a test that hits the overflow case? Would be good to have one for each table type to make sure the underlying engine does the right thing

kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraKeyValueTable.java
agavra: 😆 it's because this has `prefix` in the parameters so there's a scoping conflict whereas range doesn't have that. I'll use `this` for both for symmetry if you demand it haha

kafka-client/src/main/java/dev/responsive/kafka/internal/utils/Utils.java
```diff
@@ -117,4 +118,12 @@ public static String extractThreadIdFromThreadName(final String threadName) {
     LOG.warn("Unable to parse the stream thread id, falling back to thread name {}", threadName);
     return threadName;
   }
+
+  public static Bytes incrementWithoutOverflow(final Bytes input) {
+    try {
+      return Bytes.increment(input);
+    } catch (final IndexOutOfBoundsException e) {
+      return null;
```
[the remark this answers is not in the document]
agavra: Good call, fixed bugs and added tests.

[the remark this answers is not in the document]
agavra: I fixed it for Mongo/In Memory but it's a bit more work for C* so I'll just create a ticket for that now.

PR #412, opened 2025-01-14
Use Date format for TTL fields in Mongo tables

### LINKED ISSUE #410: MongoDB TTL Does Not Work
**Describe the bug**
MongoDB TTL indexes require a Date type field

**To Reproduce**

```java
  @Test
  @Disabled("fix this when we fix ")
  public void shouldExpireRecords() {
    props.put(MONGO_TOMBSTONE_RETENTION_MS_CONFIG, 1000);
    final ResponsiveConfig config = ResponsiveConfig.responsiveConfig(props);
    final Duration ttl = Duration.ofMillis(100);
    final MongoKVTable table = new MongoKVTable(
        client, name, UNSHARDED, defaultOnlyTtl(ttl), config
    );

    var writerFactory = table.init(0);
    var writer = writerFactory.createWriter(0, 0);
    writer.insert(bytes(10, 11, 12, 13), byteArray(2), 100);
    writer.flush();

    // When:
    final BooleanSupplier isExpired = () -> {
      final byte[] bytes = table.get(0, bytes(10, 11, 12, 13), 100);
      System.out.println(bytes == null);
      return bytes == null;
    };

    // Then:
    IntegrationTestUtils.awaitCondition(isExpired, Duration.ofSeconds(30), Duration.ofSeconds(1));
  }
```

You can check out the following if you set a debugger and connect via `mongosh` 

```
db.kv_data.getIndexes();
[
  { v: 2, key: { _id: 1 }, name: '_id_' },
  {
    v: 2,
    key: { tombstoneTs: -1 },
    name: 'tombstoneTs_-1',
    expireAfterSeconds: 1
  },
  { v: 2, key: { ts: -1 }, name: 'ts_-1', expireAfterSeconds: 1 }
]
```

The following record won't be expired:
```
  {
    _id: '0UgA1E',
    epoch: Long("1"),
    partition: 0,
    ts: Long("100"),
    value: Binary.createFromBase64("Ag==", 0)
  }
```

But this one does
```
db.kv_data.insertOne({ _id: 'test', ts: new Date(Date.now() - 1000) });
```

**Expected behavior**
The record above should be expired. This happens because MongoDB requires `Date` fields for the TTL to properly work.

ableegoldman:
fixes #410 

Need to use Date instead of long for ttl to work, apparently

Since java just stores `Date` as an "epoch millis" long the only overhead is in creating the objects themselves, so it should be pretty minimal

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
@@ -301,7 +301,7 @@ public class CassandraFactTable implements RemoteKVTable<BoundStatement> {
       final int kafkaPartition,
       final Bytes key,
       final byte[] value,
-      final long epochMillis
+      final long timestampMs
   ) {
     if (ttlResolver.isPresent()) {
       final Optional<TtlDuration> rowTtl = ttlResolver.get().computeTtl(key, value);
@@ -319,7 +319,7 @@ public class CassandraFactTable implements RemoteKVTable<BoundStatement> {
             .bind()
             .setByteBuffer(DATA_KEY.bind(), ByteBuffer.wrap(key.get()))
             .setByteBuffer(DATA_VALUE.bind(), ByteBuffer.wrap(value))
-            .setInstant(TIMESTAMP.bind(), Instant.ofEpochMilli(epochMillis))
+            .setInstant(TIMESTAMP.bind(), Instant.ofEpochMilli(timestampMs))
             .setInt(TTL_SECONDS.bind(), rowTtlOverrideSeconds);
       }
     }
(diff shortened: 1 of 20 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/CassandraFactTable.java
```diff
@@ -301,7 +301,7 @@ public BoundStatement insert(
       final int kafkaPartition,
       final Bytes key,
       final byte[] value,
-      final long epochMillis
+      final long timestampMs
```
ableegoldman, line 304: changed the name of this field just for future clarity (and to differentiate from the other kind of "epoch" ie the field in our stores)

agavra (approved):
Approved pending test changes :)

kafka-client/src/test/java/dev/responsive/kafka/async/AsyncProcessorIntegrationTest.java
```diff
@@ -106,6 +106,7 @@
 public class AsyncProcessorIntegrationTest {
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+  // TODO: consider running a mix of mongo and cassandra for the async tests
```
ableegoldman, line 109: I did run these as Mongo for another check, but we should leave it as Cassandra for the CI/CD for now (or until we address this TODO)

PR #418, opened 2025-02-03
RS3: Add static logical store mapping config

hachikuji:
Add a new config `responsive.rs3.logical.store.mapping` which stores a static mapping from the table name to the RS3 logical store ID in the form `[t1]:[id1],[t2]:[id2]`.


rodesai (approved):
LGTM!

PR #420, opened 2025-02-10
Add retry logic for RS3 get/getoffsets requests

hachikuji:
Get requests are probably the simplest to retry, so I've started there. This patch adds a new `responsive.rs3.retry.timeout.ms` and uses exponential backoff logic (borrowed from Kafka) to retry after UNAVAILABLE errors from the rs3 server.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3Client.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3Client.java
@@ -81,24 +81,28 @@ public class GrpcRS3Client implements RS3Client {
   private <T> T withRetry(Supplier<T> supplier, Supplier<String> opDescription) {
     // Using Kafka default backoff settings initially. We can pull them up
     // if there is ever strong reason.
-    final ExponentialBackoff backoff = new ExponentialBackoff(50, 2, 1000, 0.2);
-    final long startTimeMs = time.milliseconds();
+    final var backoff = new ExponentialBackoff(50, 2, 1000, 0.2);
+    final var startTimeMs = time.milliseconds();
+    final var deadlineMs = startTimeMs + retryTimeoutMs;
 
-    long retries = 0;
+    var retries = 0;
     long currentTimeMs;
 
     do {
       try {
         return supplier.get();
       } catch (final StatusRuntimeException e) {
-        if (e.getStatus() == UNAVAILABLE) {
-          retries += 1;
-          time.sleep(backoff.backoff(retries));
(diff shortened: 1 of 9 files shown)


kafka-client/src/test/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3ClientTest.java
```diff
@@ -148,6 +158,69 @@ public void shouldGetOffsets() {
...
+    final RS3Exception exception =
+        assertThrows(RS3Exception.class, () -> client.getCurrentOffsets(STORE_ID, LSS_ID, PSS_ID));
+
+    // then:
+    assertThat(exception.getCause(), is(instanceOf(StatusRuntimeException.class)));
+    assertThat(((StatusRuntimeException) exception.getCause()).getStatus(), is(Status.UNKNOWN));
+  }
+
+  @Test
+  public void shouldTimeoutGetOffsets() throws Exception {
```
rodesai, line 197: Can we implement this test by advancing time in `MockTime` on every call to `getOffsets`. That should make it faster.

rodesai (commented):
Overall the approach LGTM. Couple of minor comments inline.

kafka-client/src/test/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3ClientTest.java
hachikuji: Yeah, that's a nice suggestion.

hachikuji: In fact, we don't need to sleep at all from the test. The backoff will advance time for us.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3Client.java
```diff
@@ -210,4 +235,41 @@ private void checkField(final Supplier<Boolean> check, final String field) {
       throw new RuntimeException("rs3 resp proto missing field " + field);
     }
   }
+
+  public static class Connector {
```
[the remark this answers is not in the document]
hachikuji: Added mainly because there are more parameters now. The static `connect` worked nicely when there were only two parameters, but it started feeling a little clumsy with the additions from the patch. Having a place to hook in mocks could be useful as well I guess.

rodesai (approved):
LGTM!

PR #421, opened 2025-02-11
Fix spammy antithesis setup complete event

hachikuji:
Currently startup events are spamming antithesis logs. We had a field `setupCompleteSignalFired` which was enabled to `false`, but never set to `true` after setup completed.


rodesai (approved):
LGTM!

PR #423, opened 2025-02-14
Fix incorrect type for rs3 retry timeout

hachikuji:
We defined the type of the rs3 retry timeout as long, but used `getInt` when accessing the property. This causes a `ClassCastException` which fails startup.


hachikuji:
A basic integration test would have caught this. I'll try to add something tomorrow.

agavra (approved):
🫠

PR #424, opened 2025-02-17
Introduce origin event tracking mechanism

agavra:
This patch has a few changes:

- It introduces the mechanism for counting origin events via `OriginEventRecorder` that inserts headers as records are consumed if (and only if) they do not already contain the origin event marker
- It adds a `UsageBasedV1` License Info which acts as a placeholder for our Usage based license when we introduce it. This is necessary so that existing licenses don't suddenly start reporting Origin Events / adding the headers
- It adds the reporting of Origin Events to a server, a `TestLicenseServer` that imitates a license server and integration tests to make sure that it works with complicated subtopologies

These are done in three separate commits if you want to review one at a time.


kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+
+  @Override
+  public <K, V> ConsumerRecords<K, V> onPoll(final ConsumerRecords<K, V> records) {
+    for (final ConsumerRecord<K, V> record : records) {
+      final var header = record.headers().lastHeader(ORIGIN_EVENT_HEADER_KEY);
+      if (header == null) {
+        record.headers().add(ORIGIN_EVENT_HEADER_KEY, ORIGIN_EVENT_MARK);
+        inc(new TopicPartition(record.topic(), record.partition()));
+      }
+    }
```
agavra, line 100: this is _the_ main part of the PR where it records a new origin event and marks it as an origin event

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+                LOG.debug("Successfully reported usage (status: {}): {}", status, responseBody);
+              }
+            })
+            .exceptionally(e -> {
+              LOG.error("Failed to report usage! Please make sure "
+                  + "you have correctly configured your Responsive License", e);
+              return null;
+            });
+      } catch (final JsonProcessingException | URISyntaxException e) {
+        // this should never happen
```
rodesai, line 205: we should probably just log in this case right? We don't want to fail the app if there's some bug here

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+  private synchronized void inc(final TopicPartition tp) {
+    originEvents.compute(tp, (k, v) -> v == null ? 1 : v + 1);
+  }
+
+  @Override
+  public <K, V> ConsumerRecords<K, V> onPoll(final ConsumerRecords<K, V> records) {
+    for (final ConsumerRecord<K, V> record : records) {
+      final var header = record.headers().lastHeader(ORIGIN_EVENT_HEADER_KEY);
+      if (header == null) {
+        record.headers().add(ORIGIN_EVENT_HEADER_KEY, ORIGIN_EVENT_MARK);
```
rodesai, line 97: do these headers actually get propagated down to the producer when it writes? I wonder if it would make more sense to just add the headers from the producer itself to prevent the header from being removed/messed with by streams

rodesai: On a related note, don't we need to filter out changelog records? Though I suppose they are only read by the restore consumer.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+        inc(new TopicPartition(record.topic(), record.partition()));
+      }
+    }
+    return records;
+  }
+
+  @Override
+  public void onCommit(final Map<TopicPartition, OffsetAndMetadata> offsets) {
+    final var now = System.currentTimeMillis();
```
rodesai, line 106: The offsets being committed aren't necessarily what's been polled so far. Streams puts the polled records into its own buffer, and can commit at any time, even if the internal buffer isn't drained. So technically we should only be reporting origin events up to the offsets being committed. How to do that is tricky. Need to think about it some more.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+      if (header == null) {
+        record.headers().add(ORIGIN_EVENT_HEADER_KEY, ORIGIN_EVENT_MARK);
+        inc(new TopicPartition(record.topic(), record.partition()));
+      }
+    }
+    return records;
+  }
+
+  @Override
+  public void onCommit(final Map<TopicPartition, OffsetAndMetadata> offsets) {
```
rodesai, line 105: don't we need a fn like this producer-side for EOS?

kafka-client/src/main/java/dev/responsive/kafka/api/config/ResponsiveConfig.java
```diff
@@ -307,15 +312,20 @@ public class ResponsiveConfig extends AbstractConfig {
...
+  private static final String RESTORE_OFFSET_REPAIR_ENABLED_DOC = "When set to 'true', " + RESTORE_OFFSET_REPAIR_ENABLED_CONFIG
       + " will ignore OffsetOutOfRangeException and instead seek to the earliest available offset. This exception "
       + "should only happen in situations where there is truncation/retention on the changelog topic and restoring from the latest "
       + "committed offset in the remote store is no longer possible. Note that in some situations this may cause data "
       + "loss, use this configuration with caution";
 
+  public static final String ORIGIN_EVENT_REPORT_INTERVAL_MS_CONFIG = "responsive.origin.event.report.interval.ms";
+  private static final long ORIGIN_EVENT_REPORT_INTERVAL_MS_DEFAULT = 10_000L;
+  private static final String ORIGIN_EVENT_REPORT_INTERVAL_MS_DOC =
+      "How often to report origin event usage information. This shoudl generally not be changed in production environments";
```
rodesai, line 327:
```suggestion
      "How often to report origin event usage information. This should generally not be changed in production environments";
```

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
agavra: I split the two - if there's a URI syntax error then it's a problem with the configuration which users should change.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
agavra: oops! good call, I'll add tests for EOS as well

kafka-client/src/main/java/dev/responsive/kafka/internal/license/LicenseChecker.java
```diff
@@ -15,12 +15,21 @@
...
 public class LicenseChecker {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LicenseChecker.class);
+
   public void checkLicense(final LicenseInfo licenseInfo) {
     if (licenseInfo instanceof TimedTrialV1) {
       verifyTimedTrialV1((TimedTrialV1) licenseInfo);
+      LOG.info("Checked and confirmed valid Time Trial license");
+    } else if (licenseInfo instanceof UsageBasedV1) {
+      LOG.info("Checked and confirmed valid Usage Based license");
```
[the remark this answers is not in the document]
agavra: yeah, this is where we could issue the remote check if we wanted to

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
agavra: OK I fixed this and I believe the previous one as well by tracking origin events in a bitset instead of just as a long, and then counting that bitset on commit. It's a bit heavyweight, but I think it should be fine since if we are tracking 100K offsets per commit we'll only need ~1.5k longs which should be hardly any memory overhead and 1.5k bitwise ops for counting the number of 1s shouldn't take much either. PTAL

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
```diff
@@ -0,0 +1,210 @@
...
+    originEvents.compute(tp, (k, v) -> v == null ? 1 : v + 1);
+  }
+
+  @Override
+  public <K, V> ConsumerRecords<K, V> onPoll(final ConsumerRecords<K, V> records) {
+    for (final ConsumerRecord<K, V> record : records) {
+      final var header = record.headers().lastHeader(ORIGIN_EVENT_HEADER_KEY);
+      if (header == null) {
+        record.headers().add(ORIGIN_EVENT_HEADER_KEY, ORIGIN_EVENT_MARK);
+        inc(new TopicPartition(record.topic(), record.partition()));
```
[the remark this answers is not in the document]
agavra: This is fixed with the new offset tracker

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/NoopOriginEventRecorder.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2024 Responsive Computing, Inc.
```
ableegoldman, line 2: 2025 now 😉

kafka-client/src/test/resources/log4j.properties
```diff
@@ -17,5 +17,6 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c:%L)%n
 
 log4j.logger.dev.responsive.kafka.api.async=DEBUG
+#log4j.logger.dev.responsive.kafka.internal.clients.OriginEventRecorder=DEBUG
```
ableegoldman, line 20: did you mean to include this?

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OffsetTracker.java
```diff
@@ -0,0 +1,64 @@
...
+  private BitSet offsets = new BitSet();
+  private long baseOffset;
+
+  public OffsetTracker(long baseOffset) {
+    this.baseOffset = baseOffset;
+  }
+
+  public void mark(final long offset) {
+    if (offset < baseOffset) {
+      throw new IllegalArgumentException(
```
ableegoldman, line 34: nit: can we log an error before throwing that includes the value of `offsets` (here and below)

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OffsetTracker.java
```diff
@@ -0,0 +1,64 @@
...
+
+package dev.responsive.kafka.internal.clients;
+
+import java.util.BitSet;
+
+/**
+ * This class allows us to efficiently count the number of events
+ * between two offsets that match a certain condition. This is somewhat
+ * memory efficient in that we can track 100K offsets with ~1.5K longs
+ * (64 bits per long), or roughly 12KB.
```
ableegoldman, line 21: I'm a little confused by the point of this class. This is just counting the number of origin events since the last commit, right? It doesn't seem like we ever expose what the actual origin event offsets are, so why do we need to save them in this `BitSet` scheme -- ie why not just use an int that increments in `#mark` and resets/returns in `#countAndShift`?

Lmk if you'd rather sync online because I'm sure I'm missing something here 🙂

agavra: This class should definitely have javadoc explaining it, originally I had implemented what you're saying but Rohan pointed out these two concerns: https://github.com/responsivedev/responsive-pub/pull/424#discussion_r1965094537 and https://github.com/responsivedev/responsive-pub/pull/424#discussion_r1965097847 both fixed by this bitset tracker.

kafka-client/src/test/resources/log4j.properties
agavra: Yeah I did, I was using it earlier and figured I'd probably be turning it off and on. I can remove it though if it bothers our perfectionism 😆

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OffsetTracker.java
```diff
@@ -0,0 +1,81 @@
...
+      throw new IllegalArgumentException(
+          "Commit offset " + commitOffset + " cannot be less than baseOffset " + baseOffset);
+    }
+
+    final int shift = Math.toIntExact(commitOffset - baseOffset);
+    final int count = offsets.get(0, shift).cardinality();
+
+    // shift the bitset so that commitOffset is the new base offset
+    // for which we track marked entries
+    final var old = this.offsets;
```
rodesai, line 72: can we just do `this.offsets.get(shift, this.offsets.length())`? That will be more efficient too since internally it copies over whole longs rather than going bit-by-bit.

kafka-client/src/test/java/dev/responsive/kafka/integration/OriginEventIntegrationTest.java
```diff
@@ -0,0 +1,413 @@
...
+    final KafkaProducer<String, String> producer = new KafkaProducer<>(props);
+    try (final var kafkaStreams = new ResponsiveKafkaStreams(builder.build(), props)) {
+      startAppAndAwaitRunning(Duration.ofSeconds(15), kafkaStreams);
+      pipeTimestampedRecords(producer, inputTopic(), inputs);
+      assertThat(latch.await(30, TimeUnit.SECONDS), Matchers.is(true));
+    }
+
+    // Then:
+    final String appId = (String) props.get(StreamsConfig.APPLICATION_ID_CONFIG);
+    IntegrationTestUtils.awaitCondition(
```
rodesai, line 143: why do we need to do this in `awaitCondition`? The app is shut down so whatever counts that the license server has should be reflected at this point right? That way we fail immediately rather than on a timeout if we misreport.

rodesai (commented):
Looks good! Would still be good to get @ableegoldman's feedback on whether the problems I raised that require the bitset make sense.

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OffsetTracker.java
agavra: nice 👍

rodesai (approved):
LGTM!

kafka-client/src/test/resources/log4j.properties
ableegoldman: nope I'm good with it, just wanted to make sure!

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OriginEventRecorder.java
ableegoldman: yeah this is a good catch, what's been polled =/= what's being committed. I think we could find other ways to work around this (for example by wrapping the `ConsumerRecord` class) but your bitwise solution seems efficient and simple enough.

Just please explain why it does what it does in the javadocs because without this context it's pretty hard to understand what it's there for lol

kafka-client/src/main/java/dev/responsive/kafka/internal/clients/OffsetTracker.java
ableegoldman: ah, I missed those discussions, but yeah he's right. thanks for filling in the javadocs

ableegoldman (approved):
LGTM!

PR #429, opened 2025-03-05
Add retry logic for RS3 WriteWalSegments

hachikuji:
Retries for WriteWalSegments are not as straightforward as some of the blocking calls because we are using an asynchronous stream. The simplest approach I could come up with was to buffer the writes as they are being written. If an UNAVAILABLE error occurs while writing to the stream, then we continue writing to the buffer, but stop writing to the stream.  Once all wal entries have been received, then the retry will be done using the synchronous `writeWalSegments` API.


kafka-client/src/test/java/dev/responsive/kafka/internal/db/rs3/RS3KVTableIntegrationTest.java
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.common.utils.MockTime;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+public class RS3KVTableIntegrationTest {
```
hachikuji, line 50: This shows up as an addition, but it is a simple rename. I thought `IntegrationTest` seemed more appropriate since we are using the RS3 container. I wanted `RS3KVTableTest` to allow mocking dependencies. If the rename seems reasonable, I can submit a separate PR so that it doesn't show up in the diff.

rodesai (commented):
Thanks! My main question is how we actually get errors returned back to us when doing a streaming rpc. Is it always through the response callback? Or can the stream sender throw exceptions directly. If it's the latter, then we wouldn't be handling those exceptions.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -156,64 +160,140 @@ public CompletionStage<RemoteWriteResult<Integer>> flush() {
...
+      maybeSendNext(table.delete(kafkaPartition, key));
+    }
+
+    private void maybeSendNext(WalEntry entry) {
+      retryBuffer.add(entry);
+
+      if (isStreamActive) {
+        try {
+          streamSender.sendNext(entry);
+        } catch (IllegalStateException e) {
```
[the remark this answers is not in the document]
hachikuji: Yeah, I think you are right. I let openai mislead me here since it suggested that `onError` could be invoked internally by grpc. If that were the case, then subsequent calls would fail with `IllegalStateException`. Instead it looks like `onNext` should throw the status exception and the caller is expected to either retry `onNext` or invoke `onError` to signal that it is giving up on the stream. I will revise the patch.

kafka-client/src/test/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcRS3ClientEndToEndTest.java
```diff
@@ -0,0 +1,249 @@
...
+        Optional.of(5L),
+        key
+    );
+    assertThat(getResult.isPresent(), is(true));
+    final var resultValue = getResult.get();
+    assertThat(resultValue, equalTo(value));
+  }
+
+  @Test
+  public void shouldRetryPutWithNetworkInterruption() {
```
hachikuji, line 106: These tests get us a little closer to simulating bad network behavior, but still rely on assumptions about how network errors are raised in the grpc stack.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -156,64 +161,131 @@ public CompletionStage<RemoteWriteResult<Integer>> flush() {
...
-            LOG.debug("last flushed offset for pss/lss {}/{} is {}", pssId, lssId, flushedOffset);
-            return RemoteWriteResult.success(kafkaPartition);
-          }
-      );
+      ifActiveStream(StreamSender::finish);
+
+      return sendRecv.receiver().handle((result, throwable) -> {
+        Optional<Long> flushedOffset = result;
+        if (throwable instanceof RS3TransientException) {
+          flushedOffset = streamFactory.writeWalSegmentSync(retryBuffer);
```
rodesai, line 277: the retries for this call happen in the grpc client right?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcStreamSender.java
```diff
@@ -30,16 +32,34 @@ class GrpcStreamSender<M, P> implements StreamSender<M> {
...
   }
 
   @Override
   public void finish() {
-    grpcObserver.onCompleted();
+    try {
+      grpcObserver.onCompleted();
+    } catch (Exception e) {
+      grpcObserver.onError(e);
+      active.set(false);
```
rodesai, line 50: should we be setting `active` to `false` whether or not there's an exception?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -156,64 +161,131 @@ public CompletionStage<RemoteWriteResult<Integer>> flush() {
...
     @Override
     public CompletionStage<RemoteWriteResult<Integer>> flush() {
-      streamSender.finish();
-      return resultFuture.thenApply(
-          flushedOffset -> {
-            LOG.debug("last flushed offset for pss/lss {}/{} is {}", pssId, lssId, flushedOffset);
-            return RemoteWriteResult.success(kafkaPartition);
-          }
-      );
+      ifActiveStream(StreamSender::finish);
```
rodesai, line 272: we probably need to catch exceptions from `finish` as well

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
```diff
@@ -156,64 +161,131 @@ public CompletionStage<RemoteWriteResult<Integer>> flush() {
...
+
+      return sendRecv.receiver().handle((result, throwable) -> {
```
rodesai, line 274: If there's a failure when sending the stream, is it guaranteed to be propagated to the `GrpcMessageReceiver` and therefore result in `handle` being called with the error here? I can't find any docs that says that's the case. Therefore we probably want to fall back to the sync path if we ever hit any errors when sending the stream.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcStreamSender.java
```diff
@@ -30,16 +32,34 @@ class GrpcStreamSender<M, P> implements StreamSender<M> {
 
   @Override
   public void sendNext(M msg) {
-    grpcObserver.onNext(protoFactory.apply(msg));
+    try {
+      grpcObserver.onNext(protoFactory.apply(msg));
+    } catch (Exception e) {
+      grpcObserver.onError(e);
```
rodesai, line 38: related to my q below on how errors are surfaced, this `grpcObserver` is the grpc side of the client->server stream. Does that propagate the error to our side of the server->client stream?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
hachikuji: Yes, that's right. I was trying to keep retry logic encapsulated in the grpc client. But there's not really a way I could think of to encapsulate retries for the async API.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
hachikuji: In `ifActiveStream`, we catch `RS3TransientException`.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3KVFlushManager.java
hachikuji: I handled this in the latest commit by tracking a completion for both the send and recv sides. If either fails, then the combined completion fails as well.

kafka-client-examples/e2e-test/docker/antithesis/async/config/volumes/app-rs3.properties
```diff
@@ -19,3 +19,4 @@ responsive.rs3.hostname=rs3
 responsive.rs3.port=50051
 responsive.rs3.logical.store.mapping=e2e:b1a45157-e2f0-4698-be0e-5bf3a9b8e9d1
 responsive.rs3.tls.enabled=false
+responsive.rs3.retry.timeout.ms=1800000
```
rodesai, line 22: what default are we using? should probably be set to like half the max poll interval

rodesai (approved):
LGTM! The e2e test looks nice

kafka-client-examples/e2e-test/docker/antithesis/async/config/volumes/app-rs3.properties
hachikuji: I had it at 30s, which might be on the low side. Do we use the default poll interval of 5min?

hachikuji: I changed the default to 2 minutes. That is close to half the poll interval and matches the default producer delivery timeout.

PR #432, opened 2025-03-07
change OE reporting from per-thread to per-node

agavra:
The old code used to create a reporter per stream thread, now we're creating a reporter once globally. Whenever a stream thread commits it add the current counts of origin events to the reporter. The reporter is a scheduled service that just commits whatever it has every 60s.


rodesai (approved):
LGTM!

PR #435, opened 2025-03-21
Rs3 range query support

hachikuji:
Adds support for range queries over GRPC. We expose a new method `Rs3Client.range` which returns a blocking iterator. Internally in the iterator implementation, transient errors are caught and the operation is retried using the same exponential backoff we use for other operations.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcMessageQueue.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package dev.responsive.kafka.internal.db.rs3.client.grpc;
+
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class GrpcMessageQueue<T> {
+  private final ConcurrentLinkedQueue<T> queue = new ConcurrentLinkedQueue<>();
```
rodesai, line 20: why not just use a [BlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html) to implement this?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/RangeBound.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private Unbounded() {}
+
+    @Override
+    public <T> T map(final Mapper<T> mapper) {
+      return mapper.map(this);
+    }
+  }
+
+  interface Mapper<T> {
```
rodesai, line 113: nit: maybe call this a `Visitor` instead? This looks like the classic visitor pattern to me

rodesai (commented):
This looks great! My only question is about the queue implementation - why not just use the java BlockingQueue rather than rolling our own?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/grpc/GrpcMessageQueue.java
hachikuji: Probably a premature optimization. I think blocking queues require a lock on every queue access. I had hoped to avoid locking when the queue was not empty.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/client/RangeBound.java
hachikuji: Yeah, it's a visitor pattern, but I'm using it as a poor man's version of pattern matching to implement a `map` function. I thought using the `Visitor` name made its usage feel a little too vague. I don't feel strongly about it though..

rodesai: Up to you - I felt the opposite 😛 . I was a bit confused at first and had to look at how it was used but it would've felt obvious if it was named `Visitor`. But that could just be me. We had the Visitor pattern drilled into us working on ksql syntax trees 😄

PR #438, opened 2025-03-31
Separate segment window flush manager

hachikuji:
The current `WindowFlushManager` is coupled with the segmenting approach even though it is not really used in `SegmentedOperations`. This patch moves `WindowFlushManager`to `SegmentedWindowFlushManager` and then reuses the `WindowFlushManager` name for an interface sitting between the two. I've also renamed `SegmentedOperations` to `RemoteWindowOperations` to reflect more general usage.

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentedWindowFlushManager.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentedWindowFlushManager.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2024 Responsive Computing, Inc.
+ *
+ * This source code is licensed under the Responsive Business Source License Agreement v1.0
+ * available at:
+ *
+ * https://www.responsive.dev/legal/responsive-bsl-10
+ *
+ * This software requires a valid Commercial License Key for production use. Trial and commercial
+ * licenses can be obtained at https://www.responsive.dev
+ */
+
+package dev.responsive.kafka.internal.db;
+
+import dev.responsive.kafka.internal.db.partitioning.Segmenter;
+import dev.responsive.kafka.internal.db.partitioning.Segmenter.SegmentPartition;
+import dev.responsive.kafka.internal.stores.RemoteWriteResult;
+import dev.responsive.kafka.internal.utils.PendingFlushSegmentMetadata;
+import dev.responsive.kafka.internal.utils.WindowedKey;
+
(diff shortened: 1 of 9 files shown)


kafka-client/src/main/java/dev/responsive/kafka/internal/db/SegmentedWindowFlushManager.java
```diff
@@ -0,0 +1,113 @@
...
+
+package dev.responsive.kafka.internal.db;
+
+import dev.responsive.kafka.internal.db.partitioning.Segmenter;
+import dev.responsive.kafka.internal.db.partitioning.Segmenter.SegmentPartition;
+import dev.responsive.kafka.internal.stores.RemoteWriteResult;
+import dev.responsive.kafka.internal.utils.PendingFlushSegmentMetadata;
+import dev.responsive.kafka.internal.utils.WindowedKey;
+
+public abstract class SegmentedWindowFlushManager implements WindowFlushManager<SegmentPartition> {
```
hachikuji, line 21: Note this is just a rename. The only change is adding the `@Override` on `streamTime()`.

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/RemoteWindowOperations.java
```diff
@@ -53,7 +53,7 @@
 import org.apache.kafka.streams.state.WindowStoreIterator;
 import org.slf4j.Logger;
 
-public class SegmentedOperations implements WindowOperations {
+public class RemoteWindowOperations implements WindowOperations {
```
ableegoldman, line 56: nit: WDYT about calling this `TtlWindowOperations` since expiry is based on ttl for this, as opposed to the SegmentedWindowOperations where we expire in segments

Or even just "UnsegmentedWindowOperations" -- my point is that it feels a bit weird to call the two variants "Segmented" and "remote", just makes it seem like the Segmented variant isn't remote. Just a thought, I'll leave it up to you

hachikuji: Hmm, I guess I thought that segmented operations were mostly encapsulated by the flush manager. The only references to segments in this class is in the factory logic for the respective stores. So I think we can use this class for RS3 as well, just with different factory logic (perhaps we could even factor out this logic, but I was punting on that for now). So `RemoteWindowOperations` would be used for both segmented and TTL based expiration. Does that make sense?

ableegoldman (approved):
LGTM

PR #440, opened 2025-04-03
CreateStore API for RS3

ableegoldman:
blocked on https://github.com/responsivedev/rs3/pull/57


kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -14,38 +14,79 @@
...
+
+    if (!createdStores.contains(storeName)) {
+      final int kafkaPartitions = computeNumKafkaPartitions.get();
+
+      final Optional<Long> defaultTtl = ttlResolver.isPresent() && ttlResolver.get().defaultTtl().isFinite()
+          ? Optional.of(ttlResolver.get().defaultTtl().duration().toMillis())
+          : Optional.empty();
+
+      final var options = new CreateStoreOptions(
+          ttlResolver.isPresent() ? Optional.of(ClockType.WALL_CLOCK) : Optional.empty(),
```
ableegoldman, line 72: just hardcode WALL_CLOCK until we have support for STREAM_TIME

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -14,38 +14,79 @@
...
+          defaultTtl,
+          Optional.empty()
```
ableegoldman, line 74: this is the filterBitsPerKey, not sure if we'd want to make this a public/Responsive config? Have we thought at all about how to handle slateDB configuration yet?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
```diff
@@ -262,15 +270,19 @@ private static RemoteKVTable<?> createMongo(
...
       final ResponsiveMetrics.MetricScopeBuilder scopeBuilder
   ) {
     return sessionClients.rs3TableFactory().kvTable(
         params.name().tableName(),
         config,
+        ttlResolver,
         responsiveMetrics,
-        scopeBuilder
+        scopeBuilder,
+        () -> numPartitionsForKafkaTopic(sessionClients.admin(), changelogTopicName)
```
ableegoldman, line 285: I'm using the number of kafka partitions in the changelog for the number of logical shards for now, but not sure if it we might want to derive it in some other way later on (eg the upstream kafka topic's partitions rather than changelog) ?

ableegoldman: for example IIRC when they scaled up the deduplicator it was much simpler to do because only the input topics (and I guess repartitions?) had to be scaled, not the changelogs. 

although...I would think Streams would check the changelog partition count and throw an error if it doesn't match the expected number based on source topic partitions...so maybe I'm just misremembering this? @agavra do you happen to recall the details here?

rodesai: IIRC the changelog topic still need to be scaled (after all, the store still needs to write into the changelog topic partition that corresponds to the source topic partition(s) the task consumes). I think we just didn't need to redistribute the data because we don't restore from 0 each time. So as long as the changelog topic is fully flushed to the remote store before the scaleup, the scaleup is safe and cheap (no redistribution).

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
rodesai: I think we can just take the configs on a case-by-case basis and add a config property for the ones where we think it's appropriate. I wouldn't really expect users to configure this particular setting. I'd just let this be the default and then allow it to be overrided for specific stores (the one I have in mind is the fact store where we want to set it to something high (I'm thinking 20 bits) to optimize for dedupers)

ableegoldman: ah I didn't realize we were going to implement a fact store for rs3. have you put any thought into what exactly it would look like yet? ie any difference besides the filter bits value?

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/PartitionedOperations.java
ableegoldman: ah that's right I remember now. thanks for clarifying

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -14,38 +14,76 @@
...
       final ResponsiveMetrics responsiveMetrics,
-      final ResponsiveMetrics.MetricScopeBuilder scopeBuilder
+      final ResponsiveMetrics.MetricScopeBuilder scopeBuilder,
+      final Supplier<Integer> computeNumKafkaPartitions
   ) {
-    Map<String, String> storeIdMapping = config.getMap(
+    final Map<String, String> storeIdMapping = config.getMap(
         ResponsiveConfig.RS3_LOGICAL_STORE_MAPPING_CONFIG);
-    final String storeIdHex = storeIdMapping.get(name);
+    final String storeIdHex = storeIdMapping.get(storeName);
```
rodesai, line 54: once we switch to using a non-uuid store id, we should be able to drop this mapping and just create the store using its name right?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
rodesai: Nah that was the main difference in my mind for the near term. The only other thing I can think of that we may want to tweak is the compaction algorithm. We can probably take advantage of the fact that it's insert-only and time-ordered. But that's for later.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
ableegoldman: yep i'll do that in a followup soon

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
ableegoldman: ah ok. was just wondering about this since discussing how to support window stores in CreateStore. Sounds like we should add a "storetype" field that combines the "schema" enums for all store types, so the options would be kv, fact, window, window-with-duplicates, and session

PR #443, opened 2025-04-03
Rs3 Window Operations

hachikuji:
Adds support for window tables. Basically a big bunch of banal boilerplate to send window timestamps as needed in the GRPC protocol.


kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -54,6 +49,35 @@ public RemoteKVTable<WalEntry> kvTable(
...
+        storeId,
+        rs3Client,
+        pssPartitioner,
+        responsiveMetrics,
+        scopeBuilder
+    );
+  }
+
+  private UUID lookupStoreId(ResponsiveConfig config, String name) {
+    Map<String, String> storeIdMapping = config.getMap(
```
ableegoldman, line 72: very minor nit: use `final` everywhere possible (we actually have a checkstyle rule for this in kafka streams, might be useful to add one here too 🤷‍♀️ )

ableegoldman: ditto (and especially) for the method parameters

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowFlushManager.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @Override
+  public String logPrefix() {
+    return tableName() + ".rs3.flushmanager";
+  }
+
+  @Override
+  public long streamTime() {
+    return 0;
```
ableegoldman, line 131:
```suggestion
    return streamTime;
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowFlushManager.java
```diff
@@ -0,0 +1,177 @@
...
+    writers.clear();
+    return RemoteWriteResult.success(null);
+  }
+
+  @Override
+  public String failedFlushInfo(
+      final long batchOffset,
+      final Integer failedTablePartition
+  ) {
+    return String.format("");
```
ableegoldman, line 121: I forget how exactly this gets used but we probably want to log at least the batch offset and failedTablePartition, probably  also the current streamTime. And definitely the table name if it's not already included in the error message (maybe compare other implementations to see what's included in this info and what isn't)

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowFlushManager.java
```diff
@@ -0,0 +1,177 @@
...
+        writtenOffsets.get(pssId),
+        kafkaPartition
+    );
+    writers.put(pssId, writer);
+    return writer;
+  }
+
+  @Override
+  public void writeAdded(final WindowedKey key) {
+    streamTime = Long.max(streamTime, key.windowStartMs);
```
ableegoldman, line 99: I know it's not actually being used anywhere at the moment so it doesn't matter too much, but to avoid confusion in the future we might want to name this more specifically to indicate that it isn't the persisted streamTime but rather the stream-time of the current pending batch (eg  `batchStreamTime` or `pendingStreamTime` or something better)

OTOH just thinking about how we'll want to use the streamTime in the future, maybe we actually do want the persisted streamTime only, to ensure it monotonically increases? Or will slateDB handle that for us automatically 🤔

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
```diff
@@ -0,0 +1,226 @@
...
+
+  public RS3WindowTable(
+      final String name,
+      final UUID storeId,
+      final RS3Client rs3Client,
+      final PssPartitioner pssPartitioner
+  ) {
+    this.name = Objects.requireNonNull(name);
+    this.storeId = Objects.requireNonNull(storeId);
+    this.rs3Client = Objects.requireNonNull(rs3Client); // TODO: Use metered client
```
ableegoldman, line 64: just want to make sure this doesn't get lost, is this planned for a followup PR?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
```diff
@@ -0,0 +1,226 @@
...
+          name,
+          kafkaPartition
+      ));
+    }
+
+    this.lssId = new LssId(kafkaPartition);
+    LssMetadata lssMetadata = rs3ClientUtil.fetchLssMetadata(lssId);
+    this.fetchOffset = lssMetadata.lastWrittenOffset();
+
+    long initialStreamTime = -1; // TODO: Initialize from RS3 metadata?
```
ableegoldman, line 84: nit: this can be final

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
```diff
@@ -0,0 +1,226 @@
...
+        kafkaPartition,
+        pssPartitioner,
+        keySerde,
+        lssMetadata.writtenOffsets(),
+        initialStreamTime
+    );
+    return flushManager;
+  }
+
+  private void throwIfPartitionNotInitialized(int kafkaPartition) {
```
ableegoldman, line 99:
```suggestion
  private void throwIfPartitionNotInitialized(final int kafkaPartition) {
```

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
```diff
@@ -0,0 +1,226 @@
...
+    return new Put(keySerde.serialize(key), value);
+  }
+
+  @Override
+  public WalEntry delete(final int kafkaPartition, final WindowedKey key) {
+    return new Put(keySerde.serialize(key), null);
+  }
+
+  @Override
+  public long fetchOffset(final int kafkaPartition) {
```
ableegoldman, line 223: not really related to this PR but `fetchOffset` is kind of a vague/misleading name for this API. Maybe we should rename to something like `lastWrittenOffset` or `restorationStartOffset`, etc?

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowedKeySerde.java
```diff
@@ -0,0 +1,26 @@
+package dev.responsive.kafka.internal.db.rs3;
+
+import dev.responsive.kafka.internal.utils.WindowedKey;
+import java.nio.ByteBuffer;
+import org.apache.kafka.common.utils.Bytes;
+
+public class RS3WindowedKeySerde {
+
+  public byte[] serialize(WindowedKey key) {
+    byte[] result = new byte[key.key.get().length + 8];
```
ableegoldman, line 10: (sorry for all these nits): use final here and for both method params

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3Writer.java
```diff
@@ -0,0 +1,144 @@
...
+    retryBuffer.add(entry);
+    ifActiveStream(sender -> sender.sendNext(entry));
+  }
+
+  private void ifActiveStream(Consumer<StreamSender<WalEntry>> streamConsumer) {
+    if (sendRecv.isActive()) {
+      try {
+        streamConsumer.accept(sendRecv.sender());
+      } catch (final RS3TransientException e) {
+        // Retry the stream in flush()
```
ableegoldman, line 93: guessing this is a TODO for later (which is fine) but perhaps we should log this exception at `warn` (or `info`?)

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/RemoteWindowOperations.java
```diff
@@ -106,6 +109,22 @@ public static RemoteWindowOperations create(
       case MONGO_DB:
         table = createMongo(params, sessionClients, partitioner, responsiveConfig);
         break;
+      case RS3:
+        final var responsiveMetrics = loadMetrics(appConfigs);
+        final var scopeBuilder = responsiveMetrics.storeLevelMetricScopeBuilder(
+            Utils.extractThreadIdFromThreadName(Thread.currentThread().getName()),
+            changelog,
+            params.name().tableName()
```
ableegoldman, line 117: I know we use `#tableName` for the existing caller of this method but I think it should actually be `#kafkaName` (here and in PartitionedOperations)

 We seem to use the kafka store name rather than the table name for these metrics elsewhere

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/RemoteWindowOperations.java
```diff
@@ -213,6 +232,26 @@ private static RemoteWindowTable<?> createMongo(
...
+      final SessionClients sessionClients,
+      final ResponsiveConfig config,
+      final ResponsiveMetrics responsiveMetrics,
+      final ResponsiveMetrics.MetricScopeBuilder scopeBuilder
+  ) {
+    if (params.schemaType() == SchemaTypes.WindowSchema.STREAM || params.retainDuplicates()) {
+      throw new UnsupportedOperationException("Duplicate retention is not yet supported in RS3");
+    }
+
+    // TODO: Pass through retention period once we have support for TTL
```
ableegoldman, line 246: FYI I'm actually doing this as part of the [CreateStore implementation](https://github.com/responsivedev/responsive-pub/pull/440/files#diff-7093c37c9cd2cd8b4ee477e375bb6a7802c8e134cf1e540cfd7505119aaced47R282)

It's just setting the default ttl for the store (so we still need the STREAM_TIME clock type for full window store ttl support) but we can do the same thing here. So I can handle this as part of integrating the CreateStore API with the RS3 window stores after both PRs are merged

ableegoldman (commented):
Mostly minor comments, from a high level this all looks good

hachikuji:
As I was discussing the patch with @ableegoldman a bit, I realized there is a fork in the road that I probably haven't given enough consideration to. I've taken the approach here of making the RS3 server agnostic to window store semantics. The window start time is baked directly into the key that gets shipped to RS3 and the server knows nothing about it. An alternative is to involve the server directly. We could pass through the window start time in the `WriteWalSegment` API and we could leave it up to the RS3 server to decide how to encode it in the data. We could then modify the range API to additionally pass through the window start time that is being queried. This is a bit more future proof since the data format would be encapsulated in RS3. But the larger point is allowing RS3 to understand window semantics, which may be useful when trying to optimize performance or just for trying to debug/understand usage. For example, it would allow us to create time-based indexes.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3Writer.java
hachikuji: Oh, actually it is already done. The comment is just explaining why its safe to ignore the exception.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
hachikuji: Agreed. I opened a spearate PR https://github.com/responsivedev/responsive-pub/pull/444.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
hachikuji: Yeah, old habits die hard. I'll try to fix the ones I spot.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowTable.java
hachikuji: Oh actually I implemented this through the alternate constructor above. I'll remove the TODO.

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -92,6 +92,38 @@ public RemoteKVTable<WalEntry> kvTable(
...
+        name,
+        storeId,
+        rs3Client,
+        pssPartitioner,
+        responsiveMetrics,
+        scopeBuilder
+    );
+  }
+
+  private UUID lookupStoreId(
```
ableegoldman, line 114: fyi i'm just about to merge the `rs3` PR to convert store ids from UUID to string, haven't done the `responsive-pub` one yet but it probably makes sense to just drop the UUID and use string for the window store now so you don't have to rebase this

kafka-client/src/main/java/dev/responsive/kafka/internal/stores/ResponsiveWindowStore.java
```diff
@@ -102,7 +103,8 @@ public void init(final StateStoreContext storeContext, final StateStore root) {
           params,
           appConfigs,
           config,
-          window -> window.windowStartMs >= minValidTimestamp()
+          window -> window.windowStartMs >= minValidTimestamp(),
+          Optional.empty() // FIXME: Provide TTLResolver
```
hachikuji, line 107: FIXME!

ableegoldman: ah, yeah, we actually don't support ttl with window stores at the moment (and probably never will, imo custom ttl only makes sense for KV stores)

ableegoldman: ie `Optional.empty` is correct

ableegoldman: (in theory we could pass in a ttlResolver with the default ttl instead of the retentionPeriod parameter since we use ttl for retention here, but I think it's cleaner to keep them separate, especially on the off chance we do want to support custom user-defined ttl for window stores some day in the future 🤷‍♀️ )

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3TableFactory.java
```diff
@@ -64,23 +80,47 @@ public RemoteKVTable<WalEntry> kvTable(
...
+      final Duration defaultTtl,
+      final ResponsiveMetrics responsiveMetrics,
+      final ResponsiveMetrics.MetricScopeBuilder scopeBuilder,
+      final Supplier<Integer> computeNumKafkaPartitions
+  ) {
+    final var rs3Client = connector.connect();
+    final UUID storeId = createdStores.computeIfAbsent(storeName, n -> createStore(
+        storeName,
+        CreateStoreTypes.StoreType.WINDOW,
+        Optional.of(ClockType.STREAM_TIME),
```
ableegoldman, line 94: just a heads up we still throw a ConfigErr on the server side if stream time is used  so we'll need to implement that as a followup

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowFlushManager.java
```diff
@@ -0,0 +1,178 @@
...
+  public String tableName() {
+    return table.name();
+  }
+
+  @Override
+  public TablePartitioner<WindowedKey, Integer> partitioner() {
+    return new PssTablePartitioner<>(pssPartitioner) {
+      @Override
+      public byte[] serialize(final WindowedKey key) {
+        // TODO: Get rid of this
```
ableegoldman, line 70: ?

ableegoldman (approved):
LGTM!

kafka-client/src/main/java/dev/responsive/kafka/internal/db/rs3/RS3WindowFlushManager.java
hachikuji: Not sure what I had in mind. I'll get rid of the TODO since the implementation seems ok to me.

PR #446, opened 2025-04-17
changes to get rs3 healthcheck working

rodesai:
- expose the pss id list from Store
- support passing the internal api key to the auth interceptor


hachikuji (approved):
LGTM

PR #450, opened 2025-04-23
Change license loading precedence

hachikuji:
Currently we will not load a license if the API key is provided through `responsive.platform.api.key`. This patch changes the precedence to use this key as a fallback only if license loading fails. 

Additionally, we change RS3 initialization logic to not pass through an API key when the license is not `UsageBasedV1`.


rodesai (approved):
LGTM!

PR #455, opened 2025-05-01
Introduce RS3ConfigSetter

ableegoldman:
Followup to https://github.com/responsivedev/responsive-pub/pull/453 to make the filterBitsPerKey (and any future rs3/storage related settings) configurable by users for individual stores

### DIFF
--- a/kafka-client/src/main/java/dev/responsive/kafka/api/config/RS3ConfigSetter.java
+++ b/kafka-client/src/main/java/dev/responsive/kafka/api/config/RS3ConfigSetter.java
@@ -1,0 +1,74 @@
+/*
+ *
+ *  Copyright 2025 Responsive Computing, Inc.
+ *
+ *  This source code is licensed under the Responsive Business Source License Agreement v1.0
+ *  available at:
+ *
+ *  https://www.responsive.dev/legal/responsive-bsl-10
+ *
+ *  This software requires a valid Commercial License Key for production use. Trial and commercial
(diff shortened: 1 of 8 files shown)


kafka-client/src/main/java/dev/responsive/kafka/api/config/RS3ConfigSetter.java
```diff
@@ -0,0 +1,74 @@
...
+  RS3StoreParams keyValueStoreConfig(String storeName, KVSchema schema);
+
+  /**
+   * Sets the configs for a {@link org.apache.kafka.streams.state.WindowStore}
+   * with the provided name and schema. Can return {@link Optional#empty()} to defer to
+   * the default settings for the given store type.
+   *
+   * @return the configs to override for this store
+   */
+  RS3StoreParams windowStoreConfig(String storeName, WindowSchema schema);
```
hachikuji, line 41: Perhaps we could return a more specific type (e.g. `RS3WindowStoreParams`) in case there are different configurations for the respective store?

ableegoldman: yeah, that's probably the right thing to do...I considered that but ended up consolidating them to reduce API surface. But the problem with a generic RS3StoreParams is that it's up to the user to use the correct base/default configs for the given store type. Which will only become worse if we do have different configs for different store types in the furure

i'll go back to the store-specific params

